PR #1, opened 2015-04-11
handle small memoryviews in extract_buffers

minrk:


minrk:
Disabled @coveralls horrible comments.

PR #4, opened 2015-04-17
import cloudpickle directly

rgbkrk:
Now that `cloudpickle` is in its [own package up on PyPI](https://pypi.python.org/pypi/cloudpickle) it can be installed and used directly.

Benefits:

* supports Python 3
* has its own suite of tests (nothing public prior)
* active development

/cc @ogrisel

### DIFF
--- a/ipython_kernel/pickleutil.py
+++ b/ipython_kernel/pickleutil.py
@@ -97,7 +97,7 @@ def use_cloudpickle():
     
     adds support for object methods and closures to serialization.
     """
-    from cloud.serialization import cloudpickle
+    import cloudpickle
     
     global pickle
     pickle = cloudpickle


ogrisel:
That is quick software adoption :)

ogrisel:
Don't you need to add `"cloudpickle"` to the `install_requires` of the `setup.py` of `ipython_kernel`?

rgbkrk:
`cloud` was always an optional dependency, I figured folks didn't want us forcing this dependency on them.

minrk:
Yup, I'm not sure if that was a serious comment. It's not a dependency, so it won't go in install_requires.

rgbkrk:
Sorry, yes, that was a serious comment. My only intention here was to migrate the cloudpickler to the one that is getting maintenance.

PR #5, opened 2015-04-17
migrate json_clean to ipython_kernel

minrk:
these parts of the code are only used in the IPython kernel, so don't belong in Jupyter Client, where jsonutil was moved.

jupyter_client.jsonutil preserves the date handling

This also disables the casting fallback, which ensures json_clean could never raise. That was probably never a good idea, and just hides bugs.


jasongrout:
@minrk: thanks!

PR #8, opened 2015-04-21
Forward port "Set axis background to transparent in inlinebackend"

Carreau:
See ipython/ipython/8336

```
Revert "Set axis background to transparent in inlinebackend"

This reverts commit 902c754fcbd0581ea33912e583cd3ccb8bdb9a5b.

Making axis transparent seem to confuse and annoy lot of people,

See #7964 (do not fix it, this will be a 4.0 or 5.0 thing),
the real fix would be to use matplotlib themes.
```


takluyver:
We already merged the PR against 3.x, so merging this.

PR #18, opened 2015-05-27
handle datetime in json_clean

minrk:
datetime objects are allowed, and serialized to ISO8601

closes ipython/ipywidgets#35

### DIFF
--- a/ipykernel/jsonutil.py
+++ b/ipykernel/jsonutil.py
@@ -162,6 +162,8 @@ def json_clean(obj):
         for k,v in iteritems(obj):
             out[unicode_type(k)] = json_clean(v)
         return out
+    if isinstance(obj, datetime):
+        return obj.strftime(ISO8601)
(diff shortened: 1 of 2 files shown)


rgbkrk:
Looks good.

jasongrout:
By the way, IIRC, there is a difference between how python and how browsers handle local vs. UTC.  If I remember correctly, we were having issues where python converted things to UTC, and the browser assumed it was local time, or vice versa, if the timezone offset wasn't explicitly specified.  There's a whole can of worms there.

rgbkrk:
I love time zones.

takluyver:
Looking at the contents service code, I think we're leaving it up to code that creates datetimes to ensure they're 'timezone aware', and then `.isoformat()` will do the right thing. Python makes it annoyingly difficult to produce timezone-aware datetimes, annoyingly.

PR #20, opened 2015-05-29
allow specifying the name of the IPython kernel

minrk:
for multiple IPython kernels at the same time.

Related to jupyter/jupyter_client#26


takluyver:
LGTM

PR #23, opened 2015-06-01
Add content from notebook/pull/124

jdfreder:
sibling PRs:
https://github.com/jupyter/notebook/pull/124

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -1,0 +1,14 @@
+Installation
+------------
+
+IPython can be installed (different python versions, virtualenv or conda environments) can
+be installed as a kernel by following these steps:
+
+* make sure that the desired python installation is active (e.g. activate the environment)
+  and ipython is installed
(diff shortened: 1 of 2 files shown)


takluyver:
I'm actually inclined to put the docs for this package in ipython and in jupyter_client, so that there's one less place for documentation to live. The only file there at present is the document on wrapper kernels, and I've already added that [to the jupyter_client docs](http://jupyter-client.readthedocs.org/en/latest/wrapperkernels.html).

jdfreder:
I agree IPython makes sense, but for both.  Why move either of these two (in this PR) to jupyter-client though?

takluyver:
Oh no, these two both belong in IPython. I meant there are some other bits that could conceivably belong here that I'd put in jupyter_client (like the wrapperkernels doc).

jdfreder:
Closing in favor of https://github.com/ipython/ipython/pull/8500

PR #25, opened 2015-06-22
Adding comm_info_request and comm_info_reply messages

SylvainCorlay:
Adding a `comm_list_[request/reply]` shell message. This requires https://github.com/jupyter/jupyter_client/pull/34.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -119,7 +119,7 @@ class Kernel(SingletonConfigurable):
         # Build dict of handlers for message types
         msg_types = [ 'execute_request', 'complete_request',
                       'inspect_request', 'history_request',
-                      'kernel_info_request',
+                      'comm_info_request', 'kernel_info_request',
                       'connect_request', 'shutdown_request',
                       'apply_request', 'is_complete_request',
                     ]
@@ -427,7 +427,7 @@ class Kernel(SingletonConfigurable):
     def do_inspect(self, code, cursor_pos, detail_level=0):
         """Override in subclasses to allow introspection.
         """
-        return {'status': 'ok', 'data':{}, 'metadata':{}, 'found':False}
+        return {'status': 'ok', 'data': {}, 'metadata': {}, 'found': False}
 
     def history_request(self, stream, ident, parent):
         content = parent['content']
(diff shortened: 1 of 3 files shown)


ipykernel/kernelbase.py
```diff
@@ -470,6 +470,19 @@ def kernel_info_request(self, stream, ident, parent):
                                 self.kernel_info, parent, ident)
         self.log.debug("%s", msg)
 
+    def comm_list_request(self, stream, ident, parent):
+        content = parent['content']
+
+        # Should this be moved to ipkernel?
+        if hasattr(self, 'comm_manager'):
+            comm_list = self.comm_manager.comms.keys()
```
minrk: This will presumably need to be cast to a list.

SylvainCorlay: that's right. Python 3 sorry.

SylvainCorlay:
Done

ipykernel/kernelbase.py
```diff
@@ -470,6 +470,19 @@ def kernel_info_request(self, stream, ident, parent):
                                 self.kernel_info, parent, ident)
         self.log.debug("%s", msg)
 
+    def comm_list_request(self, stream, ident, parent):
+        content = parent['content']
+
+        # Should this be moved to ipkernel?
```
jasongrout: ipykernel

SylvainCorlay: I think that it is @minrk call.

minrk: I think this is a fine place for it.

SylvainCorlay:
Rebased to use the new schema for the `comm_info_reply`. 

```yml
Message type: ``comm_info_request``::

    content = {
    }

Message type: ``comm_info_reply``::

    content = {
        # A dictionary of the comms, indexed by uuids.
        'comms': {
            comm_id: {
                'target_name': str,
            },
        },
    }
```

SylvainCorlay:
I added the ability to filter by target name as discussed earlier.

minrk:
thanks, @SylvainCorlay!

SylvainCorlay:
I hesitated to give a *list* of target names rather than a single target name for the filter, but it would have been the only message having a list of target names, so it is probably more consistent with the rest of the messages to only allow one.

minrk:
@SylvainCorlay good choice. I don't imagine the use case for people wanting more than one but less than all is worth having the extra logic.

SylvainCorlay:
How can we move forward with this? Should there be a release of jupyter-client?

ipykernel/tests/test_message_spec.py
```diff
@@ -419,6 +423,14 @@ def test_kernel_info_request():
     validate_message(reply, 'kernel_info_reply', msg_id)
 
 
+def test_comm_info_request():
+    flush_channels()
+
+    msg_id = KC.comm_info()
```
minrk: Can you add a `if not hasattr(KC, 'comm_info'): raise SkipTest()`? Then this test will run while comm_info is defined, but skip while jupyter_client is < 4.1.

SylvainCorlay: Done!

SylvainCorlay:
(Done)

minrk:
@SylvainCorlay thanks! Now I think we can start getting this tested for those living on master.

PR #26, opened 2015-07-05
support kernelspec install prefix

minrk:
cf jupyter/jupyter_client#42

### DIFF
--- a/ipykernel/kernelspec.py
+++ b/ipykernel/kernelspec.py
@@ -74,7 +74,7 @@ def write_kernel_spec(path=None):
     return path
 
 
-def install(kernel_spec_manager=None, user=False, kernel_name=None):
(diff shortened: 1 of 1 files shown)


ipykernel/kernelspec.py
```diff
@@ -115,7 +115,18 @@ def install(kernel_spec_manager=None, user=False, kernel_name=None):
     parser.add_argument('--name', type=str, default=KERNEL_NAME,
         help="Specify a name for the kernelspec."
         " This is needed to have multiple IPython kernels at the same time.")
+    parser.add_argument('--prefix', type=str,
+        help="Specify an install prefix for the kernelspec."
+        " This is needed to install into a non-defualt location, such as a conda/virtual-env.")
```
rgbkrk: s/defualt/default/

minrk: fixed.​

PR #19, opened 2015-05-27
Hardcode widgets

jdfreder:
sibling PRs: 
https://github.com/jupyter/notebook/pull/120
https://github.com/ipython/ipywidgets/pull/38

PR #31, opened 2015-07-14
only catch ImportError for widget hardcoding

minrk:
finish up #19


takluyver:
This doesn't now change the `except` clause at all :-S

minrk:
arg, fixed.

PR #34, opened 2015-07-23
Use traitlets 4.1 .tag() for config metadata

jasongrout:
This depends on https://github.com/ipython/traitlets/pull/53, which was included in traitlets 4.1.0.

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -101,13 +101,13 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
     flags = Dict(kernel_flags)
     classes = [IPythonKernel, ZMQInteractiveShell, ProfileDir, Session]
     # the kernel class, as an importstring
-    kernel_class = Type('ipykernel.ipkernel.IPythonKernel', config=True,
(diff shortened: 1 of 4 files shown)


jasongrout:
@minrk, @SylvainCorlay - I think this ready for review now.

SylvainCorlay:
@Carreau?

Carreau:
> @Carreau?

Cf comment on https://github.com/jupyter/nbformat/pull/14 , and on Jupyter Hackpad. 

I'm cleaning the PR list, and this is stalling because we need to decide on 4.1 as a deps explicitly and move forward once and for all instead of having Jason rebasing 12 PRs every couple of weeks.
At least me closing can allow you to complain tomorrow at the meeting and get a decision on which we can move forward.

Carreau:
Ok, one PR at a time. Reopening this one. 
Lets get this in, add test on CI that prevent using older syntax, and do the same on another repo.

takluyver:
I don't think we really need to create tests for this. There are visible warnings if something uses the old style, that should be fine.

Carreau:
Ok, last read of that and merging.

Carreau:
Thanks @jasongrout !

SylvainCorlay:
:+1:

PR #36, opened 2015-07-23
support blocking is_complete in inprocess

minrk:
- define blocking_class (see jupyter/jupyter_client#55)
- define is_complete

sibling of jupyter/qtconsole#33 jupyter/jupyter_client#55


minrk:
Merging in the next day or so unless I hear otherwise.

PR #41, opened 2015-08-19
Avoid creating default connection dir if not needed

aburgm:
If a connection file is specified with an absolute path via the '-f'
command line option, ipykernel still makes sure that the default connection
directory exists, even though it is not needed. Instead, simply create the
directory for the path to the connection file that we end up actually
using.

This introduces a slight change in behaviour in that if one specifies a
connection file via the '-f' option and the given directory does not
exist, it will be created instead of an error being generated.

In my case, I am specifying a different path to the connection file because the default path is not writable. While I could (and probably should) change the default path with the corresponding environment variables, it still left me wondering whether the default connection directory should be created even if it is not going to be used.

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -115,9 +115,7 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
     # connection info:
     connection_dir = Unicode()
     def _connection_dir_default(self):
-        d = jupyter_runtime_dir()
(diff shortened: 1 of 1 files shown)


minrk:
The test failure looks like there is another case where the directory needs to be created.

aburgm:
Indeed... I updated the commit so it creates the directory already during initialization, instead of when the file is actually written.

PR #43, opened 2015-08-21
add `ipython kernel install`

minrk:
replacement for 'ipython kernelspec install-self'

Previously only available as `python -m ipykernel.kernelspec`

cc @stefanv @jasongrout


Carreau:
+1

rgbkrk:
Good to know.

PR #50, opened 2015-09-04
Write connection file before writing its location to the log

aburgm:
Is there a particular reason this is the other way around?

When the connection file is first written and then logged, then external programs could monitor the log output and use the connection file (e.g. connect a client) without race condition.


Carreau:
That's sill don't guaranties on all system that the log will be available after the connexion file. 

Also ipykernel is the only one that (for historical reason) write it's own connection file. Your external program should write the connection file itself, and pass it to ipykernel as an argument. 

But i'm ok with this change, though I don't want to guaranty and maintain that it will work.

minrk:
This should be a fine change, but as @Carreau mentioned, 

I really need to get to the `jupyter kernel` entrypoint, for providing a wrapper for launching kernels with a KernelManager always attached.

PR #59, opened 2015-09-22
Fixes typo in InstallIPythonKernelSpecApp

michaelpacer:
Fixes #58, and adds test.

PR #60, opened 2015-09-22
fix typo that broke master

Carreau:
Well, that should fix failure on other repos.
Merging as soon as travis go green.


Carreau:
#59 was fater :-)

PR #66, opened 2015-10-17
Fix version attributes import

adalq:
`import *` doesn't import attributes which start with an underscore, so it
won't import `__version__`


minrk:
Thanks - I never noticed this because Python 3 does the right thing, but Legacy Python does not.

adalq:
That's interesting, the Python 3 docs suggest that this behavior didn't change, and that `__version__` would not be imported by Python 3 either.

Unfortunately I don't have a Python 3 handy to test.

https://docs.python.org/3/reference/simple_stmts.html#the-import-statement

PR #67, opened 2015-10-19
Adding missing target name when creating a new comm from the front-end

SylvainCorlay:
@minrk @jdfreder 
We were missing the `target_name`attribute when creating a Comm on reception of `comm_open` messages.

### DIFF
--- a/ipykernel/comm/manager.py
+++ b/ipykernel/comm/manager.py
@@ -106,6 +106,7 @@ class CommManager(LoggingConfigurable):
                     kernel=self.kernel,
                     iopub_socket=self.iopub_socket,
                     primary=False,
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
I wonder if we could have a 4.1.1 with that fix!

jdfreder:
I'm surprised this was missing all along!  :+1:

SylvainCorlay:
Well, this is not a part of the API that we have been using so much.

jdfreder:
@minrk may we release a 4.1.1 for this?

minrk:
Definitely. I can push it tomorrow.

minrk:
Released 4.1.1

PR #71, opened 2015-10-29
Deprecate ipyparallel-specific code

### LINKED ISSUE #17: Proposal: remove some ipyparallel-specific functionality
Mainly the `apply` message-related bits. These aren't part of the wider Jupyter message protocol, and are wholly Python specific. These can be moved to ipyparallel, and attached to the Kernel in a subclass.

minrk:
- deprecation warnings on serialization, data pub
- Move ipyparallel-specific metadata to methods easier to override in subclasses
- deprecate handling messages not in spec: apply, clear, abort.

related: ipython/ipyparallel#61

closes #17

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -124,6 +124,31 @@ class IPythonKernel(KernelBase):
         super(IPythonKernel, self).set_parent(ident, parent)
         self.shell.set_parent(parent)
 
+    def init_metadata(self, parent):
+        """Initialize metadata.
+        
+        Run at the beginning of request handlers.
+        """
+        md = super(IPythonKernel, self).init_metadata(parent)
+        # FIXME: remove ipyparallel-specific code
+        md.update({
+            'dependencies_met' : True,
+            'engine' : self.ident,
+        })
+        return md
+    
+    def finish_metadata(self, parent, metadata, reply_content):
+        """Finish populating metadata.
+        
+        Run after completing a request handler.
+        """
+        # FIXME: remove ipyparallel-specific code:
+        metadata['status'] = reply_content['status']
+        if reply_content['status'] == 'error' and reply_content['ename'] == 'UnmetDependency':
+                metadata['dependencies_met'] = False
+
+        return metadata
+
     def _forward_input(self, allow_stdin=False):
         """Forward raw_input and getpass to the current frontend.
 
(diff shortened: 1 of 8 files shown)


minrk:
This should be ready to go. I'll merge in the next day or so unless I hear otherwise.

minrk:
I changed my mind a little - I'll wait until we can ship ipyparallel 5.0 before releasing this one. Should be soon.

minrk:
ipyparallel 5.0 shipped, so this should be able to go in, now.

ipykernel/ipkernel.py
```diff
@@ -124,6 +121,31 @@ def set_parent(self, ident, parent):
         super(IPythonKernel, self).set_parent(ident, parent)
         self.shell.set_parent(parent)
 
+    def init_metadata(self, parent):
+        """Initialize metadata.
+        
+        Run at the beginning of request handlers.
+        """
+        md = super(IPythonKernel, self).init_metadata(parent)
+        # FIXME: remove ipyparallel-specific code
```
takluyver: What's the plan for dealing with these?

ipykernel/kernelbase.py
```diff
@@ -164,6 +168,25 @@ def dispatch_control(self, msg):
         sys.stderr.flush()
         self._publish_status(u'idle')
 
+    def should_handle(self, stream, msg, idents):
+        """Check whether a message should be handled
+        
+        Allows subclasses to prevent handling of certain messages (e.g. aborted requests).
```
takluyver: This only affects messages on the shell channel - it's not checked for the control channel. I think that's probably what we want, but maybe it should be mentioned in the docstring.

ipykernel/kernelbase.py
```diff
@@ -340,6 +344,22 @@ def send_response(self, stream, msg_or_type, content=None, ident=None,
         """
         return self.session.send(stream, msg_or_type, content, self._parent_header,
                                  ident, buffers, track, header, metadata)
+    
+    def init_metadata(self, parent):
+        """Initialize metadata.
+        
+        Run at the beginning of request handlers.
```
takluyver: These methods are only called for `execute_request` plus the ipyparallel handlers. Is that intentional? If so, it should probably be in the docstring.

takluyver:
Other than those things, :+1:

ipykernel/ipkernel.py
minrk: Going forward, these are provided by a subclass in ipyparallel. Removing these would be a breaking change, and would break ipyparallel < 5. We can consider dropping then in ipykernel 5 or 6.

minrk:
Comments/docstrings updated to address the comments.

ipykernel/kernelbase.py
takluyver: @minrk this comment didn't get a response, in case you missed it when reviewing.

minrk: Oops, I thought I addressed it, but didn't notice that these appear twice.

PR #73, opened 2015-11-06
support a full message in initial comm open

minrk:
was missing metadata, buffers args


SylvainCorlay:
:+1: This could be worth another ipykernel release?

minrk:
I don't think we need to rush it out, but yeah, I think we can put another kernel release out with this soon.

rgbkrk:
Thanks for addressing all of these. Looking forward to getting bitjet up to date.

PR #76, opened 2015-11-09
default display_name to kernel_name if specified

### LINKED ISSUE #75: IPython kernel installation: consider defaulting --display-name to --name
@cel4 opened ipython/ipython#8956

At the moment `ipython kernel install --name <name>` defaults the `display-name` in the web UI to python2/3 which I find rather unintuitive.

It would probably be reasonable to assume that `name` and `display-name` coincide, unless the user explicitly sets the `--display-name` argument.

minrk:
`ipython kernel install --name foo` installs kernel with name foo and display_name foo, without needing to specify two arguments.

closes #75


damianavila:
LGTM, I find the code self-explanatory,so no need to add comments for me... but I am not against those if you think it makes clarifications...

PR #85, opened 2015-12-08
publish IOPub from a background thread

### LINKED ISSUE #77: Run IOPub in a thread
We should be sending IOPub messages in their own thread. This would solve various issues with flushing/threading/etc.

It shouldn't be complicated, other than preserving backward-compatibility for APIs which assume direct access to a Socket (consumed by widgets, ipyparallel).

minrk:
removes need to call `sys.stdout.flush` during blocking loops for backward-compatibility, a BackgroundSocket is passed around that transparently sends messages via the IOPub thread.

I had a hell of a time debugging weird hangs with multiprocessing, but they seem to be gone, now, thanks to a well-placed threading.Event.

closes jupyter/notebook#700
closes #77


minrk:
I think this should be ready to go, now. The one thing I'm not thrilled about is the workaround of the fork-safety check in `Session.send`, since that's in jupyter_client. It would be nice if there were a way to say "I'm fork-safe, trust me"

minrk:
I made the fork-safe check optional in jupyter/jupytr_client#107, so we can clean that up.

Since this has been ready to go for some time now, I'll merge before too long unless someone wants to do some review.

takluyver:
Sorry for leaving this unreviewed so long. I will try to have a careful look at it either this evening or tomorrow (just about to head home before the rain gets too hard).

ipykernel/iostream.py
```diff
@@ -105,87 +109,183 @@ def _check_mp_mode(self):
...
+
+class BackgroundSocket(object):
+    """Wrapper around IOPub thread that provides zmq Socket API"""
+    io_thread = None
+    
+    def __init__(self, io_thread):
+        self.io_thread = io_thread
+    
+    def __getattr__(self, attr):
+        return getattr(self.io_thread.socket, attr)
```
takluyver: When I'm overriding `__getattr__` and `__setattr__`, I like to check for attempts at `__special__` attribute access, and make them behave as normal; otherwise things can get quite confused.

ipykernel/inprocess/ipkernel.py
```diff
@@ -128,12 +129,10 @@ def _shell_class_default(self):
...
 
     def _stdout_default(self):
-        from ipykernel.iostream import OutStream
-        return OutStream(self.session, self.iopub_socket, u'stdout', pipe=False)
+        return OutStream(self.session, self.iopub_socket, u'stdout')
 
     def _stderr_default(self):
-        from ipykernel.iostream import OutStream
-        return OutStream(self.session, self.iopub_socket, u'stderr', pipe=False)
+        return OutStream(self.session, self.iopub_socket, u'stderr')
```
takluyver: These two are relying on the deprecated API of passing a socket rather than an IOPubThread instance, right?

ipykernel/iostream.py
```diff
@@ -105,87 +109,183 @@ def _check_mp_mode(self):
...
+            tracker = self._pipe_out.send_multipart([self._pipe_uuid] + msg, *args, **kwargs)
+            try:
+                tracker.wait(1)
+            except Exception as e:
+                print("Failed to send: %s" % e, file=sys.__stderr__)
+                pass
+
+
+class BackgroundSocket(object):
+    """Wrapper around IOPub thread that provides zmq Socket API"""
```
takluyver: We're wrapping a zmq socket in a thread, and then wrapping the thread in both a ZMQ-socket-like object and a file-like object?

Is the intention that BackgroundSocket will one day be removed? I can see the use for backwards compatibility, but it seems like a layer of abstraction more than we should need.

takluyver: And for everything besides sending data, it provides an attractive shortcut right across threads, which doesn't seem like a great idea.

ipykernel/iostream.py
```diff
@@ -105,87 +109,183 @@ def _check_mp_mode(self):
...
 
-    def _schedule_flush(self):
-        """schedule a flush in the main thread
+        if self.thread.is_alive():
+            self.io_loop.add_callback(lambda : self._really_send(*args, **kwargs))
+        else:
+            self._really_send(*args, **kwargs)
+    
+    def _really_send(self, msg, *args, **kwargs):
+        """The callback that actually sends messages"""
```
takluyver: Should the docstring note that there are two different destinations that it might actually send messages to?

ipykernel/iostream.py
```diff
@@ -26,76 +31,75 @@
...
-        self._pipe_poller.register(self._pipe_in, zmq.POLLIN)
-        if IOLoop.initialized():
-            # subprocess flush should trigger flush
-            # if kernel is idle
-            IOLoop.instance().add_handler(self._pipe_in,
-                lambda s, event: self.flush(),
-                IOLoop.READ,
-            )
+        self._pipe_in = ZMQStream(pipe_in, self.io_loop)
+        self._pipe_in.on_recv(self._handle_pipe_msg)
```
takluyver: I know this is not really part of this PR, but these names seem backwards to me. Messages go *in* to `_pipe_out`, and come *out* of `_pipe_in`. I'm sure there's a logic to the way they are, but it confused me a bit.

ipykernel/iostream.py
```diff
@@ -105,87 +109,183 @@ def _check_mp_mode(self):
...
+        """
+        with self._flush_lock:
+            if self._flush_timeout is not None:
+                return
+            # non-None to avoid races while waiting for _schedule
+            self._flush_timeout = 'placeholder'
+        
+        # add_timeout has to be handed to the io thread with add_callback
+        def _schedule_in_thread():
+            # FIXME: call_later is a little nicer, if we can require tornado ≥ 4
```
takluyver: Don't we already require tornado 4?

ipykernel/iostream.py
```diff
@@ -105,87 +109,183 @@ def _check_mp_mode(self):
...
+
+    def _schedule_flush(self):
+        """schedule a flush in the IO thread
+        
+        call this on write, to indicate that flush should be called soon.
+        """
+        with self._flush_lock:
+            if self._flush_timeout is not None:
+                return
+            # non-None to avoid races while waiting for _schedule
```
takluyver: I'm not sure what this comment means.

takluyver: On further examination, I think I get it, but a bit more explanation would be useful. We're not all you ;-)

ipykernel/iostream.py
```diff
@@ -105,87 +109,183 @@ def _check_mp_mode(self):
...
+        unless the thread has been destroyed (e.g. forked subprocess).
+        """
+        with self._flush_lock:
+            self._flush_timeout = None
+            data = self._flush_buffer()
+        if data:
+            # FIXME: this disables Session's fork-safe check,
+            # since pub_thread is itself fork-safe.
+            # There should be a better way to do this.
+            self.session.pid = os.getpid()
```
takluyver: Won't this disable the fork check for everything using that same session? Could that cause problems if other things then try to use non-fork-safe bits of Session?

ipykernel/iostream.py
minrk: Yeah, it's a backward-compatibility thing, since iopub_socket is used throughout subclasses that would break without it. I don't see away around it. Nothing in the main thread should *ever* touch the actual zmq socket at any time.

ipykernel/iostream.py
```diff
@@ -105,87 +109,183 @@ def _check_mp_mode(self):
...
-                string = self._flush_buffer()
-                tracker = self._pipe_out.send_multipart([
-                    self._pipe_uuid,
-                    string.encode(self.encoding, 'replace'),
-                ], copy=False, track=True)
-                try:
-                    tracker.wait(1)
-                except:
-                    pass
+            return CHILD
```
minrk: Yup, this is simpler now than it used to be. Good catch.

ipykernel/iostream.py
minrk: The *kernel* doesn't require tornado at all, so this is only necessary if IOLoop is coming from the bundled version in pyzmq. But given how ~everybody installs it, the odds of not having tornado are slim. I only noticed because we don't install tornado on travis for this repo.

ipykernel/iostream.py
minrk: Yes, though all the crashes that prompted the fork check were actually IOPub sends. It should technically be possible, though. I'd like a better way to indicate that IOPub is itself fork-safe, but the Session API doesn't allow it.

minrk: Alternately, we could apply this hack to a dedicated Session that's only used for IOPub that's a clone of the main Session.

ipykernel/iostream.py
takluyver: Does pyzmq's bundled event loop still come from an older version of tornado, then?

minrk:
Re: BackgroundSocket I didn't deprecated the socket API because I don't really have an idea for a better one. If nothing else, it's going to be an object with a send method that takes a message, so I'm not sure how different it's going to be from the socket API that we need to keep for backward compatibility.

Do you have any ideas on that point?

ipykernel/iostream.py
minrk: Yes, through 15.1. 15.2 is up-to-date with tornado 4.3.

I'm inclined to add the tornado dependency to this repo.

takluyver:
Fair enough. Is it worth deprecating the shortcut access to all of the socket methods besides send? What actually uses those?

ipykernel/iostream.py
takluyver: I would be fine with making tornado >= 4 a dependency. For the typical use case of installing it along with the notebook, users will get tornado anyway. And it's not a very heavy dependency.

minrk:
>  Is it worth deprecating the shortcut access to all of the socket methods besides send?

I expect so. I'll take a stab at that, and see what might be using them, if anything.

ipykernel/inprocess/ipkernel.py
minrk: this uses iopub_thread now

ipykernel/iostream.py
minrk: `__methods__` don't passthrough and Socket attr access is deprecated. I didn't find any actual use of it, so it may be overly cautious, though it may be that ipyparallel will rely on it for some transition.

ipykernel/iostream.py
minrk: The only public API is now `send/send_multipart`

ipykernel/iostream.py
minrk: This is the norm for pipes, I think. Messages *go out* on pipe_out (like stdout) and *come in* on pipe_in (like stdin); pipe_out leads out of its process, pipe_in leads into its process.

takluyver: The terminology I'm more used to for pipes is the read end and the write end, which doesn't have this ambiguity. As in:

```python
r, w = os.pipe()
```

ipykernel/iostream.py
```diff
@@ -105,87 +109,184 @@ def _check_mp_mode(self):
...
+    io_thread = None
+    
+    def __init__(self, io_thread):
+        self.io_thread = io_thread
+    
+    def __getattr__(self, attr):
+        """Wrap socket attr access for backward-compatibility"""
+        if attr.startswith('__') and attr.endswith('__'):
+            # don't wrap magic methods
+            raise AttributeError(attr)
```
takluyver: I think `__setattr__` should also be doing a check like this (though it should call super, rather than raising AttributeError.

minrk: Done, and super'd

PR #87, opened 2015-12-16
Check for number types using the python type hierarchy

jasongrout:
This lets numpy numeric types to work with json cleaning functions, for example.

CC @ChakriCherukuri

### DIFF
--- a/ipykernel/jsonutil.py
+++ b/ipykernel/jsonutil.py
@@ -7,6 +7,7 @@ import math
 import re
 import types
 from datetime import datetime
+import numbers
 
 try:
     # base64.encodestring is deprecated in Python 3.x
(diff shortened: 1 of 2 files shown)


jasongrout:
I'm not sure how the test that had `Int(5)` worked, since it gave errors because Int wasn't defined...

minrk:
Thanks!

jasongrout:
@minrk - This is biting us.  It would be awesome if we could do a small patch release with this patch.  How much effort is it to do a patch release?

SylvainCorlay:
@jasongrout thanks for fixing this. I am :+1: for a patch release.

minrk:
Not much. There's another patch we want to get in, then we can cut a release.

minrk:
@jasongrout released 4.2.2 with this.

PR #88, opened 2015-12-18
Don't use FormattedTB as kernel crash handler

minrk:
Just show the regular traceback

FormattedTB adds extra noise designed for debugging interactive code.

Do we want any additional "I died!" or "bug reports here..." messages on a crash like this?

cc @ssanderson @fperez

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -9,12 +9,12 @@ import atexit
 import os
 import sys
 import signal
+import traceback
 
 import zmq
 from zmq.eventloop import ioloop
 from zmq.eventloop.zmqstream import ZMQStream
 
-from IPython.core.ultratb import FormattedTB
 from IPython.core.application import (
     BaseIPythonApplication, base_flags, base_aliases, catch_config_error
 )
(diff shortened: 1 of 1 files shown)


ipykernel/kernelapp.py
```diff
@@ -149,11 +149,13 @@ def abs_connection_file(self):
...
+    
     def init_crash_handler(self):
-        # Install minimal exception handling
-        sys.excepthook = FormattedTB(mode='Verbose', color_scheme='NoColor',
-                                     ostream=sys.__stdout__)
+        sys.excepthook = self.excepthook
+    
+    def excepthook(self, etype, evalue, tb):
+        # write uncaught traceback to 'real' stderr, not zmq-forwarder
+        traceback.print_exception(etype, evalue, tb, file=sys.__stderr__)
```
ssanderson: Looks like the previous `FormattedTB` output was going to `sys.__stdout__`?  Is there a reason for the change here?  (I don't have an opinion one way or another, just curious.)

minrk: Seems like a crash should go to stderr. I'll call the stdout destination a bug in addition to the formatting.

PR #92, opened 2016-01-06
patch faulthandler defaults to underlying stderr

### LINKED ISSUE #91: faulthandler.enable() fails with IOError
Reported by @poldrack on the [notebook](https://github.com/jupyter/notebook/issues/920), but ultimately the issue is really a kernel problem, I can reproduce it from a plain console by connecting to an open kernel using ipykernel master:

```
denali[newjob]> jupyter console --existing kernel-95921.json
Jupyter Console 4.1.0.dev

[ZMQTerminalIPythonApp] Loading IPython extension: storemagic

In [1]: import faulthandler

In [2]: faulthandler.enable()
UnsupportedOperation                      Traceback (most recent call last)
<ipython-input-2-140d8a320d56> in <module>()
----> 1 faulthandler.enable()

/Users/fperez/dev/ipython/ipykernel/ipykernel/iostream.py in fileno(self)
    203 
    204     def fileno(self):
--> 205         raise UnsupportedOperation("IOStream has no fileno.")
    206 
    207     def write(self, string):

UnsupportedOperation: IOStream has no fileno.
```

minrk:
faulthandler requires access to C-level FD, which doesn't make sense for forwarded stdout/stderr. Patch default to be underlying `__stderr__` instead of overridden `stderr`.

closes #91

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -303,7 +303,32 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
         if self.displayhook_class:
             displayhook_factory = import_item(str(self.displayhook_class))
             sys.displayhook = displayhook_factory(self.session, self.iopub_socket)
+        
+        self.patch_io()
+    
+    def patch_io(self):
+        """Patch important libraries that can't handle sys.stdout forwarding"""
+        try:
+            import faulthandler
+        except ImportError:
+            pass
+        else:
+            faulthandler_enable = faulthandler.enable
+            faulthandler_register = faulthandler.register
+            # Warning: this is a monkeypatch of `faulthandler.enable`, watch for possible
+            # updates to the upstream API and update accordingly (up-to-date as of Python 3.5):
+            # https://docs.python.org/3/library/faulthandler.html#faulthandler.enable
 
+            # change default file to __stderr__ from forwarded stderr
+            def enable(file=sys.__stderr__, all_threads=True, **kwargs):
+                return faulthandler_enable(file=file, all_threads=all_threads, **kwargs)
+
+            def register(signum, file=sys.__stderr__, all_threads=True, chain=False, **kwargs):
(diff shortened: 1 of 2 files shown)


Carreau:
Wouldn't that fail on Python 2 that have no fault handler module ?

minrk:
Yup, forgot about Python 2.

Carreau:
+1.

fperez:
The patch as it stands right now looks good, I just wonder if it wouldn't be a good idea to add a small test that makes the actual `faultahndler.enable()` call, to increase our chances of catching any errors in case the underlying API evolves in the future.  Since we're basically adding a pass-through, our patch is sensitive to changes in that API.  

Though given Python's lack of formal interface description and testing, we can't really catch forward changes, so perhaps there isn't that much value to it. They are unlikely to introduce backwards-incompatible changes, and short of adding a signature-inspection warning using `inspect`, this may be of little value.

Thoughts?

minrk:
I thought about making it `*args, **kwargs` so it would be forward-compatible, but then the logic for handling file as positional or kwarg seemed like more complexity than it should have. I'm not sure how a test would cover the future change, though, since it would be a backward-compatible addition that's most likely to be missed, and a test wouldn't catch that.

fperez:
Yes, that was exactly my concern, which is why I'm wondering if adding a test would be kind of a pointless exercise in putting in tests for their own sake.

Maybe we should just leave it as-is, and instead just add an explicit note as a comment in the code, just a simple:

```python
# Warning: this is a monkeypatch of `faulthandler.enable`, watch for possible updates to the
# upstream API and update accordingly: 
# https://docs.python.org/3/library/faulthandler.html#faulthandler.enable
```

Better than putting in code that isn't really adding value.

minrk:
Added your comment, a smoke test, and passthrough of any *added* kwargs, which should cover it for a long time, at least.

fperez:
OK, this is as much as we can reasonably do on this. Thanks! Merging.

PR #93, opened 2016-01-07
Handle added but ignored keys in connection file tests

minrk:
jupyter_client will add `kernel_name` to connection files (jupyter/jupyter_client#96)
Instead of checking full-dict equality, check the keys we know/care about,
so the tests aren't sensitive to backward-compatible additions.


Carreau:
+1

PR #101, opened 2016-02-15
Don't show %guiref in banner

takluyver:
We're getting rid of the %guiref magic - see jupyter/qtconsole#98

### DIFF
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -33,7 +33,7 @@ from IPython.core.error import UsageError
 from IPython.core.magics import MacroToEdit, CodeMagics
 from IPython.core.magic import magics_class, line_magic, Magics
 from IPython.core import payloadpage
-from IPython.core.usage import default_gui_banner
+from IPython.core.usage import default_banner
 from IPython.display import display, Javascript
 from ipykernel import (
     get_connection_file, get_connection_info, connect_qtconsole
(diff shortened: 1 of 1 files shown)


ipykernel/zmqshell.py
```diff
@@ -33,7 +33,7 @@
 from IPython.core.magics import MacroToEdit, CodeMagics
 from IPython.core.magic import magics_class, line_magic, Magics
 from IPython.core import payloadpage
-from IPython.core.usage import default_gui_banner
+from IPython.core.usage import default_banner
```
ccordoba12: I'd prefer that we could do the removal of `default_gui_banner` in the next minor IPython release instead of the next major one (given that's such a small change).

So I propose to enclose this in a try/except block, like this:

```python
try:
    from IPython.core.usage import default_gui_banner as default_banner
except ImportError:
    from IPython.core.usage import default_banner
```

takluyver: The try/except is not necessary, because `default_banner` was already defined. The issue is that the IPython release that removes `default_gui_banner` will break compatibility with any ipykernel releases *before* this PR is merged, including at least ipykernel 4.2.2, the current release. So I don't think we should backport that change to a 4.x release.

Why would it be a problem to have a couple of unused variables sitting in IPython, anyway?

ccordoba12: Sorry, I was thinking that a new release of `ipykernel` would be hold until IPython 5.0 is ready.

But you're right, `default_banner` is present in previous IPython releases, so there's no problem with this change :-)

PR #105, opened 2016-02-25
deprecate find_connection_file

### LINKED ISSUE #103: Duplicate function `find_connection_file` with similar API to one in `jupyter_client`
From Gitter [October 21, 2015 5:11 PM](https://gitter.im/jupyter/jupyter?at=5627ff8948fc9e2d6fe877bf):

> So, I was using `find_connection_file` from `jupyter_client.connect`. It turns out there is another implementation under `ipykernel.connect`, which does work. Is there a reason for two different implementations of this function?

Seems like there are two copies of the same function, but they act differently.

If this is intentional, it may be worth finding a different name to highlight the differences and deprecate the old one. If they are meant to behave more similarly, then a bug fix is in order and some discussion should occur about which one to deprecate, if possible, and why.

Related: https://github.com/jupyter/jupyter_client/issues/85

minrk:
It's unused, and the one in jupyter_client should be used instead.

The only case where it might be used is searching for connection files by profile, which no longer makes sense since connection files aren't stored in IPython profiles.

closes #103

### DIFF
--- a/ipykernel/connect.py
+++ b/ipykernel/connect.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import
 import json
 import sys
 from subprocess import Popen, PIPE
+import warnings
 
 from IPython.core.profiledir import ProfileDir
 from IPython.paths import get_ipython_dir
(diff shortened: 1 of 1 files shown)


minrk:
Hang on, we use it a little bit internally, but only within the module. A few more deprecations to add...

minrk:
Okay, should be set.

takluyver:
:+1: don't forget to update the changelog.

minrk:
Updated in #106

PR #106, opened 2016-02-25
note deprecations in 4.3 changelog

minrk:
- ipyparallel
- find_connection_file

cc @takluyver

release 4.3 after this?


takluyver:
It looks like the removal of virtualenv detection (#96) is still in, but not mentioned in the changelog. I'd still prefer not to make that change in 4.x, whether we revert it or do a 4.x branch and backport #105.

minrk:
Yeah, I was going to make the 4.x branch from the parent of that, and backport the few PRs since then.

minrk:
It also depends on whether you want to hold 4.3 for #82, or not.

takluyver:
#82 has been under discussion for long enough that I wouldn't hold 4.3 for it.

Go for it!

minrk:
Released. Will send the appropriate announcements.

Issue #113, opened 2016-03-24
New hook point for messages / message transforms.


dwillmer:
We intend to make a new hook point to allow the installation of hooks on the display publisher.

Current thoughts (largely from @minrk + @jasongrout):

- Update ZMQDisplayPublisher to do a 2-stage message build.
- Pass the message to the transform(s), which return a message or None.
- If a message is returned, `send` it in the usual manner.

This will require:

- Changing the display publisher to have a new method for registering a hook.
- Changes to iostream / ipywidgets / other related repos.

jasongrout:
Also, there are issues with threading. Since the ZMQDisplayPublisher is a global object, installing a message filter there naively will affect messages from *everywhere*. In the case of a context manager, it seems that the correct behavior is to only have the output from the block of code in the current thread. @minrk suggested making the hooks thread local attributes in the ZMQDisplayPublisher object (see https://docs.python.org/2/library/threading.html#threading.local), maybe something like
```python
from threading import local
class ZMQDisplayPublisher(...):
    def __init__(...):
        self.threadlocal = local()

    def register_hook(self, hook):
        if not hasattr(self.threadlocal, 'hooks'):
            self.threadlocal.hooks = []
        self.threadlocal.hooks.append(hook)
```

jasongrout:
Code to touch for making a display hook registration system includes the following. 

* https://github.com/ipython/ipykernel/blob/master/ipykernel/displayhook.py - this is used to 'display' the last output of a computation
* https://github.com/ipython/ipykernel/blob/master/ipykernel/iostream.py - this is used to buffer and package messages for stdout/stderr
* https://github.com/ipython/ipykernel/blob/master/ipykernel/zmqshell.py#L56 - this is used for display_data messages.

@minrk, can you verify this is right?

jasongrout:
Interestingly, since clear_output is implemented by just sending a clear_output message to the front end, this works too:
```python
with output_widget.capture_messages():
    clear_output()
```

minrk:
Yup, that should be right. Also to clarify, implementing this hook should need no changes in any other repos, but the intention is to *use* it in a new implementation of the OutputWidget downstream.

jasongrout:
@dwillmer, @minrk: next steps:

- [ ] A context manager that will register a hook and take over storing the display_pub messages. You may need to flush something when exiting the context manager.
[...]

PR #115, opened 2016-03-25
Message hook in display publisher

dwillmer:
This is the initial work for #113.

- [x] unregister hook.
- [x] unit-tests for both return possibilities from hooks.
- [ ] for the rare cases with multiple hooks, install new ones at the front, rather than the back?

### DIFF
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -161,16 +161,11 @@ class ZMQDisplayPublisher(DisplayPublisher):
                found.
         """
 
-        index = -1
-        for idx, hk in enumerate(self.thread_local.hooks):
-            if hook == hk:
-                index = idx
-
-        if index >= 0:
-            self.thread_local.hooks.remove(index)
+        try:
+            self.thread_local.hooks.remove(hook)
             return True
-
-        return False
+        except ValueError:
+            return False
 
 
 @magics_class
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -66,7 +66,7 @@ class ZMQDisplayPublisher(DisplayPublisher):
 
     # thread_local:
     # An attribute used to ensure the correct output message
-    # is processed. See ipykernel Issue #113 for a discussion.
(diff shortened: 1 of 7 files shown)


dwillmer:
pinging @minrk @jasongrout

ipykernel/zmqshell.py
```diff
@@ -70,6 +78,15 @@ def _flush_streams(self):
         sys.stdout.flush()
         sys.stderr.flush()
 
+    @default('thread_local')
```
minrk: Since I think this is the first PR that uses the traitlets 4.1 API, you may want to bump the version requirement in setup.py.

minrk:
Nice start!

ipykernel/zmqshell.py
dwillmer: Master is already at 4.2 - 

https://github.com/ipython/ipywidgets/blob/bb5092be3c0e75a12293472584a218d9a217c935/setup.py#L108

so i'll just merge on that.

Thanks!

dwillmer:
@minrk @jasongrout now has comprehensive tests over the new hook functionality - your call on whether we merge and iterate or carry on.

ipykernel/zmqshell.py
```diff
@@ -91,6 +125,48 @@ def clear_output(self, wait=False):
...
+        Parameters
+        ----------
+        hook : IPython.core.displayhook.DisplayHook
+               An instance of DisplayHook.
+
+        Returns
+        -------
+        None
+
+        The DisplayHook objects must return a message from
```
minrk: I don't think these are DisplayHook objects since DisplayHook is called on objects, not messages. I think it's any callable with:

```python
def hook(message) => message or None
```

minrk:
Awesome, thanks @dwillmer! I made one comment in-line about DisplayHook, but I think we can merge and move on shortly.

dwillmer:
@minrk - Fixed, thanks!

minrk:
Thanks!

PR #119, opened 2016-04-06
Make help links configurable

SylvainCorlay:


SylvainCorlay:
cc @aburgm

takluyver:
@minrk do the shimmed `.append()` etc. methods on config work to add to default values, or only to previously loaded values? If it's not easy to modify the default, maybe we should leave this non-configurable and add an extra attribute for config to add links.

SylvainCorlay:
@takluyver when you generate the profile, the default values get added to the config file, so it is easy to modify.

minrk:
Yup, append adds to the default. We've also done `extra_foo` configurables for clearly defining additional items, rather than modifying the existing list. I could go either way. Whether we want people to be able to replace the defaults should probably determine whether we add config=True to the existing trait vs a dedicated extra_help_links trait for more user-defined links.

SylvainCorlay:
> Yup, append adds to the default. We've also done extra_foo configurables for clearly defining additional items, rather than modifying the existing list. I could go either way. Whether we want people to be able to replace the defaults should probably determine whether we add config=True to the existing trait vs a dedicated extra_help_links trait for more user-defined links.

Do we want to force the current list of links to users?

SylvainCorlay:
> Do we want to force users to keep the current list of links?

A possible usecase is the one of someone who does not want to use external links...

takluyver:
OK, so long as it's easy to add to the defaults, I'm fine with doing it like this.

minrk:
> Do we want to force the current list of links to users?

That's the question! I don't feel the need to force these links to be always-on, so I'm fine with this PR as-is.

PR #120, opened 2016-04-07
Use traitlets>=4.1 APIs

SylvainCorlay:
`_foo_changed` -> `@observe('foo')` + signature change.
`_bar_default` -> `@default('bar')`

### DIFF
--- a/ipykernel/comm/comm.py
+++ b/ipykernel/comm/comm.py
@@ -12,7 +12,7 @@ from traitlets.config import LoggingConfigurable
 from ipykernel.kernelbase import Kernel
 
 from ipykernel.jsonutil import json_clean
-from traitlets import Instance, Unicode, Bytes, Bool, Dict, Any
+from traitlets import Instance, Unicode, Bytes, Bool, Dict, Any, default
 
 
 class Comm(LoggingConfigurable):
@@ -21,15 +21,22 @@ class Comm(LoggingConfigurable):
     shell = Instance('IPython.core.interactiveshell.InteractiveShellABC',
                      allow_none=True)
     kernel = Instance('ipykernel.kernelbase.Kernel')
-    def _kernel_default(self):
+
+    @default('kernel')
+    def _default_kernel(self):
         if Kernel.initialized():
             return Kernel.instance()
 
     iopub_socket = Any()
-    def _iopub_socket_default(self):
+
+    @default('iopub_socket')
+    def _default_iopub_socket(self):
         return self.kernel.iopub_socket
+
     session = Instance('jupyter_client.session.Session')
-    def _session_default(self):
+
+    @default('session')
+    def _default_session(self):
         if self.kernel is not None:
             return self.kernel.session
 
@@ -38,7 +45,9 @@ class Comm(LoggingConfigurable):
         which to load comm target.""")
 
     topic = Bytes()
-    def _topic_default(self):
+
+    @default('topic')
+    def _default_topic(self):
         return ('comm-%s' % self.comm_id).encode('ascii')
 
     _open_data = Dict(help="data dict, if any, to be included in comm_open")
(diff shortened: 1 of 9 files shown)


SylvainCorlay:
@takluyver there will be less deprecation warnings.

ipykernel/kernelbase.py
```diff
@@ -345,19 +343,19 @@ def send_response(self, stream, msg_or_type, content=None, ident=None,
         """
         return self.session.send(stream, msg_or_type, content, self._parent_header,
                                  ident, buffers, track, header, metadata)
-    
+
     def init_metadata(self, parent):
         """Initialize metadata.
-        
+
```
takluyver: You know what we say about whitespace changes ;-)

ipykernel/comm/comm.py
```diff
@@ -21,15 +21,22 @@ class Comm(LoggingConfigurable):
     shell = Instance('IPython.core.interactiveshell.InteractiveShellABC',
                      allow_none=True)
     kernel = Instance('ipykernel.kernelbase.Kernel')
-    def _kernel_default(self):
+
+    @default('kernel')
+    def _default_kernel(self):
```
takluyver: This feels redundant - since anonymous functions aren't really a thing in Python, we're likely going to keep naming them like this anyway. So the decorator is just duplicating information. If that's how traitlets works now, though, I guess that's what we'll do. :neutral_face:

SylvainCorlay: The new API enabled a lot of simplifications in ipywidgets but for simple things like this, it is true that the net contribution is positive in terms of lines added/removed. ..

ipykernel/kernelbase.py
SylvainCorlay: Yeah but there are *good* whitespace changes and *bad* whitespace changes... ;)

ipykernel/comm/comm.py
minrk: Yeah, I found this when doing the update for JupyterHub. For more advanced uses, the changes are really nice, but for most of our simple use it really is not an improvement. Using traitlets for simple config has become more tedious than it used to be.

ipykernel/kernelapp.py
```diff
@@ -16,17 +16,16 @@
...
-    InteractiveShellApp, shell_flags, shell_aliases
+        InteractiveShellApp, shell_flags, shell_aliases
 )
 from IPython.utils import io
 from ipython_genutils.path import filefind, ensure_dir_exists
-from traitlets import (
-    Any, Instance, Dict, Unicode, Integer, Bool, DottedObjectName, Type,
-)
+from traitlets import (Any, Instance, Dict, Unicode, Integer, Bool,
+                       DottedObjectName, Type, default)
```
minrk: I won't ask you to revert these changes, but FWIW, I prefer the form of all of these before this PR to after.

setup.py
```diff
@@ -81,7 +81,7 @@
 setuptools_args = {}
 install_requires = setuptools_args['install_requires'] = [
     'ipython>=4.0.0',
-    'traitlets>=4.1.0',
+    'traitlets>=4.2.0',
```
minrk: What in this PR changes the minimum version to 4.2? Aren't all these APIs in 4.1?

minrk:
A few comments inline, but I only care about the traitlets 4.2 question before merging.

setup.py
SylvainCorlay: That is right, but traitlets 4.2 has (unrelated) bug fixes and is out there. I can revert that if you want.

ipykernel/kernelapp.py
SylvainCorlay: I can revert this when I fix the other part.

SylvainCorlay:
@minrk I updated/squashed as per your comments.

minrk:
Thanks!

PR #122, opened 2016-04-13
gtk3: Avoid warnings from newer GObject-introspection imports.

pkgw:
Without this patch, newer versions of pygobject3 lead to the following warnings:

```
In [3]: %gui gtk3
In [4]: 
/a/lib/python2.7/site-packages/ipykernel/gui/gtk3embed.py:17: PyGIWarning: Gtk was imported without specifying a version first. Use gi.require_version('Gtk', '3.0') before import to ensure that the right version gets loaded.
  from gi.repository import GObject, Gtk
In [4]: 
```

### DIFF
--- a/ipykernel/gui/gtk3embed.py
+++ b/ipykernel/gui/gtk3embed.py
@@ -14,6 +14,9 @@
 import sys
 
 # Third-party
+import gi
+gi.require_version ('Gdk', '3.0')
+gi.require_version ('Gtk', '3.0')
 from gi.repository import GObject, Gtk
 
 #-----------------------------------------------------------------------------


takluyver:
How specific is this, i.e. does 3.0 really mean the GTK 3.0 API? GTK is now on 3.20.

And how does it interact with a developer who might want to use a different version of Gtk in their own code? E.g. if the developer has already done a `require_version` call for a different version of Gtk or Gdk, will this fail?

takluyver:
Ping @pkgw - we'll probably close this soon to clear the decks unless something happens with it.

pkgw:
Oh, grrr, totally lost track of your questions. Sorry.

I don't know the precise semantics but the version number only speaks to very broad binary compatibility. For the entire lifetime of the Gtk+ 3 series, the API version has been "3.0" (as one can see in the "typelib" files that gobject-introspection uses to describe the API, or the named directories for where binary Gtk+3 plugins go). So, for instance, Gtk+ 3.20 still recommands that you require API version 3.0.

Gtk+ 2.0 didn't use the GObject introspection system, so as far as I know there is literally no other valid choice for this parameter. If there were some hypothetical Gtk V4 you'd need to use something different, but it'd be as big of a change as Gtk 2 vs Gtk 3.

takluyver:
OK. And does it behave reasonably if something else has already called `require_version` for the same API - or will call it after us?

takluyver:
(Since it's supposed to be called before importing Gtk/Gdk, does it throw an error if they've already been imported?)

pkgw:
Oh, yeah, it's sensible about that stuff.

```
13:23 1 >>> import gi
13:23 2 >>> gi.require_version ('Gtk', '3.0')
13:23 3 >>> from gi.repository import Gtk
13:24 4 >>> gi.require_version ('Gtk', '3.0')
13:24 5 >>> from gi.repository import Gtk    
13:24 6 >>> # no complaints here
```

takluyver:
Cool, thanks.

Do we need to do something equivalent in IPython for the event loop integration when we're running in a terminal? The relevant code is here: https://github.com/ipython/ipython/blob/master/IPython/terminal/pt_inputhooks/gtk3.py

pkgw:
I believe you would want to add something there too, yes. My understanding is that any time you're importing a package from `gi.repository`, you should precede that operation with a `require_version`. Again, the idea is that the versions only change for major API changes, such that if your version isn't available, your app is just going to break; and if someone has already loaded a different version, you're not going to be able to use both versions simultaneously.

[Here's the GNOME bug on making the warning message automatic](https://bugzilla.gnome.org/show_bug.cgi?id=727379). It looks like I was wrong and you can in fact load Gtk 2.0 through GObject introspection now; but again, you can't use Gtk 2 and Gtk 3 in the same executable and the APIs are incompatible, so if two pieces of code require the two different versions then there's just going to be no way to use them simultaneously.

takluyver:
Yeah, I'm not too concerned about people trying to use it with Gtk 2, so long as it's not changing on every GTK 3.x release.

PR #123, opened 2016-04-14
Add locking around output buffer

thomasj02:
Please bear with me, as this is my first PR. This new locking in this PR appears to fix https://github.com/ipython/ipython/issues/9397, https://github.com/ipython/ipython/issues/9168

The lock/unlock does slow down output, but it should only be noticeable for notebooks that are writing data very quickly (i.e., the ones most likely to run into this issue).

TBH I don't fully understand the threading model in ipykernel, but the fact that the issue was hard to reproduce consistently (plus the fact that the locking seems to fix it) led me to believe it was a multithreading issue. 

Apologies in advance if I'm totally off-base on how the threading in ipykernel works.

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -9,6 +9,7 @@ import atexit
 import os
 import threading
 import sys
+import threading
 import uuid
 import warnings
 from io import StringIO, UnsupportedOperation
@@ -222,6 +223,7 @@ class OutStream(object):
         self._flush_lock = threading.Lock()
         self._flush_timeout = None
         self._io_loop = pub_thread.io_loop
+        self.buffer_lock = threading.Lock()
         self._new_buffer()
 
     def _is_master_process(self):
@@ -314,7 +316,9 @@ class OutStream(object):
                 string = string.decode(self.encoding, 'replace')
 
             is_child = (not self._is_master_process())
+            self.buffer_lock.acquire(True)
             self._buffer.write(string)
+            self.buffer_lock.release()
             if is_child:
                 # newlines imply flush in subprocesses
                 # mp.Pool cannot be trusted to flush promptly (or ever),
@@ -333,12 +337,15 @@ class OutStream(object):
 
     def _flush_buffer(self):
         """clear the current buffer and return the current buffer data"""
+
+        self.buffer_lock.acquire(True)
         data = u''
         if self._buffer is not None:
             buf = self._buffer
             self._new_buffer()
             data = buf.getvalue()
             buf.close()
+        self.buffer_lock.release()
         return data
 
     def _new_buffer(self):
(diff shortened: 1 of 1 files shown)


ipykernel/iostream.py
```diff
@@ -314,7 +316,9 @@ def write(self, string):
                 string = string.decode(self.encoding, 'replace')
 
             is_child = (not self._is_master_process())
+            self._buffer_lock.acquire(True)
```
minrk: you can use `with self._buffer_lock` in both places

takluyver: To expand a bit: if the code between `acquire()` and `release()` throws an error for some reason, the lock would never be released. You can make sure that some 'after' code is run even if an exception is thrown:

```python
lock.acquire()
try:
    do_stuff()
finally:
    lock.release()
```

But locks support a more convenient way to achieve the same thing:

```python
with lock:
    do_stuff()
```

minrk:
Thanks for the PR! Sorry for the brief comment, hopefully @takluyver clarified it a bit. I think you are right that it's a threading issue, and this is an appropriate fix. We may be able to find a more efficient solution in a later patch that avoids locking on every write, but this is a great place to start.

minrk:
And since this is PR, some further information: you can keep adding and pushing commits to this branch, and it will continue to be updated; then we can merge it when everything is all set.

takluyver:
@minrk in terms of a better approach, would it make sense to pass the data between threads by putting the strings into a `Queue` rather than manipulating `StringIO` instances?

minrk:
That's possible. One thing I don't like about the Queue is it makes reading off the queue require N thread-safe `.get()` calls on the reading side, matching the number of `.put` calls, instead of a single `.getvalue()`. I did a [quick test](https://nbviewer.jupyter.org/gist/minrk/a9bb3bbef9f4e52f79678022fb578356) and using a Queue is about an order of magnitude slower than using StringIO + locking, so I think this solution is a good one.

takluyver:
Interesting. I think the queue is conceptually neater, so I played around a bit more to see if I could make it perform better. I couldn't improve it much using the Queue API, but internally a Queue is based around a deque and a lock, and using those directly, I could get comparable performance to StringIO+Lock:

https://gist.github.com/takluyver/04d8df0cea9e9deb959aba9a3b089f9f

Clearing the deque seems slightly nicer to me than recreating the StringIO on each flush, but we need a lock around in either scenario, so I don't think it's worth the trouble to switch.

(In fact, for this, we don't even need a deque, we could just append to a list - I just tried that, and performance is about the same)

minrk:
Yeah, I would probably use a list+lock rather than a Queue, if that's preferable. I don't see much of a difference, so I'm inclined to let inertia decide, but I wouldn't object to a change like that.

thomasj02:
Updated to use `with` statements for locking

minrk:
@thomasj02 thanks! Welcome to the project.

PR #126, opened 2016-04-26
General Fixes for Comms

dwillmer:
Howdy,

This is a general fix-up across the Comm and CommManager classes - there was quite a lot of redundant/duplicate code which has been cleaned up. There’s also a few performance items here - eg.  multiple identical dict lookups in the same function.

I deliberately haven’t made whitespace changes, but the editor has auto-cleaned up some trailing whitespace.

The biggest changes (which need input from @minrk @takluyver) are the removal of `session` and `iopub_socket` from Comm and CommManager. The tests pass and i can't find any places where these items are used directly, but it could do with some thought from you.

Their use on CommManager was only to be passed into the Comm objects (which happened in 2 places), and their use on Comm objects is restricted to a single method, so it's simpler to just take them from the kernel object that needs passing down the hierarchy anyway. 

ipykernel/comm/manager.py

- [x] Removed lazy_keys function as it’s completely broken. There was only one use which was in a logging statement, and there was an error in the function definition because the single `dikt` argument wasn’t used. This would have caused a runtime exception had the code ever been hit. Have replaced the single use with a direct check of the logging level, which is more obviously correct to readers.
- [x] Removed incorrect comment - `# call, because we store weak refs`. The item wasn’t called, and it’s not storing weak refs.
- [x] Fixed get_comm to not perform the dict lookup twice.
- [x] Removed session from CommManager. It’s not used internally, and the only instantiation of a comm manager i can find is on the kernel, in which case kernel.session is the correct, direct lookup.
- [x] Reduced duplication of code between comm_msg and comm_close.
- [x] Removed duplication of code between register_comm and comm_open. comm_open calls register_comm anyway, so theres no need to set the shell, kernel and iopub_socket twice.
- [x] Removed iopub_socket, because with the previous change it’s only used internally in a single place, and there’s no need for the CommManager to know about it (can be taken directly from `kernel`. 
- [x] Removed shell, because with the previous changes it’s only used when being set on the comm objects, and they can just take it from the kernel object.
- [x] Removed unused imports.
- [x] Removed an EOL semi-colon.

Overall these changes reduce the line count by about  20%, whilst also making the code more efficient, and easier to understand (i.e., smaller methods).

ipykernel/comm/comm.py

- [x] Remove `shell` trait, because it’s only used in one method (and only then if a message callback is set), and there’s no need for the comm objects to know about it when that method can pull it off of `kernel`.
- [x] Remove `iopub` for the same reason as `shell` above.
- [x] Remove `session` for the same reason.

ipykernel/ipykernel.py

- [x] Removed the redundant `shell` and `parent` args from being passed into CommManager.

### DIFF
--- a/ipykernel/comm/manager.py
+++ b/ipykernel/comm/manager.py
@@ -67,7 +67,7 @@ class CommManager(LoggingConfigurable):
         except KeyError:
             self.log.warn("No such comm: %s", comm_id)
             if self.log.isEnabledFor(self.log.DEBUG):
-                self.log.debug("Current comms: %s", self.comms.keys())
+                self.log.debug("Current comms: %s", list(self.comms.keys()))
 
     # Message handlers
     def comm_open(self, stream, ident, msg):
@@ -103,7 +103,7 @@ class CommManager(LoggingConfigurable):
         if self._comm_is_not_valid(comm, comm_id):
             return
 
-        self._safe_handle(msg, comm.handle_msg)
+        self._safe_handle(msg, comm.handle_msg, name='comm_msg')
 
     def comm_close(self, stream, ident, msg):
         """Handler for comm_close messages"""
@@ -112,7 +112,7 @@ class CommManager(LoggingConfigurable):
             return
 
         del self.comms[comm_id]
-        self._safe_handle(msg, comm.handle_close)
+        self._safe_handle(msg, comm.handle_close, name='comm_close')
 
     def _comm_is_not_valid(self, comm, comm_id):
         invalid = comm is None
@@ -125,11 +125,11 @@ class CommManager(LoggingConfigurable):
         comm_id = content['comm_id']
         return (comm_id, self.get_comm(comm_id))
 
-    def _safe_handle(self, msg, callback):
+    def _safe_handle(self, msg, callback, name = ''):
         try:
             callback(msg)
         except Exception:
-            self.log.error('Exception handling ' + callback.__name__ + ' for %s', msg['content']['comm_id'], exc_info=True)
+            self.log.error('Exception handling ' + name + ' for %s', msg['content']['comm_id'], exc_info=True)
 
 
 __all__ = ['CommManager']
--- a/ipykernel/comm/manager.py
+++ b/ipykernel/comm/manager.py
@@ -6,8 +6,6 @@
 import sys
 
 from traitlets.config import LoggingConfigurable
-from IPython.core.prompts import LazyEvaluate
-from IPython.core.getipython import get_ipython
 
 from ipython_genutils.importstring import import_item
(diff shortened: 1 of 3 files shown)


ipykernel/ipkernel.py
```diff
@@ -54,8 +54,7 @@ def __init__(self, **kwargs):
         # TMP - hack while developing
         self.shell._reply_content = None
 
-        self.comm_manager = CommManager(shell=self.shell, parent=self,
-                                        kernel=self)
+        self.comm_manager = CommManager(kernel=self)
```
minrk: This should still pass `parent=self`

ipykernel/comm/manager.py
```diff
@@ -132,32 +99,37 @@ def comm_open(self, stream, ident, msg):
...
+        comm_id = content['comm_id']
+        return (comm_id, self.get_comm(comm_id))
 
+    def _safe_handle(self, msg, callback):
         try:
-            comm.handle_close(msg)
+            callback(msg)
         except Exception:
-            self.log.error("Exception handling comm_close for %s", comm_id, exc_info=True)
+            self.log.error('Exception handling ' + callback.__name__ + ' for %s', msg['content']['comm_id'], exc_info=True)
```
minrk: You can't rely on `callback.__name__` being defined here.

ipykernel/comm/manager.py
```diff
@@ -91,13 +62,12 @@ def get_comm(self, comm_id):
...
+            return self.comms[comm_id]
+        except KeyError:
             self.log.warn("No such comm: %s", comm_id)
-            self.log.debug("Current comms: %s", lazy_keys(self.comms))
-            return
-        # call, because we store weakrefs
-        comm = self.comms[comm_id]
-        return comm
+            if self.log.isEnabledFor(self.log.DEBUG):
+                self.log.debug("Current comms: %s", self.comms.keys())
```
minrk: Needs `list(self.comms.keys())` in order to get the actual keys object, rather than the `dict_keys()` generator.

ipykernel/ipkernel.py
dwillmer: Ok, what's the reason for needing parent set?

If it's required, we should just remove the `kernel` arg and use parent internally then.

minrk: parent passes config. Both parent and kernel should be passed, because it is not required for `parent` to be the kernel, only a Configurable.

minrk:
Thanks, @dwillmer!

dwillmer:
Thanks @minrk

SylvainCorlay:
This breaks widgets. (Comm messages from the front end don't get handled). Should we revert it?

@minrk @jasongrout

PR #128, opened 2016-04-29
Context manager / display hook

dwillmer:


dwillmer:
pinging @jasongrout

jasongrout:
@dwillmer, thanks for all of your work on this. I'll go ahead and take it from here.

ipykernel/hookmanager.py
```diff
@@ -0,0 +1,73 @@
...
+
+from io import StringIO
+import sys
+
+from IPython.core.getipython import get_ipython
+
+from .displayhook import ZMQMessageHook
+
+
+class MessageHookFor(object):
```
minrk: For context managers that are typically treated as functions, I think it's more conventional to use `function_style_name` rather than `ObjectStyleName`, even though it's technically a type.

ipykernel/hookmanager.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    A context manager which takes the name of a message field,
+    and installs a hook to intercept messages of that type.
+
+    Usage
+    -----
+    >>> with MessageHookFor('display_data'):
+            clear_output()
+    """
+    def __init__(self, message_name, parent=None):
```
minrk: Docstring and signature suggest that this is a message for a single type, but it also captures stream messages unconditionally.

ipykernel/hookmanager.py
```diff
@@ -0,0 +1,73 @@
...
+        self._parent = parent
+        self._message_name = message_name
+        self._pub = get_ipython().display_pub
+        self._hook = ZMQMessageHook(message_name, parent.store)
```
minrk: What is `parent` expected to be? Where is this `store` method defined?

ipykernel/hookmanager.py
```diff
@@ -0,0 +1,73 @@
...
+        display publisher.
+        """
+        self._pub.register_hook(self._hook)
+        self._old_clear = self._pub.clear_output
+        self._pub.clear_output = self.clear_output
+
+        self._old_stdout = sys.stdout
+        self._old_stderr = sys.stderr
+        sys.stdout = self._std_buffer
+        sys.stderr = self._std_buffer
```
minrk: Is it deliberate to eliminate the distinction between stdout and stderr?

ipykernel/hookmanager.py
minrk: This also makes parent a required arg, because `parent.store` is accessed unconditionally, so `parent=None` default will never be valid.

minrk:
@jasongrout does your switch to doing this in the frontend mean this PR should be closed?

jasongrout:
Probably. I'm working out the details today, and then will close the outdated PRs.

minrk:
Cool, thanks.

minrk:
@jasongrout what's the state of these? I'd like to do a 4.4 release soon, but I don't want to make a release with this work half-done.

jasongrout:
@minrk - I'm closing this PR since we worked out how to do it nicely from the frontend.

PR #131, opened 2016-05-05
Dummy inprocess Session key

### LINKED ISSUE #130: Empty session keys
In a couple of places, the inprocess kernel code creates `Session` objects with empty keys. These cause a warning in the `Session` constructor now: https://github.com/jupyter/jupyter_client/pull/106

If an empty key is a valid use-case, then https://github.com/jupyter/jupyter_client/pull/106 needs to be rethought. Otherwise, the inprocess kernel code could use a constant non-empty default key, maybe `b'inprocess'`.

rkern:
Use a nonempty dummy `Session` key for inprocess kernels. At present, instantiating a `Session` object with an empty key triggers a security warning that is not relevant to inprocess kernels.

Fixes #130
Fixes jupyter/qtconsole#119

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,12 @@ Changes in IPython kernel
 4.3
 ---
 
+4.3.2
+*****
+
+- Use a nonempty dummy session key for inprocess kernels to avoid security
+  warnings.
+
 4.3.1
 *****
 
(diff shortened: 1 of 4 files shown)


takluyver:
Looks OK to me, but I'll let @minrk have a look too.

PR #133, opened 2016-05-07
simplify Python 3 classifiers

stonebig:


takluyver:
I've said this so many times I feel like a broken record, but I don't think we should use the specific Python 3.x classifiers, because it's presumed to work with new Python 3.x releases, and we'll only forget to update them. I'd rather just use the general 'Python 3' classifier.

stonebig:
recent packages have the full list of Python 3.x classifiers, so you may define the rule for the whole Jupyter thing.

stonebig:
you are right that everything that is Python 3.3.+ only should be marked Python3 only, as Python3.2 is out of maintenance since february 2016:
https://www.python.org/dev/peps/pep-0392/#lifespan

minrk:
👍 to 2.7, 3

takluyver:
Thanks @stonebig

PR #138, opened 2016-05-30
Rewrite comm handling to make it clearer and fix a logic inversion bug

jasongrout:
This reverts part of 91c61ac16559576fa817d1712db7cf3df1142fdf (from https://github.com/ipython/ipykernel/pull/126), which introduced a logic bug which broke handling comm messages, widgets, etc. The logic in the refactoring was complicated enough that the invalid-checking function returned the opposite of what it should have returned.

Since the original code in each function was only about 10 lines, I think it’s more complicated to introduce three new undocumented functions instead of just repeating similar logic twice.


minrk:
Thanks, @jasongrout!

Issue #140, opened 2016-06-02
Why depend on tornado?


dimaqq:
setup.py requires tornado
tornado is never imported
one logger is initialised with name "tornado"

Looks like tornado package is not really needed...

minrk:
tornado provides the eventloop used in the kernel. It ends up being imported by way of zmq, but it is the tornado IOLoop. We could import it directly from tornado to make it clearer.

dimaqq:
I guess pyzmq ought to depend on tornado then, or is that optional because other event loops are possible?

PR #141, opened 2016-06-02
import IOLoop from tornado

minrk:
rather than from zmq

rely on zmq_ioloop.install() to register zmq implementation as default.


takluyver:
What's the advantage of this over using the eventloop from zmq?

takluyver:
Ah, nvm, I read this before seeing your comments on #140.

PR #145, opened 2016-06-23
More Jupyter updates for %connect_info

minrk:
follow-up to #144

- update security_dir check for jupyter paths
- remove no longer relevant mention of profiles

### DIFF
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -50,8 +50,9 @@ from traitlets import (
 )
 from IPython.utils.warn import error
 from ipykernel.displayhook import ZMQShellDisplayHook
-from jupyter_client.session import extract_header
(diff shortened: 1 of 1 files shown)


ipykernel/zmqshell.py
```diff
@@ -331,40 +332,27 @@ def connect_info(self, arg_s):
...
             connection_file = get_connection_file()
             info = get_connection_info(unpack=False)
         except Exception as e:
             error("Could not get connection info: %r" % e)
             return
 
-        # add profile flag for non-default profile
-        profile_flag = "--profile %s" % profile if profile != 'default' else ""
-
         # if it's in the security dir, truncate to basename
```
takluyver: Comment to update

minrk: Thanks, updated

PR #146, opened 2016-06-28
Remove colors_force flag

takluyver:
See ipython/ipython#9673.

This will disable colours if ipykernel is installed with IPython <5, so I'm bumping the IPython requirement in setup.py. So this probably can't be merged until after IPython 5 is released.

### DIFF
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -431,7 +431,6 @@ class ZMQInteractiveShell(InteractiveShell):
     # Override the traitlet in the parent class, because there's no point using
     # readline for the kernel. Can be removed when the readline code is moved
     # to the terminal frontend.
-    colors_force = CBool(True)
     readline_use = CBool(False)
     # autoindent has no meaning in a zmqshell, and attempting to enable it
     # will print a warning in the absence of readline.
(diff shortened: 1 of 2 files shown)


minrk:
Bumping minimum IPython to 5.0 probably means we have to bump ipykernel to 5 as well. Does removing this really mean that ipykernel doesn't work anymore with IPython 4.x?

takluyver:
It can be used, but tracebacks and oinspect output will show up without colour if it can't load readline. We can postpone this change until we're ready to do it, I was just finding all uses of `colors_force` as one task.

minrk:
What happens without this PR, using IPython 5?

takluyver:
It should be totally fine without this PR. The extra attribute will simply be ignored once ipython/ipython#9673 is merged.

minrk:
Gotcha, thanks. Then maybe add a comment indicating that it can be removed once we've bumped minimum IPython to 5.0 instead? I would prefer a stronger reason to bump the major version of the dependency (new APIs, support problems, etc.)

Carreau:
Do we **want** to require IPython >= 5.0, that seem annoying if we want to checkout an old version temporarily. Can't we make a conditional overwrite depending on the IPython version ? (Or am I missing something) ?

takluyver:
This is just a cleanup of a bit of code that's no longer needed. We can postpone it indefinitely until there's a good reason for ipykernel to require ipython 5.

Carreau:
@takluyver care to rebase ? I can do it, but you submitted the PR before "Allow Edits from maintainers".

Carreau:
Rebased as #345

PR #149, opened 2016-07-01
Remove thread check for comm sends

minrk:
IOPub is threadsafe as of 4.3

cf https://github.com/ipython/ipywidgets/issues/642


Carreau:
You can remove the threading import I think.

Carreau:
`IOLoop` as well. But I'll do it in a separate PR. 
Thanks !

Issue #143, opened 2016-06-14
IPython magic UsageErrors do not set status:error on execute_reply


takluyver:
And consequently they don't stop execution if you've queued several cells to run. Thanks to @nholtz for reporting this on the mailing list.

minrk:
I *think* this is something to fix in IPython, since that's what's swallowing the UsageError and printing instead of setting an exception state.

takluyver:
Could be either. I was thinking of overriding `show_usage_error()` on the `ZMQInteractiveShell` class here to track the error. This is like what we already do with `_showtraceback()`.

minrk:
We should do away with the `_showtraceback` hack altogether, and get an official mechanism for "this run_cell failed"

takluyver:
`run_cell` returns an `ExecutionResult` instance now, and `result.success` should be a boolean indicating if it succeeded. Hopefully that provides a good way of doing this.

takluyver:
(I'm just having a go at this - I'm pointing this out in case we're duplicating effort right now ;-)

PR #152, opened 2016-07-06
Simplify detection of errors for execute_reply messages

takluyver:
Closes gh-143

I forget when we introduced ExecutionResult to IPython, but this requires a version of IPython with that machinery.

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -5,7 +5,7 @@ import sys
 import traceback
 
 from IPython.core import release
-from ipython_genutils.py3compat import builtin_mod, PY3
+from ipython_genutils.py3compat import builtin_mod, PY3, unicode_type, safe_unicode
 from IPython.utils.tokenutil import token_at_cursor, line_at_cursor
 from traitlets import Instance, Type, Any, List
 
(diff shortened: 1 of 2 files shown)


ipykernel/ipkernel.py
```diff
@@ -348,25 +342,23 @@ def do_apply(self, content, bufs, msg_id, reply_metadata):
...
-                e_info = dict(engine_uuid=self.ident, engine_id=self.int_id, method='apply')
-                reply_content['engine_info'] = e_info
+            reply_content = {
+                u'traceback': shell._last_traceback or [],
+                u'ename': unicode_type(type(e).__name__),
+                u'evalue': safe_unicode(e),
+            }
+            # FIXME: deprecate piece for ipyparallel:
+            e_info = dict(engine_uuid=self.ident, engine_id=self.int_id, method='apply')
+            reply_content['engine_info'] = e_info
```
takluyver: @minrk this is in the handler for `apply` messages for ipyparallel - can you check if it makes sense?

minrk: Yeah, should be fine

minrk:
Looks like ExecutionResult is in 4.0, which is already the minimum supported version.

minrk:
Nice! I didn't know about ExecutionResult.

minrk:
I didn't think about the fact that ipyparallel would have its own copy of the `_reply_content` hack, since the `apply` implementation in ipykernel is all deprecated. This means that ipykernel 4.4 won't work with ipyparallel ∈ [5,5.2). I don't think it's a huge deal, as it's a shared hack and I'll release ipyparallel this week, too. Just another reason to do away with the hack altogether.

PR #156, opened 2016-07-09
Time tests.

Carreau:
Just to have an idea of what takes the most time, in case we ever want
to optimise that or, if one day a test is slow figure out if it **was**
slow before.
## 

I've a pull request upstream  that allow to export the data to Json if we ever want to upload and have something similar to ASV at some point.


minrk:
Sure. With nose deprecated, there's a good chance we'll move to pytest here before too long.

PR #159, opened 2016-07-14
Make inline the default matplotlib backend

minrk:
uses recently discovered (c/o @weathergod) MPLBACKEND env (matplotlib > 1.5) to make inline the 'default' backend, for users who don't call `matplotlib.use()` or `%matplotlib` prior to plotting.

Trigger extra IPython inline setup when inline backend is loaded. We can't call the simpler `ip.enable_matplotlib()` here because this will be triggered during pyplot import, and enable_matplotlib() imports pyplot which isn't ready yet.

This PR is a sibling of https://github.com/matplotlib/matplotlib/pull/6734

With both PRs, users importing matplotlib and using it without calling `%matplotlib <anything>` should get reasonable behavior in all of:

- terminal IPython (including 5, where GUI eventloops won't hang anymore)
- qtconsole (inline backend default, gui backends won't hang)
- notebook (same)

Either PR independently solves the hang in notebooks, but together they should do what we have been wanting for a long time.
This should remove the need for us to do anything with an import hook (#82).

SciPy hallway discussions FTW!

cc @tacaswell @mdboom @takluyver @fperez

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,21 @@
 Changes in IPython kernel
 =========================
 
+4.4
+---
+
+4.4.0
+*****
+
+`4.4.0 on GitHub <https://github.com/ipython/ipykernel/milestones/4.4>`__
+
+- Use `MPLBACKEND`_ environment variable to tell matplotlib >= 1.5 use use the inline backend by default.
+  This is only done if MPLBACKEND is not already set and no backend has been explicitly loaded,
+  so setting ``MPLBACKEND=Qt4Agg`` or calling ``%matplotlib notebook`` or ``matplotlib.use('Agg')``
+  will take precedence.
+
+.. _MPLBACKEND: http://matplotlib.org/devel/coding_guide.html?highlight=mplbackend#developing-a-new-backend
+
 4.3
 ---
 
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -368,6 +368,13 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
     def init_gui_pylab(self):
         """Enable GUI event loop integration, taking pylab into account."""
 
+        # Register inline backend as default
+        # this is higher priority than matplotlibrc,
+        # but lower priority than anything else (mpl.use() for instance).
+        # This only affects matplotlib >= 1.5
+        if not os.environ.get('MPLBACKEND'):
+            os.environ['MPLBACKEND'] = 'module://ipykernel.pylab.backend_inline'
+
         # Provide a wrapper for :meth:`InteractiveShellApp.init_gui_pylab`
         # to ensure that any exception is printed straight to stderr.
         # Normally _showtraceback associates the reply with an execution,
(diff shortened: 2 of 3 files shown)


WeatherGod:
This gets a +1 from me, at least.

fperez:
Looks good, but before merging, there should be a quick note in the docs about `$MPLBACKEND` and our suggested best practice with this. It should clarify:

- what the values should look like: is it just 'inline' or does it have to be a fully qualified name like `module://ipykernel.pylab.backend_inline`?

- that an explicit `%matplotlib foo` call still overrides `$MPLBACKEND` for users who want to change (e.g. having an inline default but opening the GUI one at the terminal).

- how we suggest folks author notebooks: I still think it's best to be explicit in notebooks meant to be shared, so they run equally in all environments irrespective of each user's choice of `$MPLBACKEND` value.

But I love the improvement!

WeatherGod:
The MPLBACKEND isn't being created for ipython. It already existed and is
documented in matplotlib. The idea is that ipython is hooking into it for
this single purpose of activating its inline backend if no other backend is
explicitly stated. There is no change to users, as far as I can see.

fperez:
Got it. At the very least our docs should mention that with a link to the relevant MPL docs... Users may enter only via our docs, so we want to let them know how we honor this, even if the full details are elsewhere. It's also so it goes into our release notes next time...

WeatherGod:
True... there is
http://matplotlib.org/faq/environment_variables_faq.html?highlight=mplbackend#envvar-MPLBACKEND
or possibly better:
http://matplotlib.org/devel/coding_guide.html?highlight=mplbackend#developing-a-new-backend

takluyver:
Awesome, nice find!

takluyver:
@minrk Do you want to add a whatsnew entry, then I think we can merge this.

tacaswell:
This should probably be configurable so ipympl can over ride this?

minrk:
@tacaswell it is configurable, via the MPLBACKEND environment variable. This only sets that if it's not already set. Do you think there should be another setting somewhere?

minrk:
Sorry for the delay, I'm still catching up on email backlog from SciPy. Added note to changelog about it, linking to MPLBACKEND docs.

@fperez I don't think our recommendations should change. `%matplotlib inline` is explicit and preferable, and still required on matplotlib < 1.5. The MPLBACKEND environment variable sits at a very low priority, so *any* mechanism by which the user might specify their preference will have higher priority:

- explicitly set MPLBACKEND environment variable
- any call to `matplotlib.use('backend')`
- any call to `%matplotlib [whatever]`

The only thing this has higher priority than is `.matplotlibrc`.

tacaswell:
Sorry, I was a bit confused.

The ENV only has an effect if it is set before matplotlib is imported for the first time.  I am not super familiar with the kernel codebase, but it looks like is always true here.

What `ipympl` is doing is using `mpl.use` (https://github.com/matplotlib/jupyter-matplotlib/blob/master/ipympl/__init__.py).  I think this is the right thing to avoid subtle import order issues and the semantics of 'you have imported ipympl, you probably want to use ipympl' seems reasonable  `mpl.use` can be used as many times as you want before `matplotlib.backends` is imported (as that is what we check in `use` to see if it safe).  This works well if there is only one extension, but I can imagine there being multiple packages that are doing the same thing.

As a side note,  we are very slowly starting push the pyplot related state all the way into `pyplot` (but that is only on the master branch at the moment).

minrk:
I think this is ready to go, and then we can release 4.4. Anything else to do/discuss here?

takluyver:
Looks good to me

Issue #161, opened 2016-07-19
Invalid displayhook for other kernels


takluyver:
KernelApp sets `sys.displayhook` to `ZMQDisplayhook`, which sends `execute_result` messages without an execution_count, crashing jupyter_console.

We don't see this in our own code because ZMQInteractiveShell then overrides it with `ZMQShellDisplayHook`, which works correctly. But this is an issue e.g. when running the xonsh kernel.

scopatz:
Thanks for hunting this down!

takluyver:
#162 should fix it.

PR #162, opened 2016-07-19
Ensure basic displayhook sends valid execute_result messages

takluyver:
This fixes a crash I was seeing with the xonsh kernel.

Closes gh-161


scopatz:
Thanks for putting this fix in @takluyver!

PR #165, opened 2016-07-28
Add a --profile flag to kernel install

johanfforsberg:
E.g.
```
$ ipython kernel install --name=mypython --profile=myprofile --user
```
The idea is to be able to configure the kernel to always use a given IPython profile. This could be pretty useful for creating custom versions of the ipykernel.

I realised that there was already almost support for this in the code, I just added a new argument and hooked it up to the internals.

### DIFF
--- a/ipykernel/kernelspec.py
+++ b/ipykernel/kernelspec.py
@@ -44,22 +44,22 @@ def make_ipkernel_cmd(mod='ipykernel', executable=None, extra_arguments=None, **
     if executable is None:
         executable = sys.executable
     extra_arguments = extra_arguments or []
-    arguments = [ executable, '-m', mod, '-f', '{connection_file}' ]
+    arguments = [executable, '-m', mod, '-f', '{connection_file}']
     arguments.extend(extra_arguments)
 
     return arguments
 
 
-def get_kernel_dict():
+def get_kernel_dict(extra_arguments):
     """Construct dict for kernel.json"""
     return {
-        'argv': make_ipkernel_cmd(),
+        'argv': make_ipkernel_cmd(extra_arguments=extra_arguments),
         'display_name': 'Python %i' % sys.version_info[0],
         'language': 'python',
     }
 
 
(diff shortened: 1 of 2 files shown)


ipykernel/kernelspec.py
```diff
@@ -82,7 +83,8 @@ def write_kernel_spec(path=None, overrides=None):
     return path
 
 
-def install(kernel_spec_manager=None, user=False, kernel_name=KERNEL_NAME, display_name=None, prefix=None):
+def install(kernel_spec_manager=None, user=False, kernel_name=KERNEL_NAME, display_name=None,
+            profile=None, prefix=None):
```
minrk: Can you add profile on the end instead of just before prefix, so that it's a backward-compatible patch in the extremely unlikely event of people passing all arguments positionally?

ipykernel/kernelspec.py
```diff
@@ -118,9 +122,13 @@ def install(kernel_spec_manager=None, user=False, kernel_name=KERNEL_NAME, displ
         overrides = dict(display_name=display_name)
     else:
         overrides = None
-    path = write_kernel_spec(overrides=overrides)
-    dest = kernel_spec_manager.install_kernel_spec(path,
-        kernel_name=kernel_name, user=user, prefix=prefix)
+    if profile:
+        extra_arguments = ["--profile", profile]
```
minrk: If profile is given, this should probably be reflected in display_name if it has not been specified manually. Something like `Python X [profile=Y]`, perhaps.

johanfforsberg: Good idea, maybe it would be enough with e.g. "Python X [Y]" to make it more compact?

minrk: I don't think the display_name horizontal space is so precious that `profile=` should be omitted, but feel free to try both and see how it feels. I don't think [name] is sufficiently clear, since it could be indicating other things, such as conda/virtual env, etc.

minrk:
Great, thanks!

PR #171, opened 2016-08-03
ensure status is included in all replies

minrk:
according to the spec

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -205,7 +205,7 @@ class IPythonKernel(KernelBase):
         if res.success:
             reply_content[u'status'] = u'ok'
         elif isinstance(err, KeyboardInterrupt):
-            reply_content[u'status'] = u'abort'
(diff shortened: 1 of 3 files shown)


ipykernel/tests/utils.py
```diff
@@ -40,7 +40,7 @@ def start_new_kernel(**kwargs):
         stdout = nose.iptest_stdstreams_fileno()
     except AttributeError:
         stdout = open(os.devnull)
-    kwargs.update(dict(stdout=stdout, stderr=STDOUT))
+    # kwargs.update(dict(stdout=stdout, stderr=STDOUT))
```
Carreau: Forgotten debug, or do you want to actively remove it ?

Carreau:
+1 otherwise.

ipykernel/tests/utils.py
minrk: Sorry, good catch. Fixed.

PR #172, opened 2016-08-08
Release notes for 4.4

minrk:
We should be ready to release 4.4 once #159 and #171 are merged.


takluyver:
Sorry, this now conflicts because you also added a changelog entry in #159.

Is it also worth mentioning that the `NoOpContext` bug is fixed, by #148?

minrk:
Rebased and added a NoOpContext mention.

minrk:
Thanks! I'll cut the release.

minrk:
4.4 released

takluyver:
Thanks :-)

PR #174, opened 2016-08-09
Workaround circular imports in pylabtools on Python 2

### LINKED ISSUE #173: Import of matplotlib.pyplot fails on Python 2
After upgrading from `4.3.1` to `4.4.0` under Win7 (same in `jupyter notebook`):

```python
C:\Users\User>jupyter console
Jupyter console 5.0.0

Python 2.7.12 (v2.7.12:d33e0cf91556, Jun 27 2016, 15:24:40) [MSC v.1500 64 bit (
MD64)]
Type "copyright", "credits" or "license" for more information.

IPython 5.0.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.



In [1]: import matplotlib.pyplot
ImportError                               Traceback (most recent call last)
<ipython-input-1-6f467123fe04> in <module>()

(...)

c:\program files\python27\lib\site-packages\ipykernel\pylab\backend_inline.py in
_enable_matplotlib_integration()
    152     backend = get_backend()
    153     if ip and backend == 'module://%s' % __name__:
--> 154         configure_inline_support(ip, backend)
    155
    156 _enable_matplotlib_integration()

c:\program files\python27\lib\site-packages\IPython\core\pylabtools.pyc in confi
ure_inline_support(shell, backend)
    359     except ImportError:
    360         return
--> 361     from matplotlib import pyplot
    362
    363     cfg = InlineBackend.instance(parent=shell)

ImportError: cannot import name pyplot
```

minrk:
Calling `pylabtools.configure_inline_support` on import of the inline backend causes circular imports on Python 2 due to an unnecessary import of pyplot in that function (fixed by ipython/ipython#9850).

This works around that problem by scheduling the configure call to happen at the end of the IPython cell if it fails. The downside of this is that inline figure display formatters will not be registered in the cell containing the import. I'm not *too* bothered by this, because it only affects explicit `display(fig)` calls, not anything triggered by matplotlib's own `show` or implicit draw_if_interactive that provide most figure display. Plus, IPython 5.1 will fix it.

I can come up with a more resilient fix that *moves* much of the inline setup to this repo in the long run, but I think this is sufficient for 4.4.1.

closes #173

cc @takluyver @xflr6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,10 @@ before_install:
 install:
     - pip install -f travis-wheels/wheelhouse --pre -e . codecov nose_warnings_filters
     - pip install -f travis-wheels/wheelhouse ipykernel[test] nose-timer
+    - |
+      if [[ "$TRAVIS_PYTHON_VERSION" != "3.3" ]]; then
+        pip install matplotlib
+      fi
     - python -c 'import ipykernel.kernelspec; ipykernel.kernelspec.install(user=True)'
 script:
     - nosetests --with-coverage --with-timer --cover-package  ipykernel ipykernel
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,14 @@ Changes in IPython kernel
 4.4
 ---
 
+4.4.1
+*****
+
+`4.4.1 on GitHub <https://github.com/ipython/ipykernel/milestones/4.4.1>`__
+
+- Fix circular import of matplotlib on Python 2 caused by the inline backend changes in 4.4.0.
+
+
 4.4.0
 *****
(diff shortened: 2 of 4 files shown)


takluyver:
:+1:

minrk:
@takluyver should we merge and make 4.4.1 straight away?

takluyver:
I think that's probably a good idea. Thanks for the quick fix.

minrk:
Thanks, I'll push it now.

PR #178, opened 2016-08-13
Added missing `errors` attribute to OutStream (fixes #177)

### LINKED ISSUE #177: OutStream miss `errors` attribute
Some tools (for example, [`distutils`](https://github.com/python/cpython/blob/14cac82b87eb499a4bee71e667834a7c15f24739/Lib/distutils/log.py#L30)) expects `sys.stdout` / `sys.stderr` having `.errors` attribute, but current `ipykernel.iostream.OutStream` doesn't have it. Is it intended?

Run this in Jupyter Notebook for reproduction:

```python
import sys
sys.stdout.errors
```

The original issue I hit with `distutils` is related to Cython `pyximport` on-the-fly compilation.

frol:
Ref #177

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -205,6 +205,8 @@ class OutStream(object):
             warnings.warn("pipe argument to OutStream is deprecated and ignored",
                 DeprecationWarning)
         self.encoding = 'UTF-8'
+        # This is necessary for compatibility with Python built-in streams
+        self.errors = None
         self.session = session
         if not isinstance(pub_thread, IOPubThread):
             # Backward-compat: given socket, not thread. Wrap in a thread.


takluyver:
Thanks, I think this looks fine. Encoding and errors don't really make sense for a stream object that isn't encoding text to bytes, but I don't see any issue with having them there for other things that expect them.

I'll give others a day or so to have a look at this, and we can merge if no-one objects.

minrk:
If we subclass TextIOWrapper, we should avoid any missing attributes in the future. Is there a reason not to do that?

minrk:
(The default on sys.stdout is 'strict', which I think maps onto our behavior, so maybe use that instead of None)

takluyver:
Good point, though I think the appropriate class to inherit from would be [TextIOBase](https://docs.python.org/3/library/io.html#io.TextIOBase), since we're not wrapping a binary stream.

minrk:
Good call.

PR #182, opened 2016-08-18
Add support for ipympl backend

jenshnielsen:
Goes with https://github.com/ipython/ipython/pull/9885


minrk:
We should have an easier way to do this than registering one by one "this backend has no eventloop integration"

PR #184, opened 2016-08-21
comm.shell is an IPython specific; wrapper kernels use this too

dsblank:
I'm not sure how long this has been broken for wrapper-kernels, but this causes problems now that we have some more complex displays working (like matplotlib plots).

Here in Comm, shell is a detail from IPython kernels, but not wrapper kernels. I tried to define wrapper-kernel.shell = None, but that causes other issues.


dsblank:
I may have described this slightly wrong: I think we can work around this if you don't want to make this change. The problem stems when a wrapper kernel (metakernel) tries to use IPython.display.display.

dsblank:
Nevermind. I'll close this. We have a solution on our end:

https://github.com/Calysto/metakernel/commit/c71126cc500b5509f2ce2171b96dafdebde162e5

PR #190, opened 2016-09-13
KeyboardInterrupt is not abort

minrk:
abort is a separate action, indicating that the request was not processed, not that a particular error was raised.

Regression in 4.4.

cf https://github.com/jupyter/qtconsole/pull/141

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -204,8 +204,6 @@ class IPythonKernel(KernelBase):
 
         if res.success:
             reply_content[u'status'] = u'ok'
-        elif isinstance(err, KeyboardInterrupt):
-            reply_content[u'status'] = u'aborted'
         else:
             reply_content[u'status'] = u'error'
 


takluyver:
Not convinced - I'd consider it an implementation detail of Python that interrupting execution is handled with an exception, and treat that as something separate from an error. I wrote the equivalent code in IRkernel assuming that interrupting execution should set `status='abort'` (though whether that actually works is another question).

minrk:
We should clarify the docs then, because this is not correct behavior as the message spec is designed. Abort should only occur if the message has been aborted via other means, and the request never processed (this functionality was added for IPython parallel, which allows aborting requests, and should never occur in the normal behavior of the IPython kernel).

takluyver:
Do we know what other kernels have implemented?

minrk:
I'm looking around. IHaskell handles 'abort' on the client side, but doesn't emit it from the kernel. IJulia doesn't emit it either, nor do IPython (prior to 4.4), ijavacript, iruby. Toree implements it, but I think it's also a client-side-only spec-completeness thing like IHaskell, but I don't understand the Scala code well enough to say outright.

I dug through the history, and did find text saying that "status='abort'" is for execution aborted by a signal, so apparently I was wrong about that at least in terms of initial intent in 2012, even though it was never implemented in IPython, and the abort that was eventually implemented in IPython was quite different.

At least in the IPython case, I think it's not correct to set status=abort if an exception of any kind is raised. I'm not quite sure what the most backward-compatible clarification in the docs would be, though.

minrk:
jupyter/jupyter_client#200 clarifies that `status='abort'` should only happen when there is no Exception (including KeyboardInterrupt). What do you think about that?

PR #191, opened 2016-09-18
Test on nightly and allow failures

Carreau:


takluyver:
I'm not sure how useful this is - if we can test on the 3.6 betas as opposed to nightly, I'd say we can just add that as a regular test, without `allow_failure`.

Carreau:
Linking to my response on the flit repo: 

https://github.com/takluyver/flit/pull/90#issuecomment-247859533

PR #199, opened 2016-10-21
remove locks from stdout

### LINKED ISSUE #198: locks in stdout
The addition of the IOPub thread in 4.3 for async threaded output uses locks in a few places to ensure threadsafety. This generally works, but causes a problem in one particular case: interactive debugging. The reason being that stepping through an interactive debug session introduces concurrency between a non-running thread (the debugging thread) and the debugger itself, allowing for the following sequence of events:

1. debugged thread\* acquires lock on stdout buffer, pauses waiting for next step
2. debugger UI tries to draw the current frame to stdout, and waits for debugged thread to release the lock, which never happens.

Which ultimately means that you can't debug anything with a print statement (post-mortem debugging is not affected).

I believe that I can write a lock-free implementation, but I might need to use inproc zmq sockets to do so. I will see if I can do it with plain `os.pipe` as well, but I'm not sure that I can as easily.

cc @garth-wells who reported the issue.

minrk:
use inproc push/pull to send events to the IOPub thread

- moves all read/write on the buffer to the IOPub thread
- event callbacks are stored in a dict by ID, the ID is sent on the socket to signal event procesing on the thread.
- removes all Locks
- adds IOPubThread.schedule method for the call-this-in-the-thread action,
  since events should process immediately in cases (e.g. subprocesses) where
  the thread is no longer alive.

closes #198

### DIFF
--- a/ipykernel/inprocess/socket.py
+++ b/ipykernel/inprocess/socket.py
@@ -31,7 +31,8 @@ class SocketABC(with_metaclass(abc.ABCMeta, object)):
     
     @classmethod
     def register(cls, other_cls):
-        warnings.warn("SocketABC is deprecated.", DeprecationWarning)
+        if other_cls is not DummySocket:
+            warnings.warn("SocketABC is deprecated.", DeprecationWarning)
         abc.ABCMeta.register(cls, other_cls)
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Carreau:
kicked travis.

ipykernel/inprocess/socket.py
```diff
@@ -31,7 +31,8 @@ def send_multipart(self, msg_parts, flags=0, copy=True, track=False):
     
     @classmethod
     def register(cls, other_cls):
-        warnings.warn("SocketABC is deprecated.", DeprecationWarning)
+        if other_cls is not DummySocket:
+            warnings.warn("SocketABC is deprecated.", DeprecationWarning)
```
Carreau: `stacklevel=2` ? it was not there before so I'm fine not changing that now.  Also since when is it deprecated (4.5.0 AFAICT).

PR #204, opened 2016-11-08
support transient field, update_display_data

minrk:
API addition:

`DisplayPublisher.publish` accepts two new keyword-args: `update: bool, transient: dict` (new args, fully backward-compatible).

Message spec: 

adds third `transient` dict next to data, metadata for data that shouldn't be persisted as part of files.

display_id will be added to this this transient dict by `display(obj, display_id=...)`

IPython PR: ipython/ipython#10048

This is the ipykernel piece of https://github.com/jupyter/jupyter_client/issues/209

cc @rgbkrk

### DIFF
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -91,20 +91,42 @@ class ZMQDisplayPublisher(DisplayPublisher):
             self._thread_local.hooks = []
         return self._thread_local.hooks
 
-    def publish(self, data, metadata=None, source=None):
+    def publish(self, data, metadata=None, source=None, transient=None,
+        update=False,
+    ):
+        """Publish a display-data message
+
+        Parameters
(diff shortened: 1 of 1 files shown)


rgbkrk:
I've got an initial implementation that I'll start writing tests for in nteract and evaluating the implementation with direct action mocks. Looking forward to trying this out with the IPython kernel.

rgbkrk:
What do you think the order of merging and shipping this + https://github.com/ipython/ipython/pull/10048?

minrk:
This one can go first, but I'll make sure that IPython behaves gracefully with ipykernel stable, so release order shouldn't matter too much. The main question that's relevant to both is deciding on the new signature for `DisplayPublisher.publish`, which is a common API in both packages.

This should be good to go now, in fact.

PR #205, opened 2016-11-15
Allow comms to be instanciated outside of a kernel

SylvainCorlay:
We need this to generate pure server-side widget state.


SylvainCorlay:
@minrk the widget sphinx extensions must be able to instantiate widgets without a front-end just to serialize them and insert them in the documentation...

SylvainCorlay:
@jasongrout

SylvainCorlay:
Thanks @minrk

PR #209, opened 2016-11-20
Add a launcher module to ignore the CWD for imports while ipykernel starts

takluyver:
After helping someone who'd shadowed a stdlib module with a file in the cwd (jupyter/help#97), I realised that we could fairly easily make the kernel able to start in such situations.

This PR adds an ipykernel_launcher module which removes the CWD from sys.path and then launches the kernel. There is already code in IPython which adds the CWD back to sys.path before user code runs.

This doesn't get rid of issues with name clashes: modules that we import will shadow modules of the same name in the cwd. So if you wrote math.py, you wouldn't be able to import it. But I think this is easier
for users to understand than the kernel completely failing to start, especially if it's not easy for them to see the logs showing the error.

### DIFF
--- a/ipykernel/kernelspec.py
+++ b/ipykernel/kernelspec.py
@@ -22,7 +22,7 @@ KERNEL_NAME = 'python%i' % sys.version_info[0]
 RESOURCES = pjoin(os.path.dirname(__file__), 'resources')
 
 
-def make_ipkernel_cmd(mod='ipykernel', executable=None, extra_arguments=None, **kw):
+def make_ipkernel_cmd(mod='ipykernel_launcher', executable=None, extra_arguments=None, **kw):
     """Build Popen command list for launching an IPython kernel.
 
     Parameters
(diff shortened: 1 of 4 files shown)


takluyver:
Aside: though we always talk about module shadowing in terms of the standard library, the same thing applies if you clash with a third party module name like tornado or zmq. The stdlib just uses names which you're more likely to clash with accidentally, like math and socket.

ipykernel/kernelspec.py
```diff
@@ -22,7 +22,7 @@
 RESOURCES = pjoin(os.path.dirname(__file__), 'resources')
 
 
-def make_ipkernel_cmd(mod='ipykernel', executable=None, extra_arguments=None, **kw):
+def make_ipkernel_cmd(mod='ipykernel_launcher', executable=None, extra_arguments=None, **kw):
```
minrk: Why add a new module and change default behavior to use it, rather than just changing the behavior of `ipykernel.__main__`?

takluyver: When you run `python -m ipykernel`, it loads `ipykernel/__init__.py` before `ipykernel/__main__.py`. The former already imports a bunch of stuff. Getting in before that happens would involve checks whenever `ipykernel` is imported, which seems like a bad thing.

With a separate module which is only useful as a means to start the kernel, we can check and modify `sys.path` before anything else is imported.

minrk: Gotcha, thanks.

Issue #207, opened 2016-11-16
Tests hang on exit


takluyver:
@minrk I'm trying to debug this failure on Jenkins. I recall you saying it was probably related to threads failing to exit. So far I can reproduce it by manually running:

```
PYTHONHASHSEED=6 nosetests -v ipykernel
```

I don't see it running a specific part of the tests, only running the whole test suite. I can remove some test files and still get the hang, but there are several files where if I remove any of them I 
no longer see the hang.

minrk:
Thanks, I'll try to dig into it.

takluyver:
My last attempt before I wrapped up for the day was an atexit handler to
show which threads were still active. I got a traceback from inside the
threading module.

infinity0:
@takluyver Hi Thomas, long time no see!

> there are several files where if I remove any of them I no longer see the hang.

Could you tell us which ones they are? This has been marked as a [release-critical bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=849220) for the Debian package and currently our plan is to disable the tests. However I'd rather disable parts of the suite rather than the whole suite.

takluyver:
Sorry, got distracted onto other things. I'll try to get back to this, but unfortunately, I cannot remember which specific files I was referring to. It was not a set that had any obvious common factor.

PR #217, opened 2017-01-18
Add 3.6 target as nightly points to 3.7.0a

srinivasreddy:


takluyver:
It looks like #207 is now affecting Travis as well :-(

minrk:
working on the hang now

minrk:
I'm on the trail. It has to do with the IOPubThread interacting with the inprocess kernel. Should have a PR by the end of the night.

PR #218, opened 2017-01-20
make sure IOPubThreads are cleaned up in inprocess tests

### LINKED ISSUE #207: Tests hang on exit

minrk:
I think this closes #207


takluyver:
Thanks Min. Let's see how that goes.

PR #222, opened 2017-01-31
Use the new IPython completer API

Carreau:
Work in progress, I want to see if I can use the metadata fields for frontends to opt-in to some extra informations.

Also how backward compatible would it be to have a metadata field on the `complete_request` ? My idea is for testing would have been to have an "Accept:" like field for the kernel to know whether to compute the new completions.

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -2,7 +2,6 @@
 
 import getpass
 import sys
-import traceback
 
 from IPython.core import release
 from ipython_genutils.py3compat import builtin_mod, PY3, unicode_type, safe_unicode
@@ -12,6 +11,15 @@ from traitlets import Instance, Type, Any, List
 from .comm import CommManager
 from .kernelbase import Kernel as KernelBase
 from .zmqshell import ZMQInteractiveShell
+
+
+try:
+    from IPython.core.completer import rectify_completions, provisionalcompleter
+    _use_experimental_60_completion = True
+except ImportError:
+    _use_experimental_60_completion = False
+
+_EXPERIMENTAL_KEY_NAME = '_jupyter_types_experimental'
 
 
 class IPythonKernel(KernelBase):
@@ -246,6 +254,9 @@ class IPythonKernel(KernelBase):
         return reply_content
 
     def do_complete(self, code, cursor_pos):
+        if _use_experimental_60_completion:
+            return self._experimental_do_complete(code, cursor_pos)
+
         # FIXME: IPython completers currently assume single line,
         # but completion messages give multi-line context
         # For now, extract line from cell, based on cursor_pos:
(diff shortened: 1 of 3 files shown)


minrk:
> Also how backward compatible would it be to have a metadata field on the complete_request ?

Adding a field should be a minor revision, and if things work when that field is empty / missing, it's not a compatibility problem at all.

rgbkrk:
I'm ok with introducing new fields rather than metadata for the final version of this (that can be ignored by some frontends), effectively `matches` and `enhanced_matches`.

The annoying thing I've noticed about our "metadata" fields is that people come to rely on the metadata if it gets merged in as a metadata field, and then it's no longer extraneous metadata -- it's an expected format with it's own versioning.

Carreau:
> I'm ok with introducing new fields rather than metadata for the final version of this (that can be ignored by some frontends), effectively matches and enhanced_matches.
> 
> The annoying thing I've noticed about our "metadata" fields is that people come to rely on the metadata if it gets merged in as a metadata field, and then it's no longer extraneous metadata -- it's an expected format with it's own versioning.


Agreed, but right now adding things into the metadata field allow us to experiment without changing the wireformat and having to deal with the serialisation layer. 

Once we get the fields right it's "easy" to make an enhancement proposal to move them up one level. 

I've push a commit that send these new informations in the metadata field for now. 
We can start bikeshedding on names and necessary fields.

rgbkrk:
Sounds great, thanks Matthias.

minrk:
Sounds like a good plan.

ipykernel/ipkernel.py
```diff
@@ -261,6 +270,42 @@ def do_complete(self, code, cursor_pos):
...
+            completions = list(rectify_completions(code, raw_completions))
+            
+            comps = []
+            for comp in completions:
+                comps.append(dict(
+                            start=comp.start,
+                            end=comp.end,
+                            text=comp.text,
+                            type=comp.type,
+                              ))
```
ellisonbg: Dedent these `))` to the level of `comps`.

ipykernel/ipkernel.py
```diff
@@ -261,6 +270,42 @@ def do_complete(self, code, cursor_pos):
...
+
+        if completions:
+            s = completions[0].start
```
ellisonbg: Seems reasonable to use `completions[0]` but might there ever be a reason to not pick this one?

ipykernel/ipkernel.py
```diff
@@ -261,6 +270,42 @@ def do_complete(self, code, cursor_pos):
...
+            matches = [c.text for c in completions]
+        else:
+            s = cursor_pos
+            e = cursor_pos
+            matches = []
+
+        return {'matches': matches,
+                'cursor_end': e,
+                'cursor_start': s,
+                'metadata': {'_experimental': comps},
```
ellisonbg: The only big question is how to encode this in the message data. I think this approach of putting this into the metadata makes sense. Maybe add a suffix to help a frontend to distinguish if there are multiple experimental extensions:

`_experimental_type_data`

ellisonbg: Some ideas:

* `_jupyter_types_prototype`
* `_jupyter_types_experimental`

ipykernel/ipkernel.py
Carreau: No. `rectify_completion` is made to have all completions have the same beginning and end and pad these potentially ugly characters. I can re-assert that `len(set(c.start for c in completions)) == 1`, but that seem unnecessary.

ellisonbg: Sounds good!

Carreau:
Rebased and comment taken into account.
This will likely also affect a bit completions in the normal notebooks, it there are bugs, please let me know.

Carreau:
Thanks !

rgbkrk:
😮 

🎉 🎉 🎉

PR #223, opened 2017-02-01
install kernelspec in wheels

minrk:
- wheel kernelspec relies on PATH env to find python, which should be okay when installed to sys.prefix, as opposed to system-wide.
- can't use universal wheels anymore, since Python version is in kernelspec name and display name

It would be ideal to get a hardcoded path at install time (https://github.com/pypa/pip/issues/4032), but for envs this should do until that becomes possible.

After this, it should be a lot harder for the native kernelspec to not be installed, which would be ideal. Plus, uninstalling ipykernel will also uninstall the kernelspec.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ __pycache__
 \#*#
 .#*
 .coverage
+
+data_kernelspec
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,7 @@ sudo: false
 install:
     - |
       pip install --upgrade setuptools pip
-      python setup.py bdist_wheel
-      pip install --pre dist/*.whl ipykernel[test]
+      pip install --pre . ipykernel[test]
     - |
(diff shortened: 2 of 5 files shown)


takluyver:
I'm not convinced by relying on PATH - that assumes that people always activate a virtualenv rather than launching scripts inside it directly.

minrk:
We could handle the executable `python` in `jupyter_client` to explicitly mean `sys.executable` (https://github.com/jupyter/jupyter_client/pull/235), which should handle that.

minrk:
Well, crap. I swear I didn't click the merge button, but GitHub interpreted a click in the comment area as a click of the merge button (not sure if it's GitHub's or my browser's fault). I can revert and re-issue the PR if you want, or we can continue to review this PR and I'll open a new one for any changes that should result.

I cleaned up travis.yml a bit to do a 'regular' install rather than an editable one for testing.

takluyver:
Let's carry on without reverting.

cben:
Nice!  It seems this installs the kernel.json `pip install .` but not on `pip install -e .` (which README.md recommends for developement).

PR #224, opened 2017-02-01
Patch PYTHONPATH when running tests

takluyver:
Fixes running tests with packages --user installed.

Prompted by jupyter/notebook#2111

### DIFF
--- a/ipykernel/tests/__init__.py
+++ b/ipykernel/tests/__init__.py
@@ -3,6 +3,7 @@
 
 import os
 import shutil
+import sys
 import tempfile
 
 try:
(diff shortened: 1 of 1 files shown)


Carreau:
+1. 4.5.2 ,4.6 or 5.0 ?

takluyver:
I can never remember how we're using version numbers on different projects. I don't think it's important enough to backport to 4.5.2, but I don't have a strong feeling between 4.6 and 5.0.

Carreau:
> I can never remember how we're using version numbers on different projects. I don't think it's important enough to backport to 4.5.2, but I don't have a strong feeling between 4.6 and 5.0.

No backport it is then.

PR #229, opened 2017-02-13
publish busy/idle for aborted requests

minrk:
follow spec: kernels should send busy/idle messages on IOPub during processing of every request

It's ambiguous whether aborted requests are 'processed', but any code that assumes requests get busy/idle could get tripped up by the absence of these messages.  Since the abort is rare, making this case as ordinary as possible is probably best.


takluyver:
Sounds sensible.

Issue #64, opened 2015-10-05
[IPKernelApp] ERROR | UNKNOWN MESSAGE TYPE: 'comm_open'


scopatz:
Hello all, I am writing a wrapper kernel for xonsh (see scopatz/xonsh#399). We noticed that the following error gets printed to the command line on jupyter v4.0.6, jupyter notebook v4.0.5, and ipykernel v4.0.3.  

```
[IPKernelApp] ERROR | UNKNOWN MESSAGE TYPE: 'comm_open'
```

Note that this error does not get printed if the jupyter notebook is on v4.0.4 and the other two stay the same.  Not really sure what is going on here. It doesn't seem to effect anything on Ubuntu 15.04, but may be effecting things on Windows.

scopatz:
(That is, nothing is working on Windows right now...)

minrk:
It's a little weird that a comm_open message is arriving, but wrapper kernels should safely ignore unhandled messages and move on. Does it work on 4.0.4, or also not work, and the only difference is the unknown message?

scopatz:
It works just fine on v4.0.4.  The only difference is this unknown message, that does seem to be safely ignored.

minrk:
And just upgrading/downgrading the notebook package in the same env is the only change? Bizarre, but at least that makes it a very small set of changes to look through. Are you using conda or pip to do this?

scopatz:
Yes, the upgrade/downgrade was the only change. I used conda to do this.

minrk:
And this is only an issue on Windows? Does an IPython kernel work in the same env? Does it matter what notebook you try to open?

I'm going with downright bonkers for now, but I'll see if I can dig out a comparison of the conda packages tomorrow. I don't see anything in the [repo comparison](https://github.com/jupyter/notebook/compare/4.0.4...4.0.5) that would be relevant, but there have been occasional issues with conda packages that get something amiss.

scopatz:
hmmm yeah.  Nothing here seems to indicate this kind of error message.  

The issue for xonsh is on Windows, but may be unrelated. 

Don't loose too much sleep over this @minrk.

adrian17:
I just noticed it too. Version 4.0.6. If it's not an actual issue, maybe the name should be changed to "warning"?

evhub:
Same error is still appearing under 4.1.2.

dimpase:
We also see it here: https://trac.sagemath.org/ticket/21416

takluyver:
[...]

PR #230, opened 2017-02-24
Make 'Unknown message' log less scary

takluyver:
See gh-64

With optional parts of the spec, and future additions of message types, the kernel will sometimes get a message that it doesn't know how to handle. This is not generally an error, so we don't need the log message to be so shouty about it.


Carreau:
I LIKE IT.

PR #233, opened 2017-03-27
Disable parent-poller if PID is 1

minrk:
PID 1 is special, and we can't treat it the same as we do in other cases. When run in Docker, the parent (notebook server) can be PID 1, which results in IPython kernels exiting immediately on start, which manifests as "Kernel Restarting..." and "Kernel appears to have died". This is caused by the parent poller incorrectly detecting that the parent has exited, and shutting down.

Also adds logging for parent-exit-triggered shutdown, which would have helped figure this out much quicker.

closes https://github.com/jupyterhub/kubespawner/issues/31
related: https://github.com/cassinyio/SwarmSpawner/issues/6

cc @yuvipanda

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -166,7 +166,9 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
             if self.interrupt or self.parent_handle:
                 self.poller = ParentPollerWindows(self.interrupt, self.parent_handle)
         elif self.parent_handle and self.parent_handle != 1:
-            # PID 1 is special. Parent polling doesn't work if ppid == 1 to start with.
+            # PID 1 (init) is special and will never go away,
+            # only be reassigned.
+            # Parent polling doesn't work if ppid == 1 to start with.
             self.poller = ParentPollerUnix()
 
     def _bind_socket(self, s, port):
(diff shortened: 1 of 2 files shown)


minrk:
appveyor is failing due to something weird with a tornado prerelease. Not related to this PR

choldgraf:
someone should write a series of mystery novel titles "Jupyterhub and the mystery of the PID killer"

yuvipanda:
Aaah, I had no idea we had a *parent* poller (not just a child one). This makes sense!

yuvipanda:
I've tested this and it seems to fix the issue!

ipykernel/kernelapp.py
```diff
@@ -165,7 +165,8 @@ def init_poller(self):
         if sys.platform == 'win32':
             if self.interrupt or self.parent_handle:
                 self.poller = ParentPollerWindows(self.interrupt, self.parent_handle)
-        elif self.parent_handle:
+        elif self.parent_handle and self.parent_handle != 1:
+            # PID 1 is special. Parent polling doesn't work if ppid == 1 to start with.
```
willingc: Maybe add: `# PID 1 (system init process) has no parent. Parent...`

minrk:
Thanks! Fixed the appveyor build in #237

PR #234, opened 2017-03-27
inherit IOStream from TextIOBase

### LINKED ISSUE #187: subclass TextIOBase in OutStream
should avoid future AttributeErrors, etc. like #178

minrk:
ensures all expected APIs are present and raise appropriate errors for undefined methods, rather than AttributeError.

closes #187


minrk:
tornado prerelease has confused appveyor

PR #237, opened 2017-03-29
run appveyor build in cmd, not ps

minrk:
No idea if this will actually fix it, but worth a shot.

PS seems to die when there are warning messages, which is odd.

currently triggered when installing tornado 4.5b1


minrk:
This works.

FWIW, it appears that PS sometimes treats *any* output on stderr as an error, so warning messages get treated as failures. I don't care to debug it further, because we can always just use cmd.

PR #241, opened 2017-04-10
Remove calls to `start_event_loop_qt|wx` from IPython.lib.guisupport

### LINKED ISSUE #240: %gui qt5 no longer works in 4.6.0, while working fine in 4.5.2 in QtConsole
It seems that ipykernel 4.6.0 broke Qt5 (non-blocking) event loop integration, so it is not possible to use non-blocking Qt5 UI directly in QtConsole. Following code shows the issue:
```python
%gui qt5
from PyQt5 import QtWidgets
w=QtWidgets.QLabel("Hello world!")
w.show() #<- I can see the label and close the window in 4.5.2, but on 4.6.0 it is frozen
```
Just to be clear, downgrading to 4.5.2 makes `%gui qt5` work as desired again.

minrk:
These are no-ops in IPython >= 5.2 if `get_ipython().active_eventloop` is defined, so the eventloops never start.

Replace them with private functions that are guaranteed to actually run the eventloop,
effectively replicating the `start_event_loop` behavior from IPython < 5.2, without the need for is-it-running checks due to being a public API.

closes #240

cf https://github.com/ipython/ipython/issues/9974 , https://github.com/ipython/ipython/pull/10419

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -72,11 +72,23 @@ def register_integration(*toolkitnames):
     return decorator
 
 
+def _loop_qt(app):
+    """Inner-loop for running the Qt eventloop
+
+    Pulled from guisupport.start_event_loop in IPython < 5.2,
+    since IPython 5.2 only checks `get_ipython().active_eventloop` is defined,
+    rather than if the eventloop is actually running.
+    """
+    app._in_event_loop = True
+    app.exec_()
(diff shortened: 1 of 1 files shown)


jenshnielsen:
Resolves the issue for me, Thanks @minrk

willingc:
Thanks for the update @jenshnielsen. Glad this works.

PR #242, opened 2017-04-10
Changelog for 4.6.1

minrk:
Fixing some regressions seen in 4.6.0. I'd like to get this out pretty soon, since there are a couple of small but annoying regressions in recent releases.

- [x] Qt windows never showing up #241
- [x] deprecation warnings #239

cc @JamiesHQ

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,15 @@ Changes in IPython kernel
 4.6
 ---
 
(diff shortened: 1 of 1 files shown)


willingc:
@minrk #241 looks good to me but Travis seems a bit flaky. Going to let you or @takluyver merge this so there's another set of eyes.

PR #249, opened 2017-06-06
Autocomplete: Expose Jedi signature to frontends

lgeiger:
This adds support for function signatures introduced in https://github.com/ipython/ipython/pull/10507 so we can try them in other frontends.

/cc @Carreau

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -423,7 +423,7 @@ class IPythonKernel(KernelBase):
                             end=comp.end,
                             text=comp.text,
                             type=comp.type,
-                            signature=getattr(comp, 'signature', '')
+                            right_label=getattr(comp, 'signature', ''),
                 ))
 
         if completions:
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -423,6 +423,7 @@ class IPythonKernel(KernelBase):
                             end=comp.end,
                             text=comp.text,
                             type=comp.type,
+                            signature=getattr(comp, 'signature', '')
                 ))
 
         if completions:


ipykernel/ipkernel.py
```diff
@@ -274,21 +274,22 @@ def do_complete(self, code, cursor_pos):
...
-            
+
             comps = []
             for comp in completions:
                 comps.append(dict(
                             start=comp.start,
                             end=comp.end,
                             text=comp.text,
                             type=comp.type,
+                            signature=getattr(comp, 'signature', '')
```
lgeiger: Depending on the IPython version, completions may or may not return the function signature. Therefor I used `getattr` to make sure we stay compatible.

Carreau: +1, one thing I was thinking about is to give that a name more generic name (`right_extra`?) to see how that would extend to other languages. I can't really think of any language where signature is on the left, or maybe language that have `operators` would want to send extra text to display on the left and right ? and call that `extra_left` and `extra_right` ?

lgeiger: Good Point! `signature` probably only makes sense in case of functions.
Atom's [autocomplete-plus](https://github.com/atom/autocomplete-plus/wiki/Provider-API#suggestions) uses the terms `rightLabel` and `leftLabel` so `right_label` would be another option.

minrk: I'm always happy to copying something that already exists, so +1 to `right_label`

ivanov:
Hey there @lgeiger. I'm going through old issues and it seems this one stalled. 

I've just rebased and changed the generic 'signature' to a 'right_label' as per discussion above. 

Do you still have cycles to push forward on this? Either way is fine, I think we just need to document the change before merging, so others can know they can start using this. So that would be the next step (unless, of course, I've missed something entirely and this has been superseded by another avenue of transmitting signatures).

ivanov:
Python 3.9 test failed here, but in a weird timeout - so I'm just gonna try running it again.

krassowski:
Note: LSP uses just `label` which includes both the name of the function and the arguments. I think that returning `text + signature` and calling it `label` is a more future-proof approach as compared to using `right_label`.

While LSP does use `signature` in other contexts, this is definitely a completion `label`; also thinking about other uses of completer which modify the already entered text which is already possible with IPython - e.g. the LaTeX symbols completions, pre-pending text on the left would not be unexpected for me. Think about a user trying to complete `sqrt` but they have not imported it but they have `math` available - in that case the `label` could be `math.sqrt(x)` (and the `text` would be `math.sqrt` with adjusted start/end coordinates). How does that sound?

Carreau:

PR #254, opened 2017-07-27
Support image/gif mimetype

gnestor:
https://github.com/jupyter/notebook/pull/2702 and https://github.com/ipython/ipython/pull/10716 depends on this

### DIFF
--- a/ipykernel/jsonutil.py
+++ b/ipykernel/jsonutil.py
@@ -45,6 +45,9 @@ PNG64 = b'iVBORw0KG'
 JPEG = b'\xff\xd8'
 # front of JPEG base64-encoded
 JPEG64 = b'/9'
+# constants for identifying gif data
+GIF = b'GIF87a'
+GIF89 = b'GIF89a'
 # front of PDF base64-encoded
 PDF64 = b'JVBER'
 
@@ -83,6 +86,13 @@ def encode_images(format_dict):
         if not jpegdata.startswith(JPEG64):
             jpegdata = encodebytes(jpegdata)
         encoded['image/jpeg'] = jpegdata.decode('ascii')
+        
+    gifdata = format_dict.get('image/gif')
+    if isinstance(gifdata, bytes):
+        # make sure we don't double-encode
+        if not gifdata.startswith(GIF) or not gifdata.startswith(GIF89):
+            gifdata = encodebytes(gifdata)
+        encoded['image/gif'] = gifdata.decode('ascii')
 
     pdfdata = format_dict.get('application/pdf')
     if isinstance(pdfdata, bytes):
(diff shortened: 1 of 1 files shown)


gnestor:
@Carreau Any idea why no build?

ipykernel/jsonutil.py
```diff
@@ -83,6 +86,13 @@ def encode_images(format_dict):
         if not jpegdata.startswith(JPEG64):
             jpegdata = encodebytes(jpegdata)
         encoded['image/jpeg'] = jpegdata.decode('ascii')
+        
+    gifdata = format_dict.get('image/gif')
+    if isinstance(gifdata, bytes):
+        # make sure we don't double-encode
+        if not gifdata.startswith(GIF) or not gifdata.startswith(GIF89):
```
takluyver: I don't think this is right. For one thing, it's `not`-ing two mutually exclusive conditions, so I think it will always be True overall. But more importantly, this is meant to be a check whether it's *already* base64 encoded - if you look at all the other formats, it's checking against a base64 encoding of the magic number, not the raw magic number.

There's a handy feature of `.startswith()`: you can give it a tuple to check if any of the strings are prefixes. So I think this should be `if not gifdata.startswith((GIF_64, GIF89_64))`, where the `_64` variables are base64 encoded versions of the magic numbers.

takluyver: When you generate the base64 encoded versions, let's grab a few different GIFs and base64 encode them just to check that the prefixes are accurate, as well.

gnestor: Thanks! Makes sense 👌

ipykernel/jsonutil.py
```diff
@@ -45,6 +45,9 @@
 JPEG = b'\xff\xd8'
 # front of JPEG base64-encoded
 JPEG64 = b'/9'
+# constants for identifying gif data
+GIF_64 = b'R0lGODdh'
+GIF89_64 = b'R0lGODlh'
```
gnestor: @takluyver As long as this looks good to you, then I think that this PR is ready to go!

gnestor:
https://github.com/ipython/ipython/pull/10716 depends on this PR. @Carreau @takluyver Can we cut a new ipykernel release soon?

Carreau:
What do you mean by "depend" ? Would IPython not work with an old ipykernel ? We don't need to wait on a release to merge the other one if that's not the case.

takluyver:
IPython would expose functionality that wouldn't work with the new ipykernel. It's not an actual depdency, because that goes the other way (ipykernel depends on ipython), but I'd treat this like IPython depending on the new version of ipykernel.

PR #257, opened 2017-08-02
Add flag to deactivate experimental completions

Carreau:


ipykernel/ipkernel.py
```diff
@@ -27,6 +27,8 @@ class IPythonKernel(KernelBase):
                      allow_none=True)
     shell_class = Type(ZMQInteractiveShell)
 
+    use_experimental_completions = Bool(True, help="Deactivate use of experimental IPython completion API").tag(config=True)
```
takluyver: I think the help string should either describe what the True state does, or start with "Set to False to...". It's a bit confusing for it to describe what the negative state does.

minrk: > Set to False to deactivate the use of experimental IPython completion APIs.

PR #259, opened 2017-08-08
always b64-encode bytes objects on Python 3

minrk:
There’s ambiguity in Python 2 bytes objects as to whether they are bytes or text, so we have to guess. If it’s valid text (be conservative: only ascii), treat as text, otherwise binary. On Python 3, we don't have to guess.

The guess logic on Python 2 is a bit different from before: Previously, all bytestrings were assumed to be text unless the were specific mime-types.

Upstream can always avoid ambiguity by only passing unicode. Binary data should be passed as base64-encoded unicode objects (https://github.com/ipython/ipython/pull/10739).

Handling of already-base64-encoded unicode objects is unchanged, so providing unicode objects directly as in the ipython PR is a backward- and forward-compatible choice upstream.

### DIFF
--- a/ipykernel/displayhook.py
+++ b/ipykernel/displayhook.py
@@ -68,7 +68,7 @@ class ZMQShellDisplayHook(DisplayHook):
         self.msg['content']['execution_count'] = self.prompt_count
 
     def write_format_data(self, format_dict, md_dict=None):
-        self.msg['content']['data'] = encode_images(format_dict)
+        self.msg['content']['data'] = json_clean(encode_images(format_dict))
         self.msg['content']['metadata'] = md_dict
 
     def finish_displayhook(self):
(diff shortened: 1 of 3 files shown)


takluyver:
It looks like in `ipykernel.displayhook`, encode_images is called but json_clean is not. Is that going to leave bytes in the messages if formatters produce bytes? Or am I missing a call to json_clean somewhere else?

minrk:
Let me check. I *think* json_clean gets called on every message we produce, eventually, but I'll make sure.

minrk:
Nope, you're right. We weren't calling it there. Should be okay, now.

takluyver:
Not sure what the failure on Appveyor is about, but the tests are actually passing, it just shows a non-zero exit code for some unknown reason:

```
Ran 94 tests in 31.889s
OK (SKIP=3)
Command exited with code -1073741819
```

minrk:
That's weird. I've seen that on a few other repos as well recently.

Issue #247, opened 2017-05-20
activate_matplotlib() is not called.


iktakahiro:
## Summary

Unless you intentionally omitted it, It seems to need to call ``IPython.core.pylabtools.activate_matplotlib()`` function in ``_enable_matplotlib_integration()``.

## Details

[Since version 4.4.0](https://github.com/ipython/ipykernel/blob/523cb549add6e2d5984a831582eaabd9c9af6aa3/docs/changelog.rst#440), the default backend has been changed to ``inline`` (``module://ipykernel.pylab.backend_inline``). In this issue, it's called *default inline mode*. 

Consequently, we no longer need to specify inline mode using `%matplotlib inline` with Jupyter Notebook. The mode when using the magic command is called *explicit inline mode*.

However, I've realized that the behavior of default inline mode and that of explicit inline mode are different.

1. the behavior of default inline mode

```python
from matplotlib import pyplot as plt

x = [1, 2, 3]
y = [3, 6, 9]

fig = plt.figure()
ax = fig.add_subplot(111)
ax.set_title('Sample Bar Chart')
```

![default-inline](https://cloud.githubusercontent.com/assets/1491856/26275492/c7ab4110-3d9c-11e7-9118-7da789bd14ea.png)

The figure is not rendered when a code-cell is executed. To render a graph, I have to call``plt.show()``.

2. In the case of explicitly inline mode.

```python
from matplotlib import pyplot as plt
%matplotlib inline

x = [1, 2, 3]
y = [3, 6, 9]

fig = plt.figure()
ax = fig.add_subplot(111)
ax.set_title('Sample Bar Chart')
```

![explicityly-inilne](https://cloud.githubusercontent.com/assets/1491856/26275494/cf08bc1c-3d9c-11e7-9f72-7d4c295741c8.png)

The figure is rendered even if ``plot.show()`` is not executed.

I don't know whether these behaviors are correct or not. On the other hand, I think that users are confused by it.

## Proposals

When ``%matplotlib inline`` is called, ``configure_inline_support()`` function and ``activate_matplotlib()`` are execited [^1] . But in `_enable_matplotlib_integration()`, ``activate_matplotlib()`` is not called.

Therefor, maybe simply it should be called in the Initialization process.

Could you tell me your opinion?

[^1]: https://github.com/ipython/ipython/blob/3be8fc96b227125f2dc79fc881f77b5b4e9652fb/IPython/core/interactiveshell.py#L2919

[...]

PR #264, opened 2017-09-16
activate_matplotlib() is not called in _enable_matplotlib_integration()

iktakahiro:
This PR was sent to fix the behavior of Matplotlib inline mode on Jupyter Notebook.

- before:

before

- after

after

Please refer the issue #247


minrk:
Thanks!

darthoctopus:
Can I have a way to enable the old behaviour? The new release of ipykernel is breaking all my DPI settings (causing my figures to become tiny) and also breaking plotting in the same cell as a matplotlib import statement.

PR #276, opened 2017-10-30
avoid deadlock on IOPub.flush during import

minrk:
The background thread cannot be expected to wake properly in the midst of import, so we cannot expect `evt.set` to be called until import completes. In this case, flush is still scheduled, but it is not awaited.

[Reported with PyPy](https://bitbucket.org/pypy/pypy/issues/2690/importing-module-when-pypy-is-kernel-for)

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -7,11 +7,15 @@
 from __future__ import print_function
 import atexit
 from binascii import b2a_hex
+try:
+    from importlib import lock_held as import_lock_held
+except ImportError:
+    from imp import lock_held as import_lock_held
 import os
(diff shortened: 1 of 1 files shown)


ipykernel/iostream.py
```diff
@@ -42,6 +46,10 @@ class IOPubThread(object):
     whose IO is always run in a thread.
     """
 
+    # timeout for flush to avoid infinite hang
+    # in case of misbehavior
+    FLUSH_TIMEOUT = 10
```
mattip: This should be in the OutStream class, not here

Carreau:
Thanks !

PR #278, opened 2017-11-10
update use of tornado IOLoop.current

minrk:
prepare for tornado 5:

- use thread-local .current over deprecated global .instance
- avoid making IOPubThread loops main thread's instance for a brief period during initialization

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -63,7 +63,7 @@ class IOPubThread(object):
         self.background_socket = BackgroundSocket(self)
         self._master_pid = os.getpid()
         self._pipe_flag = pipe
-        self.io_loop = IOLoop()
+        self.io_loop = IOLoop(make_current=False)
         if pipe:
             self._setup_pipe_in()
         self._local = threading.local()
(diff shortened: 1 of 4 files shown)


takluyver:
Something seems to be making the tests hang on Python 3.4 and up.

minrk:
Tests are passing for me, investigating this right now. I'd like to get this in 4.7, but no need to wait if I take too long since it's support for not-yet-released tornado.

minrk:
Seems like there wasn't anything to do here. It looks to have been a bug in pyzmq 17b2 that's fixed in 17b3, which has been released since I made the PR. Tests pass now.

blink1073:
LGTM, will merge once Appveyor agrees.  Thanks!

blink1073:
wat: "ZeroDivisionError: integer division or modulo by zero" (from the appeyor report).

minrk:
ZeroDivisionError isn't an issue, that's a test for error output (not sure why the output isn't suppressed). It's another test for a simple print that's failing with a timeout

blink1073:
Ah, okay

minrk:
restarting appveyor in case it's just weirdness

minrk:
appveyor passed where it failed before, so I think it's a timeout that might be too short for Windows. Had to rebase for conflict in travis.yml. Should be okay shortly.

blink1073:
The CI gods are being kind again, in she goes!

PR #281, opened 2017-11-23
handle KeyboardInterrupt with %gui asyncio

### LINKED ISSUE #263: Interrupting kernel from notebook closes asyncio event loop
The problem can be seen [here](https://nbviewer.jupyter.org/urls/josephweston.org/uploads/asyncio-ipybug.ipynb) running `ipykernel 4.7.0.dev`.

Interrupting the kernel from the notebook closes the `asyncio` event loop, which means that no coroutines can be scheduled on the event loop after this point.

minrk:
when interrupted, re-enter eventloop and don't close it (same as all the other eventloops)

Only close the eventloop when we are done for good AND we are responsible for running it.

Also avoid any cleanup when someone else is already running the eventloop (e.g. tornado 5).

closes #263

I'll try add a test for this before it's ready to merge.

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -302,12 +302,22 @@ def loop_asyncio(kernel):
         loop.call_later(kernel._poll_interval, kernel_handler)
 
     loop.call_soon(kernel_handler)
-    try:
-        if not loop.is_running():
+    # loop is already running (e.g. tornado 5), nothing left to do
+    if loop.is_running():
+        return
+    while True:
+        error = None
+        try:
             loop.run_forever()
-    finally:
+        except KeyboardInterrupt:
+            continue
+        except Exception as e:
+            error = e
         loop.run_until_complete(loop.shutdown_asyncgens())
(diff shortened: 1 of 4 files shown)


ipykernel/eventloops.py
```diff
@@ -302,12 +302,22 @@ def kernel_handler():
...
+    while True:
+        error = None
+        try:
             loop.run_forever()
-    finally:
+        except KeyboardInterrupt:
+            continue
+        except Exception as e:
+            error = e
         loop.run_until_complete(loop.shutdown_asyncgens())
```
takluyver: The docs say that `shutdown_asyncgens` is new in 3.6, so this code might not work with 3.5? But maybe it's only on the abstract class that the method is new. This is already in the code, so I'm happy to merge this PR anyway, but it's worth checking.

basnijholt:
@minrk thanks for this work. This fixes the problems (https://github.com/ipython/ipykernel/issues/263) I was having.

PR #283, opened 2017-11-27
process iopub events in order

minrk:
- events is <del>an OrderedDict</del> a deque
- process all events when an event arrives

avoids fair-queue issues with multiple writes from concurrent threads

This means that sometimes an event will arrive with nothing to do, but that's fine

closes https://github.com/jupyter/qtconsole/issues/246

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -7,7 +7,7 @@
 from __future__ import print_function
 import atexit
 from binascii import b2a_hex
-from collections import OrderedDict
+from collections import deque
 try:
     from importlib import lock_held as import_lock_held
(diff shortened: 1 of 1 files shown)


takluyver:
Is there a reason you picked a deque rather than a `queue.Queue`?

minrk:
queue.Queue uses locks, which we had to remove from stdout.write in #199. This would be way easier if we could use locks (we wouldn't need the zmq pipes at all). But deque is threadsafe according to the docs, at least insofar as we are using it here (read only existing items from the left in one thread, write them on the right in others).

takluyver:
Gotcha, thanks.

PR #284, opened 2017-12-04
Update help links.

ian-r-rose:
Came across this as part of some menu-related work for JupyterLab (jupyterlab/jupyterlab#3311).
1. Uses `https` where appropriate (SymPy apparently does not have `https` documentation).
2. Expands the labels to be a bit clearer.


Carreau:
Thanks !

PR #287, opened 2017-12-06
run tests with pytest

minrk:
Switching the test runner Just Works™. I did a quick find-replace pass on the simpler nose asserts, but didn't get all of them, so `nose` is still sitting as a dependency. Nose asserts work fine in pytest, so I don't feel a lot of pressure to finish this up immediately.

- builtin support for showing slow tests (--durations=N slowest test times)
- better error reporting of failures
- Only downside I think is that we lose nose_warnings_filters. I don't know a pytest equivalent, but it does have good [warnings capture](https://docs.pytest.org/en/latest/warnings.html), so adapting the plugin to run as a pytest fixture shouldn't be too much work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 python:
     - "nightly"
-    - 3.6 
+    - 3.6
     - 3.5
     - 3.4
     - 3.3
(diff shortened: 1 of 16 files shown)


blink1073:
Beauty!

PR #290, opened 2017-12-15
clean exit from eventloops

### LINKED ISSUE #274: `asyncio` and `concurrent.futures.ProcessPoolExecutor` prevent kernel restarts
The problem is illustrated in [this notebook](https://nbviewer.jupyter.org/url/joseph.weston.cloud/uploads/kernel-restart-problem.ipynb).

If I use the `asyncio` event loop integration introduced in #21 along with `concurrent.futures.ProcessPoolExecutor` then hitting the "restart" button from within the notebook interface is broken. The notebook server attempts 5 times to restart the kernel, then gives up and creates a "new" one. The previous kernel process is then orphaned, but still exists on the system (with the `init` process as the parent).

This is a problem because there is a sizeable delay between hitting "restart" and being able to use a new kernel. Also this produces many orphan processes that take resources and do not go away even once the notebook server is stopped.

### LINKED ISSUE #282: Event loop problem on Mac OS X
I am trying to run ```jupyter console``` on Mac OS X (10.13.1) after an install of Python using anaconda and I get an error that seems related to ```ipykernel```:

```python
[17:00] ~ > jupyter console
ERROR:tornado.application:Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x10b2d1400>)
  File "/Users/avigan/anaconda3/lib/python3.6/site-packages/tornado/ioloop.py", line 605, in _run_callback
    ret = callback()
  File "/Users/avigan/anaconda3/lib/python3.6/site-packages/tornado/stack_context.py", line 277, in null_wrapper
    return fn(*args, **kwargs)
  File "/Users/avigan/anaconda3/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 265, in enter_eventloop
    self.eventloop(self)
  File "/Users/avigan/anaconda3/lib/python3.6/site-packages/ipykernel/eventloops.py", line 277, in loop_cocoa
    show.mainloop()
AttributeError: 'function' object has no attribute 'mainloop'
Jupyter console 5.2.0

Python 3.6.3 |Anaconda custom (64-bit)| (default, Nov  8 2017, 18:10:31)
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]:
```

It does open a prompt at the end, but the problem is that I get weird behaviours in GUIs (e.g. in matplotlib) started from this prompt. 

Also it really seems specific to Mac OS X, because with the same install on linux I have not such issue and because of the ```loop_cocoa``` line at the end.

minrk:
When any eventloop is running, ipykernel doesn't exit cleanly because we don't stop the main eventloop, we just stop tornado, which isn't running anymore when the eventloop is active. It turns out we have been relying on jupyter-client forcibly terminating kernels that don't shutdown promptly after a polite request.

To solve this, eventloop.exit_hook is added and should be called when exit is requested (the same event where we have been stopping tornado should now stop both eventloops if they are running).

closes #274
closes #282 

TODO:

- [x] qt
- [x] asyncio
- [x] cocoa
- [x] gtk
- [x] wx
- [x] tk

cc @jbweston and @basnijholt for testing

### DIFF
--- a/ipykernel/_eventloop_macos.py
+++ b/ipykernel/_eventloop_macos.py
@@ -1,0 +1,153 @@
+"""Eventloop hook for OS X
+
+Calls NSApp / CoreFoundation APIs via ctypes.
+"""
+
+# cribbed heavily from IPython.terminal.pt_inputhooks.osx
+# obj-c boilerplate from appnope, used under BSD 2-clause
+
+import ctypes
+import ctypes.util
+from threading import Event
+
+objc = ctypes.cdll.LoadLibrary(ctypes.util.find_library('objc'))
+
+void_p = ctypes.c_void_p
+
+objc.objc_getClass.restype = void_p
+objc.sel_registerName.restype = void_p
+objc.objc_msgSend.restype = void_p
+objc.objc_msgSend.argtypes = [void_p, void_p]
+
+msg = objc.objc_msgSend
+
+
+def _utf8(s):
+    """ensure utf8 bytes"""
+    if not isinstance(s, bytes):
+        s = s.encode('utf8')
+    return s
+
+
+def n(name):
+    """create a selector name (for ObjC methods)"""
+    return objc.sel_registerName(_utf8(name))
+
+
+def C(classname):
+    """get an ObjC Class by name"""
+    return objc.objc_getClass(_utf8(classname))
+
+
+# end obj-c boilerplate from appnope
+
+# CoreFoundation C-API calls we will use:
+CoreFoundation = ctypes.cdll.LoadLibrary(ctypes.util.find_library('CoreFoundation'))
+
+CFAbsoluteTimeGetCurrent = CoreFoundation.CFAbsoluteTimeGetCurrent
+CFAbsoluteTimeGetCurrent.restype = ctypes.c_double
+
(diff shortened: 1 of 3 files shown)


ipykernel/_eventloop_macos.py
```diff
@@ -0,0 +1,153 @@
+"""Eventloop hook for OS X
+
+Calls NSApp / CoreFoundation APIs via ctypes.
+"""
+
+# cribbed heavily from IPython.terminal.pt_importhooks.osx
```
takluyver: importhooks -> inputhooks

basnijholt:
Thanks, it seems to fix the specific issue raised in https://github.com/ipython/ipykernel/issues/274.

however, restarting a kernel with this still takes a very long time:
```python
import concurrent.futures
import asyncio

executor = concurrent.futures.ProcessPoolExecutor()
ioloop = asyncio.get_event_loop()

def simple_function(x):
    import time
    time.sleep(10000)
    return x

tsk = ioloop.run_in_executor(executor, simple_function, 0)
```

minrk:
Thanks!

minrk:
@basnijholt let's investigate slow restart in a separate Issue.

PR #295, opened 2018-01-10
Revert "Add activate_matplotlib() to _enable_matplotlib_integration()"

darthoctopus:
This reverts commit 42afe03eff5be34737ed88b9b852161132628997.

Prior to this, one had to explicitly call %matplotlib inline. Avoiding this was useful for generating plots across multiple cells, which is impossible with it activated. Moreover, calling activate_matplotlib() overrides matplotlibrc for important properties like the figure DPI, which can be avoided by not calling it.


Carreau:

PR #296, opened 2018-01-16
fix %gui asyncio with tornado 5

minrk:
Fixes test failures when tornado runs on asyncio (tornado 5 + Python ≥3.5).


minrk:
The main issue appears to be related to the fact that asyncio is running with tornado 5, so the eventloop integration there is a no-op. There's something we aren't doing right in that case.

PR #301, opened 2018-01-29
Include minor version in KERNEL_SPEC

neirbowj:
This change allows concurrent installation of ipykernel under
multiple minor versions of the same major version because of files
that are installed to share/jupyter/kernels/${KERNEL_NAME}.

### DIFF
--- a/ipykernel/kernelspec.py
+++ b/ipykernel/kernelspec.py
@@ -16,7 +16,7 @@ from jupyter_client.kernelspec import KernelSpecManager
 
 pjoin = os.path.join
 
-KERNEL_NAME = 'python%i' % sys.version_info[0]
+KERNEL_NAME = 'python%i.%i' % (sys.version_info[0], sys.version_info[1])
 
 # path to kernelspec resources
 RESOURCES = pjoin(os.path.dirname(__file__), 'resources')


neirbowj:
This is to facilitate FreeBSD support for concurrent installation per available minor version. See [bug 225518](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=225518) for the downstream patch.

minrk:
I think we should be thinking about going the opposite direction, and use `python` with no version. Then the same 'python' kernelspec works correctly for the current Python, regardless of env.

Changing `python3` to `python3.6` (or `python` for that matter) as is done here, though, would be a breaking change for all existing Python notebooks in the world, so we need to be extremely careful about doing this, and only *add* pythonX.Y, instead of replacing pythonX, if we decide this is desirable.

FreeBSD packages are welcome to install pythonX.Y kernelspecs if they like. This is doable now by running `ipython kernelspec install --name pythonX.Y --prefix $PREFIX` as part of the package build.

neirbowj:
Thank you for this feedback. I will investigate whether the kernelspec installation capability is a viable alternative for the problem I'm trying to solve. What I hope is that ipykernel could gain support for a layer if indirection comparable to that defined by [PEP 394](https://www.python.org/dev/peps/pep-0394/).

The FreeBSD ports tree already supports this for the python family of ports by layering lang/python and lang/pythonX that only install suitable symlinks on top of lang/pythonXY. There is lower granularity support for concurrently installing a given python-based port that installs an executable script that renames each executable that would otherwise collide with a "-X.Y" suffix and then installs bare symlinks to the instance that uses the system default X.Y.

minrk:
You may be interested in https://github.com/jupyter/jupyter_client/pull/308 which adjusts how kernels are discovered in a way that's potentially relevant.

ivanov:
Hey there @neirbowj , I'm going through old issues and it seems to me that it makes sense to close this one. Thanks for pitching in for ipykernel both here and in the FreeBSD ports tree. 

Thanks everyone and happy hacking! :bowtie:

PR #306, opened 2018-02-08
avoid absolute paths in data_files

minrk:
usig absolute paths here causes problems for installs from sdist

We should release 4.8.1 with this, since installations are having issues on Windows.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,8 @@ if any(a.startswith(('bdist', 'build', 'install')) for a in sys.argv):
     write_kernel_spec(dest, overrides={'argv': argv})
 
     setup_args['data_files'] = [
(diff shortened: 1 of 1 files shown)


minrk:
See [this appveyor build](https://ci.appveyor.com/project/conda-forge/ipykernel-feedstock/branch/master/job/xnrl4b1r16lym041) of the ipykernel conda package, which is showing errors related to absolute paths in the sdist but somehow not failing (I think it should be failing?).

takluyver:
I think that build should be failing as well - I'd guess it's a bug with conda-forge's build scripts that it's not.

takluyver:
Are you doing the 4.8.1 release, or do you want me to?

minrk:
I'll give it a go

minrk:
On PyPI, conda [pending](https://github.com/conda-forge/ipykernel-feedstock/pull/23)

PR #307, opened 2018-02-10
ensure qt zmq streams starts in a clean state

minrk:
adding a listener on stream.FD won’t reliably wake the eventloop if there are already events waiting to be processed. We must check socket.EVENTS and process any waiting events before waiting on zmq.FD, or we will never wake.

This fixes the bug reported in https://github.com/zeromq/pyzmq/issues/1140

It's caused by updating pyzmq to version 17,
where the tornado eventloop integration also relies on zmq.FD rather than zmq_poll, which allows the initial state of the FD to be consumed (by tornado) while there are events waiting to be processed (because tornado is paused indefinitely).

I think there's something fragile in our multi-eventloop integration because we aren't explicitly advancing the tornado eventloop at all when a GUI eventloop is running, only our `kernel.do_one_iteration` callback. This means that any callbacks registered directly with tornado/asyncio will not run while eventloop integration is enabled.

cc @ccordoba12 who reported the issue. Can you confirm this fixes it for you?

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -41,6 +41,11 @@ def _notify_stream_qt(kernel, stream):
     fd = stream.getsockopt(zmq.FD)
     notifier = QtCore.QSocketNotifier(fd, QtCore.QSocketNotifier.Read, kernel.app)
     notifier.activated.connect(process_stream_events)
+    # there may already be events waiting,
+    # which won't wake zmq's edge-triggered FD
+    # make sure that waiting events are processed immediately
+    # so we start in a clean state
+    process_stream_events()
 
 # mapping of keys to loop functions
 loop_map = {


ccordoba12:
Thanks for the quick response Min! I confirm this PR solves the problem I reported in pyzmq.

ipykernel/eventloops.py
```diff
@@ -41,6 +41,11 @@ def process_stream_events():
     fd = stream.getsockopt(zmq.FD)
     notifier = QtCore.QSocketNotifier(fd, QtCore.QSocketNotifier.Read, kernel.app)
     notifier.activated.connect(process_stream_events)
+    # there may already be events waiting,
```
willingc: @minrk Double checking my understanding and perhaps tweaking the comment.

```
there may already be unprocessed events waiting.
these events will not wake zmq's edge-triggered FD
since edge-triggered notification only occurs on new i/o activity.
process all the waiting events immediately 
so we start in a clean state ensuring that any new i/o events will notify.
```

minrk: Thanks! updated

minrk:
Updated to schedule the first call on the eventloop via a QTimer, rather than making a potentially long-running blocking call immediately which may be a bit early.

willingc:
@minrk Looks good. Tested locally with @ccordoba12's setting in the config file. Tests pass and gui appears. Nicely done.

stonebig:
releasing soon ?

ccordoba12:
Yep, please. We're already getting bug reports about this:

https://stackoverflow.com/questions/48831091/ipython-console-stuck-on-connecting-to-kernel-spyder-v3-2-6-py-3-6-windo

minrk:
Just released 4.8.2 with this patch.

minrk:
conda-forge [pending](https://github.com/conda-forge/ipykernel-feedstock/pull/24)

ccordoba12:
Thanks a lot Min!

PR #314, opened 2018-03-08
kernelapp: Flush stdout/stderr before replacing

cHYzZQo:


takluyver:
Thanks, I think this makes sense.

PR #315, opened 2018-03-13
kernelapp: Allow preserving stdout and stderr in kernel

cHYzZQo:
I'd like to be able to have access to stdout/stderr the normal way so I threw together the following patch.

I imagine others might want a way to turn this off but I figured I'd share and we can refine further if others are interested.

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -276,7 +276,7 @@ class OutStream(TextIOBase):
     topic = None
     encoding = 'UTF-8'
 
-    def __init__(self, session, pub_thread, name, pipe=None):
+    def __init__(self, session, pub_thread, name, pipe=None, echo=None):
         if pipe is not None:
             warnings.warn("pipe argument to OutStream is deprecated and ignored",
                 DeprecationWarning)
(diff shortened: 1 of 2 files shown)


rgbkrk:
cc @Carreau

ipykernel/kernelapp.py
```diff
@@ -323,9 +323,13 @@ def init_io(self):
         if self.outstream_class:
             outstream_factory = import_item(str(self.outstream_class))
             sys.stdout.flush()
-            sys.stdout = outstream_factory(self.session, self.iopub_thread, u'stdout')
+            sys.stdout = outstream_factory(self.session, self.iopub_thread,
+                                           u'stdout',
+                                           echo=sys.__stdout__)
```
minrk: This shouldn't be the default behavior. The default stdout/stderr should only be redirected by default.

ipykernel/iostream.py
```diff
@@ -364,6 +380,14 @@ def _flush(self):
                 parent=self.parent_header, ident=self.topic)
 
     def write(self, string):
+        if self.echo is not None:
+            try:
+                self.echo.write(string)
+            except OSError as e:
+                if self.echo != sys.__stderr__:
```
minrk: Use `is not` for this check

ipykernel/kernelapp.py
cHYzZQo: @minrk I figured as much. How do you think this would be best configured. A property on the kernel class or a property on the kernel app?

cHYzZQo: Updated to be controlled by a ```quiet``` attribute. Similar to no_stdout.

cHYzZQo:
@rgbkrk @minrk some combination of rebasing and updated dep versions fixed the tests. This is ready for another look.

minrk:
Thanks!

PR #316, opened 2018-03-31
Override writable method on OutStream

takluyver:
Closes jupyter/notebook#3349

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,7 +18,7 @@ init:
   - cmd: set PATH=%python%;%python%\scripts;%PATH%
 install:
   - cmd: |
-        pip install --upgrade pip wheel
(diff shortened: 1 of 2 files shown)


Carreau:
Appveyor test seem to be an error on appveyor side. I can't seem to have the ability to restart the tests.

takluyver:
I do have the ability, but they still fail. It looks like pip is failing to update itself.

minrk:
We ran into the pip issue elsewhere. It's the fact that `pip.exe` is open prevents upgrade. You can upgrade pip with `python -m pip install --upgrade pip` on Windows.

takluyver:
Ah, that makes sense. #317 reinstates the pip upgrade.

PR #323, opened 2018-05-24
async support

minrk:
allow handlers to return anything acceptable to `gen.maybe_future`

this should include tornado/asyncio coroutines and Futures

counterpart to https://github.com/ipython/ipython/pull/11155

closes https://github.com/jupyter/notebook/issues/3397

### DIFF
--- a/ipykernel/tests/test_async.py
+++ b/ipykernel/tests/test_async.py
@@ -1,0 +1,75 @@
+"""Test async/await integration"""
+
+from distutils.version import LooseVersion as V
+import sys
+
+import pytest
+import IPython
+
+
+from .utils import execute, flush_channels, start_new_kernel, TIMEOUT
+from .test_message_spec import validate_message
+
+
+KC = KM = None
+
+
+def setup():
+    """start the global kernel (if it isn't running) and return its client"""
+    global KM, KC
+    KM, KC = start_new_kernel()
+    flush_channels(KC)
+
+
+def teardown():
+    KC.stop_channels()
+    KM.shutdown_kernel(now=True)
+
+
+skip_without_async = pytest.mark.skipif(
+    sys.version_info < (3, 5) or V(IPython.__version__) < V("7.0"),
+    reason="IPython >=7 with async/await required",
+)
+
+
+@skip_without_async
+def test_async_await():
+    flush_channels(KC)
+    msg_id, content = execute("import asyncio; await asyncio.sleep(0.1)", KC)
+    assert content["status"] == "ok", content
+
+
+@pytest.mark.parametrize("asynclib", ["asyncio", "trio", "curio"])
+@skip_without_async
+def test_async_interrupt(asynclib, request):
+    asynclib = "asyncio"
+    try:
+        __import__(asynclib)
+    except ImportError:
+        pytest.skip("Requires %s" % asynclib)
+    request.addfinalizer(lambda: execute("%autoawait asyncio", KC))
+
+    flush_channels(KC)
+    msg_id, content = execute("%autoawait " + asynclib, KC)
+    assert content["status"] == "ok", content
+
+    flush_channels(KC)
+    msg_id = KC.execute(
+        "print('begin'); import {0}; await {0}.sleep(5)".format(asynclib)
+    )
+    busy = KC.get_iopub_msg(timeout=TIMEOUT)
+    validate_message(busy, "status", msg_id)
+    assert busy["content"]["execution_state"] == "busy"
+    echo = KC.get_iopub_msg(timeout=TIMEOUT)
+    validate_message(echo, "execute_input")
+    stream = KC.get_iopub_msg(timeout=TIMEOUT)
+    # wait for the stream output to be sure kernel is in the async block
+    validate_message(stream, "stream")
+    assert stream["content"]["text"] == "begin\n"
+
+    KM.interrupt_kernel()
+    reply = KC.get_shell_msg()["content"]
+    assert reply["status"] == "error", reply
+    assert reply["ename"] == "CancelledError"
+
+    flush_channels(KC)
--- a/ipykernel/tests/test_async.py
+++ b/ipykernel/tests/test_async.py
@@ -42,7 +42,6 @@ def test_async_await():
 @pytest.mark.parametrize("asynclib", ["asyncio", "trio", "curio"])
 @skip_without_async
 def test_async_interrupt(asynclib, request):
-    asynclib = "asyncio"
     try:
         __import__(asynclib)
     except ImportError:
--- a/ipykernel/tests/test_async.py
+++ b/ipykernel/tests/test_async.py
@@ -70,6 +70,6 @@ def test_async_interrupt(asynclib, request):
     KM.interrupt_kernel()
     reply = KC.get_shell_msg()["content"]
     assert reply["status"] == "error", reply
-    assert reply["ename"] == "CancelledError"
+    assert reply["ename"] in {"CancelledError", "KeyboardInterrupt"}
 
     flush_channels(KC)
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ install:
       else
         SPEC="ipython==$IPYTHON"
       fi
-      pip install --pre "$SPEC"
+      pip install --upgrade --pre "$SPEC"
     fi
   - pip freeze
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ python:
     - 3.6
     - 3.5
     - 3.4
-    - 2.7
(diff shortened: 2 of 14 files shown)


minrk:
In action: asyncio integration works by default with asyncio already running:

async-ipython

Carreau:
The following restart and run all leads to a sql error when storing history
```
import asyncio
await asyncio.sleep(1)
```
```
1+1
```
```
ERROR! Session/line number was not unique in database. History logging moved to new session 14506
```

Carreau:
That also leads the the busy/idle indicator to be wrong.

Carreau:
Got also this weird things killing the kernel:
```
[(await get_build(i))['@type'] for i in range(first, first+10)]
```

```
TypeError                                 Traceback (most recent call last)
cell_name in async-def-wrapper()

TypeError: must be str, not int
```

Carreau:
I'm starting to be convinced that submitting cells should block submission of subsequent cells unless said otherwise i have the following:

```
# cells 1
from collections import Counter
c = Counter([(await get_build(i))['@type'] for i in range(first,last)])
```
```
# cell 2
c
```

Cell 1 is ran async, thus cell 2 execute immediately and raise an error `c is undefined.`

axil:
Hi, @minrk. Trying to launch some async code in jupyter notebook.

Running latest jupyter-notebook (5.5.0) with your version of ipykernel (run-async branch), getting this:
image

What am I doing wrong?

Carreau:
@minrk, do you remember the discussion we had at scipy about execution messages being consumed while code is already running ? Could you give me some pointers on how to fix this ? 

I've also pushed some compatibility with IPython 7.0.

Carreau:
I have a solution which I'm sure is wrong:

```
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -23,6 +23,7 @@ from signal import signal, default_int_handler, SIGINT
 import zmq
 from tornado import ioloop
 from tornado import gen
+from tornado.locks import Semaphore
 from zmq.eventloop.zmqstream import ZMQStream

 from traitlets.config.configurable import SingletonConfigurable
@@ -38,6 +39,13 @@ from jupyter_client.session import Session

 from ._version import kernel_protocol_version

+from collections import defaultdict
+
+# shell handlers are now coroutine (for async await code),
+# so we may not want to consume all the message and block while it yields
+_msg_locks = defaultdict(Semaphore)
+
+
 class Kernel(SingletonConfigurable):

     #---------------------------------------------------------------------------
@@ -233,7 +241,9 @@ class Kernel(SingletonConfigurable):
             self.log.debug("%s: %s", msg_type, msg)
             self.pre_handler_hook()
             try:
-                yield gen.maybe_future(handler(stream, idents, msg))
+                lock = _msg_locks[msg_type]
+                with (yield lock.acquire()):
+                    yield gen.maybe_future(handler(stream, idents, msg))
             except Exception:
                 self.log.error("Exception in message handler:", exc_info=True)
             finally:
@@ -376,7 +386,6 @@ class Kernel(SingletonConfigurable):
     @gen.coroutine
     def execute_request(self, stream, ident, parent):
```

I'm going to assume the locks will of course not always be release in the order they've been taken right ? 

See also the new PR #11265 on IPython side that fix and document a couple of other things.

minrk:
>do you remember the discussion we had at scipy about execution messages being consumed while code is already running ?

Yes, it was to switch the message handler triggers from `on_recv` callbacks to `while true` coroutines:

```python
while True:
    msg = await socket.recv_multipart()
    await self.dispatch...
```

This would require pyzmq >= 16 for asyncio support in `.recv_multipart`.

Alternately, a smaller change could be to keep the `on_recv` triggers, but *hand off* to a coroutine, so that message processing is kept in order:

```python
async def on_recv(msg):
    await self.msg_queue.put(msg)

async def process_queue():
    while True:
        msg = await self.msg_queue.get()
        await self.dispatch(msg)
```

This would have the slight *possible* downside that messages would be popped off the zmq queue into Python before they are handled. Not sure if that's a real issue, though. This pattern would give us the option to run the message handlers in an IO *thread*, since handoff could be threadsafe. This might be valuable if we want to switch runners.


> One other thing I'd like to figure out when using IPykernel is how to run user code in other eventloops like trio and curio.

I think we ought to be able to use the existing eventloop integration for qt, etc. to do this. That's where I would start.

Carreau:
I've updated this to work with the latest iteration of https://github.com/ipython/ipython/pull/11265

minrk:
@Carreau I've re-added the queue and fixed some handling of ExecutionResult objects, and now it seems to work with my testing for asyncio and trio, at least.

Carreau:
Thanks ! I'll have a look ok later. Hopping for an IPython 7.0 beta in the
next couple of weeks. Do we want to release IPykernel same time as IPython
? We could do a joint beta.

Carreau:
Thanks I had a chance to look at the code. Should the aborts also flush the newly created queue instead of just looking at ZMQ ? I suppose that's what the current failure are from.

minrk:
ok, I think the last hurdle is dealing with eventloop integration and the message queue. Now that handling messages is async, we cannot allow the eventloop integration to work the way it does, witch pausing the main eventloop and pumping `kernel.do_one_iteration()`. Instead, we need eventloop integration to work the other way - tornado is unconditionally and always running, and we pump the app eventloop while tornado is idle. That should eliminate the flushing issues.

minrk:
ok, I worked out the last two wrinkles, as far as I can tell:

- eventloops now work by returning control to the main loop, instead of calling kernel.do_one_iteration
- kernel.do_one_iteration still exists, but it's a coroutine now (it has to be)
- had to restructure `_abort_queues` to flush the message queue instead of just flushing the sockets

minrk:
This now relies on https://github.com/ipython/ipython/pull/11289

ipykernel/tests/test_async.py
```diff
@@ -0,0 +1,75 @@
...
+def test_async_await():
+    flush_channels(KC)
+    msg_id, content = execute("import asyncio; await asyncio.sleep(0.1)", KC)
+    assert content["status"] == "ok", content
+
+
+@pytest.mark.parametrize("asynclib", ["asyncio", "trio", "curio"])
+@skip_without_async
+def test_async_interrupt(asynclib, request):
+    asynclib = "asyncio"
```
Carreau: TODO: remove I guess ?

Carreau:
Test was added pre big split, currently at commit  https://github.com/ipython/ipykernel/commit/71fca97a012a1ef6401d3a90637e1d6062cf2ad1 (used to be https://github.com/ipython/ipython/pull/4363/commits/e9db6d584d3eaebca694e57038551f6a842d70eb), in PR https://github.com/ipython/ipython/pull/4363, see issue https://github.com/ipython/ipython/issues/1349 for background of why this was done, and why `set_next_input` was send originally.

Carreau:
Also I've started to draft some Blog post (https://medium.com/@mbussonn/ipython-7-0-async-repl-a35ce050f7f7) still really early, and many parts to re-write, in particular if we want to make IPython/ipykernel release at the same time.

Carreau:
The change in behavior of `?` is very much [on purpose](https://github.com/ipython/ipython/blob/a286125c5654ffe65d858c0bb6b2d72d3da27f2c/IPython/core/inputtransformer2.py#L426-L429) if you look at the code.
That is to say, the `set_next_input` is called only iif the `?` is alone on its line. 

I'm not sure if this really make sens (@takluyver ?), I'm tempted to say that the `set_next_input` makes lots of sens, especially in multiline....

Carreau:
> I'm not sure if this really make sens (@takluyver ?), I'm tempted to say that the set_next_input makes lots of sens, especially in multiline....

Meh, it's broken also on 6.4 and before in a different way, and only insert the current line, so InputTransformer2 behavior suits me. I've push a commit that fixes the test (test the actual behavior instead of testing via the fact that `?`  use the behavior)

Carreau:
Test are now passing !

minrk:
Change in set_next_input multi-call behavior seems fine to me, I'm happy with the updated test. Shall we land this and get ready to make prereleases?

Carreau:
Sounds like a plan. There is a couple more PR  in IPython I want to land.
Mostly the autoreload one, and matplotlib DISPLAY one, but I'd like for it
to be tested.

I'm just refrainig from making a release (of IPython) as apparently @ivanov
never made one, so I want to leave him the horror.

minrk:
OK! Landing this one, and we can make 5.0 beta to go with the IPython 7.0 beta

Carreau:
Side, dumb question. Should we leapfrog 5.0 and 6.0 and do a 7.0 to sync the version number of IPython and IPykernel ?

PR #336, opened 2018-08-28
Adds metadata to help display matplotlib figures legibly

telamonian:
This is the sister PR of jupyterlab/jupyterlab#5232.

This PR adds some code so that when the matplotlib inline backend is enabled, one piece of extra metadata (`'needs-background'`) is passed along with each figure. This allows frontends to intelligently display a background behind the figure if it's needed to make the tick labels on the figure legible. This is implemented as so:

Whenever the inline backend calls the `display(fig, metadata)` function, `fig` is first checked to see if it has a transparent background. If it does, the color of each tick label in the figure is checked to see if it's light or dark. If all of the ticks are light, then `'needs-background': 'light'` is added to `metadata`, or if all are dark then `'needs-background': 'dark'` is added. Otherwise, nothing is added to `metadata`.

See jupyterlab/jupyterlab#5232 for an example of an implementation of a frontend legibility fix using `'needs-background'`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ python:
     - 3.6
     - 3.5
     - 3.4
-    - 3.3
     - 2.7
 sudo: false
 install:
--- a/ipykernel/pylab/backend_inline.py
+++ b/ipykernel/pylab/backend_inline.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 
 import matplotlib
 from matplotlib.backends.backend_agg import new_figure_manager, FigureCanvasAgg # analysis: ignore
+from matplotlib import colors
 from matplotlib._pylab_helpers import Gcf
 
 from IPython.core.getipython import get_ipython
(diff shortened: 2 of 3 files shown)


blink1073:
Nice, thanks @telamonian!  @minrk, it looks like tornado requires 3.4+ now, which is causing the CI failure.  Any objections to switching this library to 3.4+ as well?

blink1073:
EOL was last year: https://www.python.org/dev/peps/pep-0398/#x-end-of-life

minrk:
@blink1073 nope, go for it. Just make sure to update python_requires in setup.py

blink1073:
Pushed the changes for 3.4+

blink1073:
Oh wait, I was thinking that we were keeping python 2.7 support, should we drop that as well?

minrk:
Let's not drop 2.7 just yet. So that would be the ever-so-convenient:

```python
python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
```

blink1073:
In she goes!

ccordoba12:
Thanks @telamonian for working on this one!! This is great!

We'll use it in Spyder too.

ian-r-rose:
Very exciting to see this land, and glad it will get some use in multiple clients!

Carreau:
That's great ! Definitively a huge plus for having organized sprint and community day.

cc @rgbkrk for nteract support, and informing @tacaswell.

PR #337, opened 2018-08-29
Release 4.9

blink1073:


Carreau:
No objections. We may get a 4.10 / 5.0 with async support in a few weeks, but I think it is reasonable to release now.

blink1073:
Released on PyPI, conda-forge should be soon.

PR #341, opened 2018-09-12
Update version str to be pep 440 compliant, start beta1

Carreau:
- should we also drop 3.4 in IPykernel 5.0 ? 
- Jump version number ot 7.0 to re-sync with IPython ?

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,8 +9,10 @@ Changes in IPython kernel
 
 `5.0.0 on GitHub <https://github.com/ipython/ipykernel/milestones/5.0>`__
 
+- Drop support for Python 2. ``ipykernel`` 5.0 requires Python >= 3.4
 - Add support for IPython's asynchronous code execution (:ghpull:`323`)
 - Update release process in ``CONTRIBUTING.md`` (:ghpull:`339`)
+
 
 4.9
 ---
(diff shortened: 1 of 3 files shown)


ipykernel/_version.py
```diff
@@ -1,5 +1,16 @@
-version_info = (5, 0, 0, 'dev')
-__version__ = '.'.join(map(str, version_info))
+version_info = (5, 0, 0, 'b1')
+__version__ = '.'.join(map(str, version_info[:3]))
+
+# pep440 is annoying, beta/alpha/rc should _not_ have dots or pip/setuptools
```
minrk: This isn't an issue because we are using setuptools in setup.py. It only comes up where we skip the setuptools import in setup.py when making an sdist, which we don't do here.

docs/changelog.rst
```diff
@@ -1,6 +1,17 @@
 Changes in IPython kernel
 =========================
 
+5.0
+---
+
+5.0.0
+*****
+
+`4.9.0 on GitHub <https://github.com/ipython/ipykernel/milestones/5.0>`__
```
minrk: 5.0.0

minrk:
> Jump version number ot 7.0 to re-sync with IPython ?

I don't think so—the whole point of separate packages is that the versions are independent, so they *should* diverge. If the versions should be kept in sync, then they shouldn't be separate packages.

minrk:
5.0b1 is tagged and up on PyPI

Carreau:
THanks ! Apologies if I'm a bit slow these days.

PR #344, opened 2018-09-20
Check that sys.stdout/err is not None before flushing

jenshnielsen:
On Windows it seems that sys.stdout and possibly sys.stderr can be None. 
This happens at startup of the kernel in Spyder e.g 
https://github.com/spyder-ide/spyder/issues/7842 and multiple duplicates.

The change to flush was introduced in 4.9 with https://github.com/ipython/ipykernel/pull/314

Checking that setout/stderr is not None before flushing seems to be enough to fix the issue for me.

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -323,7 +323,8 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
         """Redirect input streams and set a display hook."""
         if self.outstream_class:
             outstream_factory = import_item(str(self.outstream_class))
-            sys.stdout.flush()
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks a lot @jenshnielsen! It looks good to me.

Carreau:
Marking as 4.10, the current master branch have diverged from 4.x, so this will need a backport.

Carreau:
@meeseeksdev backport to 4.x

Carreau:
branches were messed up, re-trying.

Carreau:
@meeseeksdev backport to 4.x

PR #360, opened 2018-10-05
Handle signal exceptions, fix for issue 11258

n8vm:
Prior to this change, signal throws an unhandled exception which terminates the ipykernel when ran on a separate thread. This is a problem for embedded applications which need the main thread dedicated to window event polling (eg for GLFW) and would like an ipython kernel running in a separate thread.

With these changes, the unhandled exception is now caught, and displayed as an error, and the kernel keeps running. 

https://github.com/ipython/ipython/issues/11258


minrk:
Thanks!

Issue #51, opened 2015-09-05
do_inspect lets ipython compute the info twice


jankatins:
Found this while trying to understand what is needed to make a decorated method get the same signature/completions as the original one, even on py2.7.

From https://github.com/ipython/ipykernel/blob/master/ipykernel/ipkernel.py#L247:

```
    def do_inspect(self, code, cursor_pos, detail_level=0):
        name = token_at_cursor(code, cursor_pos)
        info = self.shell.object_inspect(name) # computes the complete info
[...]
        if info['found']:
            info_text = self.shell.object_inspect_text(
                name,
                detail_level=detail_level,
            ) # computes it again as a first step
            data['text/plain'] = info_text
[...]
```

IMO it would be ok to replace the first `self.shell.object_inspect(...)` with a `info = self.shell._object_find(name)` and then use `if info.found:`. The problem is I don't understand what the `with self.builtin_trap:` in the `object_inspect` method does ([source](https://github.com/ipython/ipython/blob/master/IPython/core/interactiveshell.py#L1608)) and if that's already needed for `self.shell._object_find`.

minrk:
That makes good sense to me. I just want to avoid calling private APIs in IPython from ipykernel. I think we need to take a pass on oinspect in IPython to make a few more ways to call its pieces.

JanSchulz:
oinspect probably also needs a pass as `inspect.getfullargspec` is deprecated ( https://docs.python.org/3/library/inspect.html#inspect.getfullargspec and https://github.com/matplotlib/matplotlib/pull/5014#issuecomment-137253854) and using the new `Signature` interface will help with things like getting the proper signature from decorated functions (see https://github.com/matplotlib/matplotlib/pull/4829#issuecomment-137912544 and following comments)

PR #361, opened 2018-10-07
Allow `do_inspect` to return a rich display when `enable_html_pager` is True

tonyfast:
The JupyterLab inspector can will accept __text/html__ results.  This pull request uses `object_inspect_mime` instead of `object_inspect_text` to inspect objects.  `object_inspect_text` is already calling `object_inspect_mime` to return __plain/text__.  This new feature optionally removes the __text/html__ result data if `enable_html_pager` is False.

This also addresses an old issue #51 and inspects the object once.


minrk:
Nice!

Issue #370, opened 2018-11-06
Kernel crashes after any unhandled exception


giantamoeba:
I have been experiencing this strange problem after updating to version 5.0 (5.1.0 still has the problem). All runs fine on version 4.9.

Minimal Example:
```python

import sys
from PyQt5.QtWidgets import QApplication

from qtconsole.rich_ipython_widget import RichJupyterWidget
from qtconsole.inprocess import QtInProcessKernelManager

app = QApplication(sys.argv)

kernel_manager = QtInProcessKernelManager()
kernel_manager.start_kernel()
kernel = kernel_manager.kernel
kernel.gui = 'qt'

kernel_client = kernel_manager.client()
kernel_client.start_channels()

widget = RichJupyterWidget()

widget.kernel_manager = kernel_manager
widget.kernel_client = kernel_client

widget.show()

sys.exit(app.exec_())
```
Then in the console, raise any exception:
```
In [1]: raise RuntimeError('ohai')
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-1-dba2c2e9b9ef> in <module>
----> 1 raise RuntimeError('ohai')

RuntimeError: ohai

In [2]: print()
ERROR: execution aborted
```

Has anyone else experienced this problem?

gdementen:
yes, I have the same problem. Only workaround I found is downgrade back to 4.10

giantamoeba:
I hope this is not the end of in-process kernels, I need in my applications a shared namespace so that the console can manipulate the application, and a kernel launched as a subprocess does not provide such a functionality..

rayosborn:
It is vital to the [NeXpy application](http://nexpy.github.io/nexpy/) that this is resolved. I am trying to understand the new asynchronous message queues that were introduced in v5.0.0. It looks as if exceptions are not flushed from these queues before entering a new line in the terminal. I know that the main ipykernel developers do not consider the in-process kernel a high priority, although it is vital to a number of other applications, but if someone can explain where the message queues are reset in multi-kernel applications, that might help us to prepare a Pull Request to fix this issue.

rayosborn:
I have submitted a Pull Request #376, which I think fixes this issue. It turned out to be as simple as adding a dummy `_abort_queues` function to the `InProcessKernel` that overrides the default function in the parent `Kernel` class. If you are able to test this fix, please let me know if still has any issues.

rayosborn:
[...]

PR #376, opened 2018-12-04
Fix compatibility with InProcessKernel

rayosborn:
From v5.0.0, the `Kernel` class contains an `_abort_queues` function invoked when the shell executes a request that generates an exception. This schedules a message, which clears the abort when dispatched. 

This does not work in the `InProcessKernel`. After an exception, all shell execute requests return an "ERROR: execution aborted" message. This is fixed by overriding the `_abort_queues` function in the parent `Kernel` class by a dummy function in the `InProcessKernel` sub-class (it actually replaces the `_abort_queue` function, which doesn't appear to be used any more).

This is a fix for #370.

### DIFF
--- a/ipykernel/inprocess/ipkernel.py
+++ b/ipykernel/inprocess/ipkernel.py
@@ -83,7 +83,7 @@ class InProcessKernel(IPythonKernel):
         """ Override registration of dispatchers for streams. """
         self.shell.exit_now = False
 
-    def _abort_queue(self, stream):
+    def _abort_queues(self):
         """ The in-process kernel doesn't abort requests. """
         pass
 


minrk:
Thanks!

astrofrog:
Would it be possible to release 5.1.1 shortly? (I'm getting more and more user reports of this issue for glue (glueviz.org) which uses the inprocesskernel)

yves-surrel:
The problem comes from the line
```
yield gen.sleep(self.stop_on_error_timeout)
```
that hangs forever in `_dispatch_abort()` of `kernelbase.py`, although it should not, according to the `tornado.gen.sleep` doc, that states that 

> When used with yield in a coroutine, this is a non-blocking analogue to time.sleep .

Commenting out the line solves the problem (setting `self.stop_on_error_timeout` to 0 does not)

PR #377, opened 2018-12-12
make tmp install dir user writable after copytree

ecoughlan:
When installing if the RESOURCES path is not user writable, installation fails. This happens for instance on NixOS, but it's probably a bad idea to depend on the source modes in any case.

### DIFF
--- a/ipykernel/kernelspec.py
+++ b/ipykernel/kernelspec.py
@@ -10,6 +10,7 @@ import json
 import os
 import shutil
 import sys
+import stat
 import tempfile
 
 from jupyter_client.kernelspec import KernelSpecManager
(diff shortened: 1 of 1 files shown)


ivanov:
Hey there @Mic92 and @ecoughlan, a very belated thank you for getting involved and making this contribution and using the patch in the nix packaging. I'm going through old issues and it seems to me that it makes sense to close this one. It looks like an equivalent fix was submitted and merged in #593, which made its way into ipykernel 5.5.0. Looks like @FRidh already took out this patch in https://github.com/NixOS/nixpkgs/commit/e3900e3e0b0fdbf1e7fd6e92bc845545cb30c01d when upgrading to ipykernel 5.5.0 in the nixpkgs.

Thanks everyone and happy hacking! :bowtie:

ivanov:
and again, my sincerest apologies that this went overlooked for so long and a newer fix ended up getting merged instead of yours. Hopefully we'll catch you further down the commit log.

PR #379, opened 2018-12-14
Increase log level for exception in execute request

fgregg:
This will make it easier for downstream tools like http://mpastell.com/pweave/ to show useful error messages


Carreau:

PR #384, opened 2019-01-30
Make ipykernel use sys.executable

SylvainCorlay:
cf https://github.com/conda-forge/ipykernel-feedstock/issues/44

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -114,7 +114,12 @@ setup_args = dict(
 if any(a.startswith(('bdist', 'build', 'install')) for a in sys.argv):
     from ipykernel.kernelspec import write_kernel_spec, make_ipkernel_cmd, KERNEL_NAME
 
-    argv = make_ipkernel_cmd(executable='python')
+    # When building a wheel, the executable specified in the kernelspec is simply 'python'.
+    # When installing from source, the full `sys.executable` can be used.
+    if any(a.startswith('bdist') for a in sys.argv):
+        argv = make_ipkernel_cmd(executable='python')
+    else:
+        argv = make_ipkernel_cmd()
     dest = os.path.join(here, 'data_kernelspec')
     if os.path.exists(dest):
         shutil.rmtree(dest)


setup.py
```diff
@@ -114,7 +114,18 @@ def run(self):
...
     from ipykernel.kernelspec import write_kernel_spec, make_ipkernel_cmd, KERNEL_NAME
 
-    argv = make_ipkernel_cmd(executable='python')
+    # If the IPYKERNEL_USE_SYS_EXECUTABLE environment variable is set to 1
+    # then `make_ipkernel_cmd` will generate a kernelspec including the full
+    # sys.executable path.
+    #
+    # Otherwise, a generic 'python' is used instead.
+
+    if os.getenv('IPYKERNEL_USE_SYS_EXECUTABLE'):
```
ccordoba12: Why an env var and not a proper option? This will be very hard to find out for other projects unaware of this change.

Also, why not making this the default?

SylvainCorlay: Regarding the default, the only issue is that (AFAIK), python wheels don't do any relocation.

So having the kernel spec in the wheel for ipykernel merely "python" seems to be the way to go.

minrk: Rather than a boolean, maybe `IPYTHON_KERNELSPEC_PYTHON` and use the value in the env? e.g. `IPYTHON_KERNELSPEC_PYTHON=$PYTHON python setup.py install`

This seems more complicated and harder to find than the `python -m ipykernel install` command that always works.

If we do this, it would be worth adding a comment here about why we have to use `python` as the default in order to make wheels portable, and that the flag is really only for use with conda-build.
Ind

SylvainCorlay: I agree that it is harder to find, but the old way should still work in the same fashion.

I merely want to enable that option for the conda recipe as per the issue mentioned above so that the conda package does not require `jupyter-kernel.exe` on windows.

minrk: > Yeah, although the issue is that it does not work for some reason with the latest conda.

That's a much more significant issue, though, and shouldn't be relevant to this package: an executable from a different package (jupyter-client) is not being installed in a form that works at all. Changing this behavior to get past a fundamentally broken dependency package is probably not the right fix, in my view.

minrk:
> It is weird that wheels don't include any utility for relocation.

They do, but *only* for scripts. I opened an Issue about generalizing it on pip a few years ago, specifically for ipykernel's use case: https://github.com/pypa/pip/issues/4032

SylvainCorlay:
@minrk you may like the new version better

- running `python setup.py install` now use the full `sys.executable` in the kernelspec.
- running `python setup.py build` or `python setup.py bdist_*` will simply use `'python'`.

jdemeyer:
See also the somewhat related issue https://github.com/jupyter/echo_kernel/issues/3

SylvainCorlay:
Hey @minrk do you have any comment with the latest solution I posted?

SylvainCorlay:
> Makes sense, thanks!

Great. I will fix the test and get this in.

ceteri:
@SylvainCorlay @minrk 
> running python setup.py install now use the full sys.executable in the kernelspec.

Is there any documentation which shows how to build a wrapper kernel and package it for PyPi, Conda, etc.?

I find that the https://jupyter-client.readthedocs.io/en/stable/wrapperkernels.html instructions assume a trivial layout which wouldn't be compatible with PyPi packaging.

The examples such as https://github.com/jupyter/echo_kernel rely on a relatively opaque and complex `install.py` which I cannot find documented anywhere, and this approach appears to be out of date?

Many thanks in advance.

blink1073:
Hi @ceteri, the echo kernel is using [hatch](https://github.com/jupyter/echo_kernel/blob/b483ceb1b50425bf37157ddc1e0f338e21917647/pyproject.toml#L50) to install the kernel, which works on PyPI and conda-forge. `install.py`  is still there but could be removed, simplified, and absorbed into https://github.com/jupyter/echo_kernel/blob/main/hatch_build.py.

ceteri:
thank you for the help @blink1073 

fwiw, i worked back from the https://github.com/takluyver/bash_kernel/tree/master/bash_kernel example to reverse-engineer a generic kernel template, with the `hatch` dependency.

PR #396, opened 2019-04-09
Add Python 3.8 to Travis CI

hroncok:
(Let's see what burns.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
 language: python
 python:
   - "nightly"
-  - "3.7-dev"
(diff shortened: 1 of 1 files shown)


hroncok:
This is consistent with nightly ATM, so I'm not sure if wroth it. But it seems to work.

willingc:
Sorry for the huge delay in merging this. I just took a look at the failing test version and reran it. It was a spurious failure before. Happy that this will start the testing on 3.8 dev version. Thanks @hroncok :cake:

PR #399, opened 2019-04-11
Allow pyside2 gui loop

### LINKED ISSUE #398: os.environ['QT_API'] gets overwritten in event-loop. No pyside2 integration.
Consider this IPython interactive session:

```python
Jupyter QtConsole 4.5.0.dev0
Python 3.6.8 |Anaconda, Inc.| (default, Feb 21 2019, 18:30:04) [MSC v.1916 64 bit (AMD64)]
IPython 7.3.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: IP = get_ipython()

In [2]: import os

In [3]: os.environ['QT_API']
Out[3]: 'pyside2'

In [4]: IP.enable_gui('qt')

In [5]: os.environ['QT_API']
Out[5]: 'pyqt5'

In [6]: os.environ['QT_API'] = 'pyside2'

In [7]: os.environ['QT_API']
Out[7]: 'pyqt5'

In [8]: from IPython.lib.guisupport import get_app_qt4

In [9]: app = get_app_qt4()

In [10]: app
Out[10]: <PySide2.QtWidgets.QApplication at 0x7ff9ddd4c568>
```

What happened between 6 and 7?

The answer is that os.environ['QT_API'] gets overwritten in the event-loop all the time once it has been activated. And even though I initially selected pyside2 it changes the env-var to pyqt5. But apparently the desired backend (pyside2) is still used. I do however run into problems, e.g. when importing directly from PyQt5 versus PySide2 depending on the environment variable.

I will attempt a fix.

bergtholdt:
Fixes #398

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -125,7 +125,8 @@ def loop_qt4(kernel):
 @register_integration('qt', 'qt5')
 def loop_qt5(kernel):
     """Start a kernel with PyQt5 event loop integration."""
-    os.environ['QT_API'] = 'pyqt5'
+    if os.environ['QT_API'] not in ['pyqt5', 'pyside2']:
+        os.environ['QT_API'] = 'pyqt5'
     return loop_qt4(kernel)
 
 


bergtholdt:
Sad that this gets ignored completely.

blink1073:
@ccordoba12 does this change look reasonable to you?

blink1073:
@bergtholdt please understand that we are volunteers and not all of us are equally familiar with the code base.

ipykernel/eventloops.py
```diff
@@ -125,7 +125,8 @@ def loop_qt4(kernel):
 @register_integration('qt', 'qt5')
 def loop_qt5(kernel):
     """Start a kernel with PyQt5 event loop integration."""
-    os.environ['QT_API'] = 'pyqt5'
+    if os.environ['QT_API'] not in ['pyqt5', 'pyside2']:
+        os.environ['QT_API'] = 'pyqt5'
```
ccordoba12: I really don't understand this change. I mean, this

    if os.environ['QT_API']

assumes that `QT_API` is already in the environment, which is not always the case.

And this

    not in ['pyqt5', 'pyside2']

seems to assume that the only possible values of `QT_API` can be `pyqt5` and `pyside2`, when they can also be `pyqt` and `pyside`.

bergtholdt:
I removed the offending line

os.environ['QT_API'] = 'pyqt5'

In another pull request #444 

completely and did not find any issues in qtconsole with the qt5 gui integration and matplotlib.

PR #405, opened 2019-05-26
only flush stream in dispatch_shell if possible

Chilipp:
When using an inprocess kernel, the kernel client calls the `dispatch_shell` method with an `ipykernel.inprocess.socker.DummySocket`

https://github.com/ipython/ipykernel/blob/b81632e0a26671e6b100471282cf49ca4a21e681/ipykernel/inprocess/client.py#L170

This method, however, expects the given stream to have a `flush` method 

https://github.com/ipython/ipykernel/blob/b81632e0a26671e6b100471282cf49ca4a21e681/ipykernel/kernelbase.py#L286

which is not True, for the `DummySocket`. This PR proposes to accept failures for the latter by replacing this line with

```python
try:
    stream.flush(zmq.POLLOUT)
except AttributeError:
    pass
```

### DIFF
--- a/ipykernel/inprocess/client.py
+++ b/ipykernel/inprocess/client.py
@@ -172,6 +172,9 @@ class InProcessKernelClient(KernelClient):
         idents, reply_msg = self.session.recv(stream, copy=False)
         self.shell_channel.call_handlers_later(reply_msg)
 
+    def flush(self):
+        # dummy method (see https://github.com/ipython/ipykernel/pull/405)
+
 
 #-----------------------------------------------------------------------------
 # ABC Registration
--- a/ipykernel/inprocess/client.py
+++ b/ipykernel/inprocess/client.py
@@ -173,7 +173,7 @@ class InProcessKernelClient(KernelClient):
         self.shell_channel.call_handlers_later(reply_msg)
 
     def flush(self):
-        # dummy method (see https://github.com/ipython/ipykernel/pull/405)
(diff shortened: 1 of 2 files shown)


minrk:
Thanks for the PR! I think it's a valid assumption that the stream provide the stream API. Maybe the better fix is to add a no-op flush method to DummySocket so that it complies with the stream API?

Chilipp:
You're right @minrk, I fully agree. It revert the previous implementation (https://github.com/ipython/ipykernel/pull/405/commits/14a2c7a6ad5c6949bf1449fce322a2faa84c9863) and added a dummy flush method (https://github.com/ipython/ipykernel/pull/405/commits/c9ec71313a5058fe1dcc3a4cc5a78ca67a4fe407)

ipykernel/inprocess/client.py
```diff
@@ -172,6 +172,9 @@ def _dispatch_to_kernel(self, msg):
         idents, reply_msg = self.session.recv(stream, copy=False)
         self.shell_channel.call_handlers_later(reply_msg)
 
+    def flush(self):
+        # dummy method (see https://github.com/ipython/ipykernel/pull/405)
```
minrk: Needs a docstring or `pass` to be valid syntax here.

```suggestion
        """no-op to comply with stream API"""
```

minrk:
Thanks!

Issue #406, opened 2019-05-27
Test failures with 3.8.0a4+


hroncok:
In Fedora, we get the following test failures with Python 3.8.0a4:

```
+ /usr/bin/python3 -m pytest ipykernel -k 'not test_sys_path and not test_sys_path_profile_dir and not test_complete'
============================= test session starts ==============================
platform linux -- Python 3.8.0a4, pytest-4.4.1, py-1.8.0, pluggy-0.11.0
rootdir: /builddir/build/BUILD/ipykernel-5.1.1
collected 100 items / 5 deselected / 95 selected

ipykernel/inprocess/tests/test_kernel.py .s..                            [  4%]
ipykernel/inprocess/tests/test_kernelmanager.py ....                     [  8%]
ipykernel/tests/test_async.py EEEE                                       [ 12%]
ipykernel/tests/test_connect.py ..                                       [ 14%]
ipykernel/tests/test_embed_kernel.py ...                                 [ 17%]
ipykernel/tests/test_eventloop.py E                                      [ 18%]
ipykernel/tests/test_io.py .                                             [ 20%]
ipykernel/tests/test_jsonutil.py .....                                   [ 25%]
ipykernel/tests/test_kernel.py FFFFFsFF.FsFF                             [ 38%]
ipykernel/tests/test_kernelspec.py ..........                            [ 49%]
ipykernel/tests/test_message_spec.py EEEEEEEEEEEEEEEEEEEEE               [ 71%]
ipykernel/tests/test_pickleutil.py FFFF.                                 [ 76%]
ipykernel/tests/test_serialize.py ..........F.F.                         [ 91%]
ipykernel/tests/test_start_kernel.py ..                                  [ 93%]
ipykernel/tests/test_zmq_shell.py ......                                 [100%]
```

Tracebacks are too long, attached in full log.

[builder-live.log](https://github.com/ipython/ipykernel/files/3224009/builder-live.log)

I'm working on a reproducer.

hroncok:
```
$ git clone https://github.com/cython/cython
$ git clone https://github.com/numpy/numpy
$ git clone https://github.com/zeromq/pyzmq.git
$ git clone https://github.com/jupyter/jupyter_client.git

$ python3.8 -m venv __venv__
$ . __venv__/bin/activate
$ (__venv__) $ python --version
Python 3.8.0a4

(__venv__) $ cd cython/
(__venv__) [cython (master)]$ git checkout 0.29.x
[...]

PR #408, opened 2019-05-28
Python 3.8: PEP 570 -- Python Positional-Only Parameters

hroncok:
https://www.python.org/dev/peps/pep-0570/

Fixes https://github.com/ipython/ipykernel/issues/406


minrk:
Thanks!

PR #411, opened 2019-06-02
Use shell.input_transformer_manager when available

### LINKED ISSUE #375: IPykernel triggers repeated DeprecationWarnings due to input_splitter
@minrk When certain commands are run (e.g. ``print(<object>)``) in Spyder's Console, ``ipykernel`` (including the latest version, 5.1.0) raises one or something many ``DeprecationWarning``s like the following every time enter is pressed, and often doesn't run the command until forced with ``Shift-Enter``:

```python-traceback
C:\Anaconda3\envs\pyart\lib\site-packages\ipykernel\ipkernel.py:448: DeprecationWarning: `input_splitter` is deprecated since IPython 7.0, prefer `input_transformer_manager`.
  status, indent_spaces = self.shell.input_splitter.check_complete(code)
```

This is not only annoying, but actually seriously affects the functionality by not allowing the user to run the command normally. It was first reported in spyder-ide/spyder#8103 and "fixed" in spyder-ude/spyder-kernels#59 , but it doesn't seem like it was actually reported here where it belongs for some reason, and instead that "solution" was just to suppress ``DeprecationWarnings`` from ``ipykernel`` entirely.

Therefore, I wanted to give you a heads up before it breaks! [Here](https://github.com/ipython/ipykernel/blob/35b1d1a823dd22a0fe4eb99e89463c2527a11271/ipykernel/ipkernel.py#L448) 's the relevant line that's triggering it, for your convenience; presumably fixing it would involve updating it to use ``input_transformer_manager`` instead but I'm no expert. Thanks!

dandavison:
Fixes #375 

With this change we don't trigger the [deprecation warning](https://github.com/ipython/ipython/blob/fb9c8ca5ccd24a897a79a217d446c8ce7c4a715a/IPython/core/interactiveshell.py#L522) on `ipython.core.InteractiveShell` if we can avoid it.

## Testing
I started a kernel (using `ipykernel.kernelapp.launch_new_instance`) on this commit, and on the parent b81632e, and connected to it with `jupyter console`, and confirmed that the deprecation message appeared at the prompt when executing code on  b81632e, but was not present on this commit.

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -445,7 +445,11 @@ class IPythonKernel(KernelBase):
         return dict(status='ok', restart=restart)
 
     def do_is_complete(self, code):
-        status, indent_spaces = self.shell.input_splitter.check_complete(code)
+        transformer_manager = getattr(self.shell, 'input_transformer_manager', None)
+        if transformer_manager is None:
+            # input_splitter attribute is deprecated
+            transformer_manager = self.shell.input_splitter
+        status, indent_spaces = transformer_manager.check_complete(code)
         r = {'status': status}
         if status == 'incomplete':
             r['indent'] = ' ' * indent_spaces


blink1073:
I'm not sure why the 3.4 build is failing.  I'd like to remove 3.4 support as part of a minor release. Skipping this for the patch release.  Thanks for the fix and your patience @dandavison!

PR #414, opened 2019-06-21
PR: Raw_input function working with special characters in python 2.7

### LINKED ISSUE #404: Problem with accent
Good morning,
I use Spyder with Python 2.7 and I get a problem with raw_input and accent : 
In Spyder console if I try, for example : 

x=raw_input("début : ")
ZMOpYnV0OiA=

So I get a strange string ????

Without accent, that works : 
x=raw_input("debut: ")
debut: 
but I am french and I must use accent !

I says that issue to spyder github but Carlos Cordoba says me : "Anyway, please report your problem in ipykernel, because that's the right place to fix it, not here."
Note that there is no problem in a terminal nor with python 3. But I must use Python 2. Can you help me please.
Thank you and excuse my very bad english.
Pascal

steff456:
This change Fixes #404

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -698,7 +698,7 @@ class Kernel(SingletonConfigurable):
             raise StdinNotImplementedError(
                 "raw_input was called, but this frontend does not support input requests."
             )
-        return self._input_request(str(prompt),
+        return self._input_request(prompt.decode('utf-8'),
             self._parent_ident,
             self._parent_header,
             password=False,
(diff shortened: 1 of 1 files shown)


ipykernel/kernelbase.py
```diff
@@ -698,7 +698,7 @@ def raw_input(self, prompt=''):
             raise StdinNotImplementedError(
                 "raw_input was called, but this frontend does not support input requests."
             )
-        return self._input_request(str(prompt),
```
ccordoba12: @steff456, please use here the function called `cast_unicode_py2` from the module `ipython_genutils.py3compat`. That should solve the problems with the tests.

ccordoba12:
@minrk, tests failures on Python 3 seem unrelated to this PR. Could you take a look at them?

minrk:
Yeah, test failures are unrelated. Thanks for the fix!

ccordoba12:
Thanks a lot Min! When do you think a new 4.x release would be available?

PR #419, opened 2019-08-07
Further cleanup on shutdown

minrk:
- restore sys.stdout, etc. prior to closing forwarded io streams.
  This prevents errors if print, excepthook, etc. are called during shutdown.
- cleanup heartbeat thread in the right order:
  - terminate context from main thread (context.term is the only threadsafe zmq method)
  - catch resulting ETERM in thread and close socket

cf #412 and jupyter/nbconvert#1052


MSeal:
Thanks you @minrk !

willingc:
Way to go @minrk. Thanks! And @blink1073 for the review too.

Issue #420, opened 2019-08-07
5.1.2 Release


MSeal:
The changes list: https://github.com/ipython/ipykernel/compare/v5.1.1...master is quite small. 
#396 
#408
#405
#394
and #412 (with #419 fixing an issue with that PR)
There's no interface changes, just internal improvements which should justify a patch release imo.

@minrk or @Carreau perhaps one of you could ferry this to a release? We're wrapping up the 5.6 release process for nbconvert this week and it'd be great if this was out first.

willingc:
I'll draft a quick PR to update the Changelog for 5.1.2.

blink1073:
Thanks @willingc!  I'll cut a release once that one finishes CI.

blink1073:
https://github.com/ipython/ipykernel/releases/tag/v5.1.2

blink1073:
Shucks, my release script didn't take into account the wheel configuration, I need to cut a new release.

blink1073:
Ah, nevermind, there is only one wheel now, we dropped py27 compat.  All good!

MSeal:
Awesome thanks everyone for helping get that released!

PR #421, opened 2019-08-07
Add changelog for 5.1.2

willingc:
Addresses #420

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,20 @@ Changes in IPython kernel
 5.1
 ---
 
+5.1.2
+*****
+
+5.1.2 fixes some socket-binding race conditions that caused testing failures in
+nbconvert.
+
+- Fix socket-binding race conditions (:ghpull: `412`, :ghpull: `419`)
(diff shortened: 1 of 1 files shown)


blink1073:
https://github.com/ipython/ipykernel/releases/tag/v5.1.2

Carreau:
Thanks @blink1073 !

As a side note, can we try to keep the consistency in repositories, with the naming of tags, and wether or not there  are keeping a leading `v` or not ? 

We could also add 2 tag with and w/o `v`, but think it would be better if same repo keep same format.

blink1073:
Agreed, the leading "v" is an artifact of my bash function for making python releases. Happy to standardize one way or the other.

Carreau:
> Agreed, the leading "v" is an artifact of my bash function for making python releases. Happy to standardize one way or the other.

Also now PyPI support upload tokens (still beta-ish), so we can try to go release via travis /azure/whatever.

Probably just a question of figuring out if it's a tagged version I guess.

blink1073:
A leading "v" would help with that. ;)

cf https://github.com/jupyterlab/jupyterlab/issues/6819
cf https://github.com/jupyterlab/jupyterlab/pull/6534

PR #426, opened 2019-08-31
Rename _asyncio.py to _asyncio_utils.py to avoid name conflicts on Python 3.6+

edisongustavo:
The name _asyncio is reserved to Python, which can lead to naming
conflicts when running the tests depending on the setup of the machine.

On PyCharm this is specially painful where tests would fail for no good
reason (specially the ones that try to instantiate a remote kernel).

Some tests though would give the following error message:

```
E             File "/home/edisongustavo/miniconda3/envs/ipykernel/lib/python3.7/asyncio/futures.py", line 387, in <module>
E               Future = _CFuture = _asyncio.Future
E           AttributeError: module '_asyncio' has no attribute 'Future'
```

So I've setup a debugger and a breakpoint at that specific line and I
found that it import from this `_asyncio.py` file under the `tests`
directory.


blink1073:
Thanks @edisongustavo and congratulations on your first PR here!

PR #433, opened 2019-09-14
PR: Skip unregistering comm with destructor.

impact27:
The comm can not still be registered when `__del__` is being called:
 - if `__del__` is called, the reference count is at 0.
 - if the reference count is at 0, then the comm manager doesn't have a reference to the comm. 

This avoids https://github.com/spyder-ide/spyder-kernels/issues/145

### DIFF
--- a/ipykernel/comm/comm.py
+++ b/ipykernel/comm/comm.py
@@ -73,7 +73,7 @@ class Comm(LoggingConfigurable):
 
     def __del__(self):
         """trigger close on gc"""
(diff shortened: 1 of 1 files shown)


ccordoba12:
@blink1073, could you take a look at this one? We need it for the next major release of Spyder, which we're trying to get out of the door in a month or so.

ccordoba12:
Thanks a lot @blink1073!

PR #434, opened 2019-09-14
Add release notes for 5.1.3

blink1073:
cc @SylvainCorlay @willingc @ccordoba12

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -11,6 +11,7 @@ Changes in IPython kernel
 
 - Fix comm shutdown behavior by adding a ``deleting`` option to ``close`` which can be set to prevent registering new comm channels during shutdown (:ghpull: `433`, :ghpull: `435`)
 - Fix ``Heartbeat._bind_socket`` to return on the first bind (:ghpull: `431`)
+- Moved ``InProcessKernelClient.flush`` to ``DummySocket`` (:gphull: `437`)
 - Don't redirect stdout if nose machinery is not present (:ghpull: `427`)
 - Rename `_asyncio.py` to `_asyncio_utils.py` to avoid name conflicts on Python 3.6+  (:ghpull: `426`)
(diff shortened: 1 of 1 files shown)


ccordoba12:
@blink1073, could you also take a look at #435 before releasing 5.1.3?

docs/changelog.rst
```diff
@@ -4,6 +4,18 @@ Changes in IPython kernel
 5.1
 ---
 
+5.1.3
+*****
+
+5.1.3 Includes several bugfixes and internal logic improvements.
+
+- Fix comm shutdown behavior (:ghpull: `433`, :ghpull: `435`)
```
willingc:
```suggestion
- Fix comm shutdown behavior by adding a ``deleting`` option to ``close`` which can be set to prevent registering new comm channels during shutdown (:ghpull: `433`, :ghpull: `435`)
```

SylvainCorlay:
Thanks!

:warning: Quick heads up: when this gets released, we should not merge the conda-forge cf-auto-tick PR as-is.

We should use the opportunity to simplify the installation https://github.com/conda-forge/ipykernel-feedstock/blob/master/recipe/meta.yaml#L15-L18 since the kernelspec from will now include the proper path when installing from a source tarball.

docs/changelog.rst
SylvainCorlay: :+1:

SylvainCorlay: (can't commit the suggestion though)

blink1073:
TODO: add https://github.com/ipython/ipykernel/pull/437 and perhaps https://github.com/ipython/ipykernel/pull/411 if CI passes

blink1073:
`ipykernel-5.1.3-py3-none-any.whl` and ` ipykernel-5.1.3-py3-none-any.whl` are now available on PyPI.

PR #435, opened 2019-09-14
Set closed flag

impact27:
Avoid an issue where the comm sends close messages after being closed.


ccordoba12:
@impact27, is this still needed after PR #433?

impact27:
#433 fixes this problem when called from `__del__`. This would prevent the same problem from happening if the comm is closed in the closing callback.

jasongrout:
Thanks for this, @impact27. I just ran into this problem as well when closing a comm from the browser, and it's nice to know all I need to is upgrade!

PR #437, opened 2019-09-16
Moved InProcessKernelClient.flush to DummySocket

Chilipp:
Corrects the wrong placement of the flush method implemented in https://github.com/ipython/ipykernel/pull/405. Don't know how this has happened, but the method should obviously have been implemented for the `DummySocket` class, not for `InProcessKernelClient`...


blink1073:
Thanks!

PR #438, opened 2019-09-17
PR: Run GUI eventloop while waiting for input

impact27:
While waiting for the PY2 `raw_input` / PY3 `input` function, ipykernel is now completely blocked.
This is a problem for example while debugging. (Pdb uses `input` to get a command).
This PR would allow two new things to happen while waiting an input:
 - `self.do_one_iteration(`) would give a chance to process `comms` messages.
 - `manager.canvas.flush_events()` would give a change to any matplotlib backend to update.

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -448,7 +448,6 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
                                 user_ns=self.user_ns,
         )
         kernel.stdin_stream = stdin_stream
-        kernel._stdin_msg = None
 
         def handle_msg(msg):
             idents, msg = self.session.feed_identities(msg, copy=False)
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -433,7 +433,6 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
         """Create the Kernel object itself"""
         shell_stream = ZMQStream(self.shell_socket)
         control_stream = ZMQStream(self.control_socket)
-        stdin_stream = ZMQStream(self.stdin_socket)
 
         kernel_factory = self.kernel_class.instance
 
@@ -447,18 +446,7 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
                                 profile_dir=self.profile_dir,
                                 user_ns=self.user_ns,
         )
-        kernel.stdin_stream = stdin_stream
 
-        def handle_msg(msg):
-            idents, msg = self.session.feed_identities(msg, copy=False)
-            try:
-                msg = self.session.deserialize(msg, content=True, copy=False)
-            except:
-                self.log.error("Invalid Message", exc_info=True)
-                return
-            kernel._stdin_msg = msg
-
-        kernel.stdin_stream.on_recv(handle_msg, copy=False)
         kernel.record_ports({
             name + '_port': port for name, port in self.ports.items()
         })
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -433,6 +433,7 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
         """Create the Kernel object itself"""
         shell_stream = ZMQStream(self.shell_socket)
         control_stream = ZMQStream(self.control_socket)
+        stdin_stream = ZMQStream(self.stdin_socket)
 
         kernel_factory = self.kernel_class.instance
 
@@ -446,6 +447,19 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
                                 profile_dir=self.profile_dir,
                                 user_ns=self.user_ns,
         )
+        kernel.stdin_stream = stdin_stream
+        kernel._stdin_msg = None
+
+        def handle_msg(msg):
+            idents, msg = self.session.feed_identities(msg, copy=False)
+            try:
+                msg = self.session.deserialize(msg, content=True, copy=False)
+            except:
+                self.log.error("Invalid Message", exc_info=True)
+                return
+            kernel._stdin_msg = msg
+
+        kernel.stdin_stream.on_recv(handle_msg, copy=False)
         kernel.record_ports({
             name + '_port': port for name, port in self.ports.items()
         })
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -446,7 +446,6 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
                                 profile_dir=self.profile_dir,
                                 user_ns=self.user_ns,
         )
-
         kernel.record_ports({
             name + '_port': port for name, port in self.ports.items()
         })
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -41,7 +41,7 @@ from jupyter_client.session import Session
 
 from ._version import kernel_protocol_version
 
-from matplotlib import _pylab_helpers
+from matplotlib._pylab_helpers import Gcf
 
 CONTROL_PRIORITY = 1
 SHELL_PRIORITY = 10
@@ -892,7 +892,7 @@ class Kernel(SingletonConfigurable):
                     # Allow comms and other messages to be processed
                     self.do_one_iteration()
                     # Allow matplotlib figures with e.g. qt5 to update
-                    manager = _pylab_helpers.Gcf.get_active()
+                    manager = Gcf.get_active()
                     if manager is not None:
                         manager.canvas.flush_events()
 
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -926,7 +926,7 @@ class Kernel(SingletonConfigurable):
             self.eventloop(self)
             return self._stdin_msg
         else:
-            ident, reply = self.session.recv(self.stdin_socket)
+            ident, reply = self.session.recv(self.stdin_socket, 0)
             return reply
 
     def _at_shutdown(self):
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -13,6 +13,7 @@ from signal import signal, default_int_handler, SIGINT
 import sys
 import time
 import uuid
+import threading
 
 try:
     # jupyter_client >= 5, use tz-aware now
@@ -917,8 +918,15 @@ class Kernel(SingletonConfigurable):
 
     def _input_request_loop_step(self):
         """Do one step of the input request loop."""
-        # Allow matplotlib figures using GUI frameworks (e.g. qt, wx, gtk, tk) to update
-        if 'matplotlib.pyplot' in sys.modules:
+        # Allow matplotlib figures using GUI frameworks (e.g. qt, wx, gtk, tk)
+        # to update
+        if sys.version_info >= (3, 4):
+            is_main_thread = (threading.current_thread() is
+                              threading.main_thread())
+        else:
+            is_main_thread = isinstance(threading.current_thread(),
+                                        threading._MainThread)
+        if is_main_thread and 'matplotlib.pyplot' in sys.modules:
             # matplotlib needs to be imported after app.launch_new_instance()
             import matplotlib.pyplot as plt
             if plt.get_fignums():
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -41,10 +41,6 @@ from jupyter_client.session import Session
 
 from ._version import kernel_protocol_version
 
-try:
-    from matplotlib._pylab_helpers import Gcf
-except ImportError:
-    Gcf = None
 
 CONTROL_PRIORITY = 1
 SHELL_PRIORITY = 10
@@ -866,6 +862,13 @@ class Kernel(SingletonConfigurable):
         )
 
     def _input_request(self, prompt, ident, parent, password=False):
+        """Send an input request to the frontend and wait for the reply."""
+        # matplotlib needs to be imported after app.launch_new_instance()
+        try:
+            from matplotlib._pylab_helpers import Gcf
+        except ImportError:
+            Gcf = None
+
         # Flush output before making the request.
         sys.stderr.flush()
         sys.stdout.flush()
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -29,7 +29,7 @@ import zmq
 from zmq.eventloop.zmqstream import ZMQStream
 
 from traitlets.config.configurable import SingletonConfigurable
-from IPython.core.error import StdinNotImplementedError
+from IPython.core.error import StdinNotImplementedError, UsageError
 from ipython_genutils import py3compat
 from ipython_genutils.py3compat import unicode_type, string_types
 from ipykernel.jsonutil import json_clean
@@ -191,6 +191,26 @@ class Kernel(SingletonConfigurable):
             self._stdin_msg = msg
 
         self.stdin_stream.on_recv(handle_msg, copy=False)
+        # Should the eventloop be run while waiting for input
+        self._input_eventloop = False
+
+    def input_eventloop(self, active):
+        """
+        Activates and desactivates the eventloop while waiting for input.
+
+        active is "True" or "False" (Strings as it is used as a magic)
+
+        This allows eg. matplotlib plots to be used while debugging.
+
+        This should not be active while debugging a gui application that
+        uses the same eventloop as the events will be processed.
+        """
+        if active == "True":
+            self._input_eventloop = True
+        elif active == "False":
+            self._input_eventloop = False
+        else:
+            raise UsageError('Please use "True" or "False"')
 
     @gen.coroutine
     def dispatch_control(self, msg):
@@ -943,7 +963,7 @@ class Kernel(SingletonConfigurable):
         else:
             is_main_thread = isinstance(threading.current_thread(),
                                         threading._MainThread)
-        if is_main_thread and self.eventloop:
+        if is_main_thread and self.eventloop and self._input_eventloop:
             self.eventloop(self)
             return self._stdin_msg
         else:
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -900,15 +900,6 @@ class Kernel(SingletonConfigurable):
         ------
         KeyboardInterrupt if a keyboard interrupt is recieved.
         """
-        # matplotlib needs to be imported after app.launch_new_instance()
-        if 'matplotlib.pyplot' in sys.modules:
-            def flush_matplotlib():
-                import matplotlib.pyplot as plt
-                if plt.get_fignums():
-                    plt.gcf().canvas.flush_events()
-        else:
-            flush_matplotlib = None
-
         # Await a response.
         reply = None
         while reply is None:
@@ -917,19 +908,22 @@ class Kernel(SingletonConfigurable):
                     self.stdin_socket, zmq.NOBLOCK)
                 if not reply:
                     time.sleep(0.01)
-                    # Allow comms and other messages to be processed
-                    self.do_one_iteration()
-                    # Allow matplotlib figures with e.g. qt5 to update
-                    if flush_matplotlib:
-                        flush_matplotlib()
-
+                    self._input_request_loop_step()
             except Exception:
                 self.log.warning("Invalid Message:", exc_info=True)
             except KeyboardInterrupt:
                 # re-raise KeyboardInterrupt, to truncate traceback
                 raise KeyboardInterrupt
-
         return reply
+
+    def _input_request_loop_step(self):
+        """Do one step of the input request loop."""
+        # Allow matplotlib figures with e.g. qt5 to update
+        if 'matplotlib.pyplot' in sys.modules:
+            # matplotlib needs to be imported after app.launch_new_instance()
+            import matplotlib.pyplot as plt
+            if plt.get_fignums():
+                plt.gcf().canvas.flush_events()
 
     def _at_shutdown(self):
         """Actions taken at shutdown by the kernel, called by python's atexit.
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -917,7 +917,7 @@ class Kernel(SingletonConfigurable):
 
     def _input_request_loop_step(self):
         """Do one step of the input request loop."""
-        # Allow matplotlib figures with e.g. qt5 to update
+        # Allow matplotlib figures using GUI frameworks (e.g. qt, wx, gtk, tk) to update
         if 'matplotlib.pyplot' in sys.modules:
             # matplotlib needs to be imported after app.launch_new_instance()
             import matplotlib.pyplot as plt
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -41,7 +41,10 @@ from jupyter_client.session import Session
 
 from ._version import kernel_protocol_version
 
-from matplotlib._pylab_helpers import Gcf
+try:
+    from matplotlib._pylab_helpers import Gcf
+except ImportError:
+    Gcf = None
 
 CONTROL_PRIORITY = 1
 SHELL_PRIORITY = 10
@@ -892,9 +895,8 @@ class Kernel(SingletonConfigurable):
                     # Allow comms and other messages to be processed
                     self.do_one_iteration()
                     # Allow matplotlib figures with e.g. qt5 to update
-                    manager = Gcf.get_active()
-                    if manager is not None:
-                        manager.canvas.flush_events()
+                    if Gcf and Gcf.get_active():
+                        Gcf.get_active().canvas.flush_events()
 
             except Exception:
                 self.log.warning("Invalid Message:", exc_info=True)
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -863,12 +863,6 @@ class Kernel(SingletonConfigurable):
 
     def _input_request(self, prompt, ident, parent, password=False):
         """Send an input request to the frontend and wait for the reply."""
-        # matplotlib needs to be imported after app.launch_new_instance()
-        try:
-            from matplotlib._pylab_helpers import Gcf
-        except ImportError:
-            Gcf = None
-
         # Flush output before making the request.
         sys.stderr.flush()
         sys.stdout.flush()
@@ -906,6 +900,12 @@ class Kernel(SingletonConfigurable):
         ------
         KeyboardInterrupt if a keyboard interrupt is recieved.
         """
+        # matplotlib needs to be imported after app.launch_new_instance()
+        try:
+            from matplotlib._pylab_helpers import Gcf
+        except ImportError:
+            Gcf = None
+
         # Await a response.
         reply = None
         while reply is None:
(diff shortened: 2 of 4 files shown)


blink1073:
We can't depend on `matplotlib`.  Can we make that import optional?

blink1073:
@impact27 @ccordoba12 do we expect more PRs leading up to the Spyder release?

impact27:
> We can't depend on `matplotlib`. Can we make that import optional?

Sure, I am not sure how you would prefer to do that?
Alternatively we can just call `self.do_one_iteration()` and subclass it in `spyder-kernels`.

> @impact27 @ccordoba12 do we expect more PRs leading up to the Spyder release?

I don't expect more PRs but I was not really expecting this one.

impact27:
I just added a `try: except ImportError`

impact27:
> @impact27 @ccordoba12 do we expect more PRs leading up to the Spyder release?

We could port some of the improvments from spyder-kernels into ipykernel. We don't need to do that before the Spyder release. These improvments are mostly ipdb related. I don't know if you would be willing to have a ipdb subclass for ipykernel but the advantages that this could bring would be:
 - Autocomplete (technically already there but doesn't work in qtconsole)
 - ipython magic in ipdb
 - history in ipdb
 - multiline support in ipdb
 (Maybe some of these should be ported in IPython instead?)

ccordoba12:
@impact27, I don't how these improvements would work in the notebook, which also has support for `input` and hence ipdb.

That's the main issue you'll probably face when trying to move them upstream.

ccordoba12:
@impact27, did you test this with the inline backend? If so, could you post a screenshot about it?

impact27:
Inline:
Screenshot 2019-09-18 at 00 26 43
Qt5:
Screenshot 2019-09-18 at 00 28 06

impact27:
(The screenshots are from qtconsole)

ccordoba12:
Nice!! Thanks for posting them!

Pinging @tacaswell about this one because he'll probably be very interested about this new development.

ccordoba12:
And @SylvainCorlay could also be interested about this because I think the comms fix should allow using bqplot in debugging mode.

ipykernel/kernelbase.py
```diff
@@ -861,6 +862,13 @@ def raw_input(self, prompt=''):
         )
 
     def _input_request(self, prompt, ident, parent, password=False):
+        """Send an input request to the frontend and wait for the reply."""
+        # matplotlib needs to be imported after app.launch_new_instance()
+        try:
```
tacaswell: Better to do `if 'matplotlib' in sys.modules: ...` as this will import `matplotlib` if it is installed, but not yet imported and it is not the quickest import.

I am not stoked about importing a private module here, I would do something like

```python
if 'matplotlib.pyplot' in sys.modules:
    def flush():
        if plt.get_fignums():
            plt.gcf().canvas.flush_events()
else:
    flush = None
```

SylvainCorlay:
I am not really sure about this.

 - ipykernel should really not do anythings specific with respect to matplotlib (it currently does but we would like it not to)
 - Also, this may break the contract that messages should be processed in order...

One thing we may want to add to the protocol though is the ability to have a timeout option to the input request so that it does not hang forever...

ipykernel/kernelbase.py
impact27: I imported that as this is the same import made here: 
https://github.com/ipython/ipykernel/blob/d283077435138f6bb5d77624e887a2552574d773/ipykernel/pylab/backend_inline.py#L11
and here in `plt.pause`:
https://github.com/matplotlib/matplotlib/blob/5e9c54e7c4feb296b387fadb56394ef67ac2927f/lib/matplotlib/pyplot.py#L307
I am not sure if there is a subtle difference between `Gcf.get_active().canvas.flush_events()` and `plt.gcf().canvas.flush_events()`?

impact27:
>     * ipykernel should really not do anythings specific with respect to matplotlib (it currently does but we would like it not to)

Alternatively, this could just call `do_one_iteration` which can be modified to flush the matplotlib events somewhere else. (In ipykernel/pylab ?)

>     * Also, this may break the contract that messages should be processed in order...

The problem I am trying to solve is to process messages while waiting for a Pdb prompt (ideally I would process all events that are normally processed while waiting for a regular prompt. My understanding is that an alternative way to do that is to run the asyncio loop somehow.  I think that would require a large part of `pdb`, `bdb`, `cmd`, and `IPython.core.debugger` to be rewritten with coroutines.  We could also modify just `pdb.cmdloop` so this processing of messages only happens when called from pdb.

What message would actually be problematic to process here? Surely e.g. a completion request would be fine. I am trying to understand the potential problems as this is what we already do in spyder.

In `spyder-kernels` we currently call `do_one_iteration` explicitly after sending a request while debugging (such as a comms message or a completion request). It would probably be more elegant to have the kernel processing messages as it does while not debugging. This approach does not work for matplotlib plots, as they need to be updated continuously, which is the reason why we need to modify this function. 

Now that I think about it, maybe a better alternative would be to run the debugger (or even files) in another process. Threads wouldn't work because matplotlib and other libraries want to be on the main thread. This would also solve the problem of asyncio not running (see https://github.com/jupyter/notebook/issues/3397) but I think I am getting off topic...

> One thing we may want to add to the protocol though is the ability to have a timeout option to the input request so that it does not hang forever...

Well if I am debugging something and go for lunch, I would expect to be able to resume debugging when I get back. Therefore I don't really see any reasonable timeout time.

impact27:
> Also, this may break the contract that messages should be processed in order...

Is it possible to send a message with low priority? Like saying explicitely: It is ok to process other messages while this message is being processed?

ccordoba12:
> Also, this may break the contract that messages should be processed in order

If comm messages are queued, then `do_one_iteration` should process them in order, isn't it?

SylvainCorlay:
> > Also, this may break the contract that messages should be processed in order...
>
> Is it possible to send a message with low priority? Like saying explicitely: It is ok to process other messages while this message is being processed?

We need to think about possible rave conditions. It may as well be ok.

impact27:
> > Also, this may break the contract that messages should be processed in order
> 
> If comm messages are queued, then `do_one_iteration` should process them in order, isn't it?

I guess the problem is that they will start processing in order, but might not finish processing in order.

SylvainCorlay:
> If comm messages are queued, then do_one_iteration should process them in order, shouldn't it?
>

comm messages are just shell messages and are queued with other shell messages. If code is written with the assumption the input() is blocking, I presume this may as well break.

impact27:
As an alternative, I created https://github.com/ipython/ipykernel/pull/439 that would allow to change this behaviour easily.

ipykernel/kernelbase.py
tacaswell: `Gcf` is not part of the public API (not withstanding the other usage in IPython).

tacaswell:
The problem at the bottom here is that for the GUI to "look alive" you need to spin the input hook or is someother way give the GUI event loop some time to process it's events (user input, redrawing, etc).  Calling `flush_events` is a framework agnostic way that Matplotlib has to do that.  Another option is to sort out what GUI is running and spin it directly.

Way more details than anyone really wants at https://github.com/matplotlib/matplotlib/pull/4779

ipykernel/kernelbase.py
impact27: `plt.gcf().canvas.flush_events()` creates a new figure if no figures are open. Is there any way to avoid that?

impact27: `plt.get_current_fig_manager()` has the same problem. It creates an active figure as a side effect.

impact27: Sorrt I missed the `plt.get_fignums()`

impact27:
This PR didn't need the do_one_iteration call. It now just updates any open matplotlib figure, and adds a method that can be modified to add support to something else.

ipykernel/kernelbase.py
```diff
@@ -899,6 +892,38 @@ def _input_request(self, prompt, ident, parent, password=False):
...
+            except Exception:
+                self.log.warning("Invalid Message:", exc_info=True)
+            except KeyboardInterrupt:
+                # re-raise KeyboardInterrupt, to truncate traceback
+                raise KeyboardInterrupt
+        return reply
+
+    def _input_request_loop_step(self):
+        """Do one step of the input request loop."""
+        # Allow matplotlib figures with e.g. qt5 to update
```
tacaswell:
```suggestion
        # Allow matplotlib figures using GUI frameworks (e.g. qt, wx, gtk, tk) to update
```

tacaswell:
Will this hook ever be called from not the main thread?  GUI toolkits tend to behave very badly if you try to run their main loop from any thread but the main thread.

impact27:
> Will this hook ever be called from not the main thread? GUI toolkits tend to behave very badly if you try to run their main loop from any thread but the main thread.

You are right, I added a test to see if this is in the main thread.

ipykernel/kernelbase.py
```diff
@@ -899,6 +893,45 @@ def _input_request(self, prompt, ident, parent, password=False):
...
+        KeyboardInterrupt if a keyboard interrupt is recieved.
+        """
+        # Await a response.
+        reply = None
+        while reply is None:
+            try:
+                ident, reply = self.session.recv(
+                    self.stdin_socket, zmq.NOBLOCK)
+                if not reply:
+                    time.sleep(0.01)
```
tacaswell: why the sleep?

impact27: https://github.com/matplotlib/matplotlib/blob/ec2046f1253583a0eaee9617357b1e82bf41d0b0/lib/matplotlib/backend_bases.py#L2264

impact27: Because I don't want to use 100% of the processor while waiting for the input. the 0.01 comes from matplotlib: https://github.com/matplotlib/matplotlib/blob/ec2046f1253583a0eaee9617357b1e82bf41d0b0/lib/matplotlib/backend_bases.py#L2264

tacaswell: Is there anyway to push this timeout into the zmq recv logic?

tacaswell:
How does the GUI event loop integration work when not in pdb?  Presumably the same situation happens there (the kernel is waiting for a message on a zmq socket and needs to spin the GUI event loop while it waits).

ipykernel/kernelbase.py
impact27: http://api.zeromq.org/2-1:zmq-recv
There doesn't appear to be a way to do that

impact27:
> How does the GUI event loop integration work when not in pdb? Presumably the same situation happens there (the kernel is waiting for a message on a zmq socket and needs to spin the GUI event loop while it waits).

It uses asyncio: 
 - Get messages: `Kernel.start`: `self.io_loop.add_callback(self.dispatch_queue)`
 - Run GUI: `Kernel._update_eventloop`: `loop.add_callback(self.enter_eventloop)`

By design asyncio does not allow its event loop to be nested, so unless you are willing to hack the asyncio loop (https://pypi.org/project/nest-asyncio/) this wouldn't work here.

To solve the loop integration while in pdb, the call to `plt.gcf().canvas.flush_events()` can maybe be replaced by a call to `self.enter_eventloop()`.

tacaswell:
> matplotlib is another thing as the user could want to see plots while debugging.

matplotlib _is_ a GUI application

> By design asyncio does not allow its event loop to be nested,

But the GUI event loop is completely independent from the asyncio event loop (in the Qt case it is implemented in c++ is not aware python exists).

It looks like the hooks are defined in https://github.com/ipython/ipykernel/blob/master/ipykernel/eventloops.py, in particular, https://github.com/ipython/ipykernel/blob/d0ef2e40157c8a8f071818b074650a2f0b81f878/ipykernel/eventloops.py#L26-L53 is the key function for Qt (there are versions for others too).  This lets the QApplication event loop run forever, but registers a callback to kill it as soon as anything comes across the ZMQ stream.  Tracing this through it looks these functions get called at https://github.com/ipython/ipykernel/blob/f0c72af1590dc56e0044d9563f5a8dc20c34b1a6/ipykernel/kernelbase.py#L309-L314 which effectively blocks the main thread while it waits for something to come across any of the streams.  Also looks like that there is a 1s dead time where the GUI will be non-responsive?

I think that even in debugging mode the communication with the user is still happening over zmq?  In that case it should be possible to re-use the same "block until data" functions and avoid poling?  However I am a bit under-educated on the number / types of input streams that the kernel is listening to.

tacaswell:
Sorry I am being difficult (and slow) about this, but I think getting the event loop / inputhook integration correct is important.

impact27:
> matplotlib _is_ a GUI application

While debugging, I wouldn't want anything to execute while the program is stopped, so I wouldn't want to run the GUI eventloop.
That being said, plotting graphs while debugging is very useful to understand what is going wrong.

So the options are:

 - Not running any event loops: I can debug anything but I can't use plots.
 - Running only matplotlib: I can see plots while debugging but if I am trying to debug the matplotlib library itself, I might run into problems.
 - Running all GUI event loops: Trying to debug any GUI applications might be problematic.

> But the GUI event loop is completely independent from the asyncio event loop (in the Qt case it is implemented in c++ is not aware python exists).

Well the event loop is called from asyncio.

>  but registers a callback to kill it as soon as anything comes across the ZMQ stream. 

Only the shell streams. It is not stopped when recieving a reply from iopub. This could be modified by adding  a `ZMQStream ` (`from zmq.eventloop.zmqstream import ZMQStream`) around `self.stdin_socket` and then calling `_notify_stream_qt` (and equivalent for other loops).

impact27:
I would say option 2 (which is implemented now) is the best trade-off.

tacaswell:
I am not clear on what the distinction between "running only Matplotlib" and "running all GUI event loops" is.  `flush_events` is more-or-less "clear all of the events currently enqueued at the GUI event loop and then return" so any new things added to the queue won't be processed.  Running that in a hot polling loop still means that all of the GUI related code will eventually run, but we are putting in some latency to handling both GUI events and network I/O.  Using the same code that the "normal" code path means that the GUI events still get serviced (but with less of a lag) and the network I/O gets serviced with no more latency than in normal operations.

> Well the event loop is called from asyncio.

There are (at least) 2 event loops involved in this operation (the asyncio event loop running tornado and the GUI event loop).  When the gui event loop is running the asyncio task, and hence the whole asyncio event loop, is blocked.  The `_notify_stream_qt` is to time-share two infinite loops on the main thread 

> It is not stopped when recieving a reply from iopub.

How is iopub serviced then?  When the GUI event loop is running it is blocking the main thread.  Also, what is `iopub`?


Option 3 has the advantages that now the behavior in debugging and in normal mode are the same and there is no explicit Matplotlib dependency or configuration discovery (if the kernel has an `event_loop` attribute use it, else don't).  However, that may be monitoring too many streams?

impact27:
iopub is described here: https://jupyter-client.readthedocs.io/en/stable/messaging.html. Is is serviced by calling `self.session.recv( self.stdin_socket)` in a while loop in `def _wait_input_request_reply(self):`. It is not integrated in any event loops.

impact27:
I think you are right. I have updated the PR to run the eventloop

ipykernel/kernelbase.py
```diff
@@ -899,6 +894,41 @@ def _input_request(self, prompt, ident, parent, password=False):
...
+                self.log.warning("Invalid Message:", exc_info=True)
+            except KeyboardInterrupt:
+                # re-raise KeyboardInterrupt, to truncate traceback
+                raise KeyboardInterrupt
+        return reply
+
+    def _input_request_loop_step(self):
+        """Do one step of the input request loop."""
+        # Allow GUI event loop to update
+        if sys.version_info >= (3, 4):
```
ccordoba12: I think the minimal supported version in the master branch is 3.5 (please check that in `setup.py`)

impact27: This doubles as a test for python 2

tacaswell:
This seems worse as now changes only for debugging are leaking into the main processing.

I am confused by trying to understand what you are saying @impact27 vs what the docs say.

`iopub` looks like it is a publish only zmq socket responsible for pushing results _to_ clients (so we never need to worry about it interrupting a GUI event loop as the kernel controls when that goes out).  

On the other hand, it looks like the `sdtin_socket` is a router we do need to recv from and is probably the _only_ stream we should be pulling on while debugging (which is what I think @SylvainCorlay was saying above but I did not yet understand).  We may have to re-write how the event_loop functions work to allow passing the streams to monitor in...

Returning the message via mutable state looks like it is going to be the source of terrible to debug race-conditions....

impact27:
Sorry I mixed `iopub` and `sdtin`.

> This seems worse as now changes only for debugging are leaking into the main processing.

Could you elaborate on that? I am not sure I understand what you mean.

> is probably the only stream we should be pulling on while debugging

Well pulling on the other streams is not a problem, as it just adds the messages in a queue. The problem would be processing the messages, which we don't.

> Returning the message via mutable state looks like it is going to be the source of terrible to debug race-conditions....

The `self.stdin_socket` is only used in this function, so I am not sure how a race condition can occur. Are you worried about two threads calling `input` at the same time? This doesn't sound likely and would already result in a race condition now.

tacaswell:
> Could you elaborate on that? I am not sure I understand what you mean.

This PR started as a change to just the debugging loop and has grown to change all of the event loop code.

> This doesn't sound likely and would already result in a race condition now.

I don't know this code particularly well (:wave: from then :peanuts: gallery) so I believe you it is OK, but that sort of non-local return is a code smell that gives me pause.  I am frequently surprised by the race conditions that come up in network related code so adopt a position of mild paranoia.

It looks like this is registering a callback onto the zmq socket to stash the message when it comes in.  Are we sure of the exact ordering of when that callback gets run vs when the GUI event loop will exit?  I would believe that sometimes the order goes [data comes in -> callback fires -> event loop exits :tada: ] but that other times the order goes [data comes in -> event loop exits -> callback fires] but now there is a chance that we return the last message.  I could also imagine that if data comes in fast we would have [data comes in -> callback fires -> data comes in -> callback files -> event loop exits] where we have now dropped a message.

You said that for the other messages we are dumping them onto a queue rather than processing them, we probably should be doing the same with the stdin messages.  Something like the callback adds them to the queue, the event loop dies, we than pull on that queue until it is empty and then restart the event loop?

Sorry I am being so slow about my side of this conversation and I could be misunderstanding how the kernel code works under the hood.

impact27:
> This PR started as a change to just the debugging loop and has grown to change all of the event loop code.

It doesn't change the behaviour of rest of the code. The only difference to the event loop code is that it now checks for messages on the stdin stream as well, but it doesn't do anything with these.

> I am frequently surprised by the race conditions that come up in network related code so adopt a position of mild paranoia.

Sounds reasonable :) I quite agree with you. I am also a bit paranoid about threads. Should I add a lock around the code?

>  Are we sure of the exact ordering of when that callback gets run vs when the GUI event loop will exit? I would believe that sometimes the order goes [data comes in -> callback fires -> event loop exits 🎉 ] but that other times the order goes [data comes in -> event loop exits -> callback fires] 

If I am not mistaken the callback is called by the event loop, so the event loop has to call the callback before exiting. It that was not the case this PR wouldn't work to solve the problem in https://github.com/spyder-ide/spyder/issues/620 (and it does.)

>  I could also imagine that if data comes in fast we would have [data comes in -> callback fires -> data comes in -> callback files -> event loop exits] where we have now dropped a message.

That would be a bug in the frontend. We send a single input request, so we are expecting a single input reply. And the code to quit the event loop is in the callback, so I am almost sure this wouldn't happen?

 > You said that for the other messages we are dumping them onto a queue rather than processing them, we probably should be doing the same with the stdin messages.

Should we? I am not sure it makes sense to recieve more than 1 message in stdin. The way it works is that we send an input_request and then wait for -one- input reply. Any other message on stdin should be discarded anyway, as they are a reply without a request.

impact27:
This breaks the code. I need a Lock.
```
import threading
import time

def getinput():
    time.sleep(1)
    print(input('thread:'))

thread = threading.Thread(target=getinput)
thread.start()
print(input('main:'))
```

ipykernel/kernelapp.py
```diff
@@ -446,6 +447,18 @@ def init_kernel(self):
...
+        def handle_msg(msg):
+            idents, msg = self.session.feed_identities(msg, copy=False)
+            try:
+                msg = self.session.deserialize(msg, content=True, copy=False)
+            except:
+                self.log.error("Invalid Message", exc_info=True)
+                return
+            kernel._stdin_msg = msg
+
+        kernel.stdin_stream.on_recv(handle_msg, copy=False)
```
ccordoba12: I think all these new lines here should be moved to the `Kernel` class (present in `kernelbase.py`). In particular, `handle_msg` should be part of `dispatch_stdin` (following what `dispatch_control` and `dispatch_shell` do for `control_stream` and `shell_streams`, respectively).

However, I don't know what the implications of adding a new stream here are for the Jupyter protocol. Pinging @takluyver for his insights about that.

impact27: I moved it. I don't want to process the messages recieved on stdin so I don't think `dispatch_stdin` is a good idea. We are not dispatching the messages, only saving the most recent. A stream is just a wrapper around a socket. I am not adding a socket, so the protocol is unchanged. The only reason I need a stream around the socket is to get a qt event when a message is recieved.

tacaswell:
https://github.com/ipython/ipython/issues/5026 this popped up in my feed today and seems at least a bit relevant.

impact27:
I think https://github.com/ipython/ipython/issues/5026 is a front end problem: the kernel issues a input request but the frontend doesn’t know what to do with it. This PR is not really related as running the GUI event loop doesn’t really help here.

tacaswell:
Mostly race conditions / failure modes in `raw_input`.

impact27:
There is a thread lock now so this shouldn’t happen

ipykernel/kernelapp.py
takluyver: Yes, AFAIK @impact27 is right - there shouldn't be a difference to the protocol. A pyzmq stream is a wrapper around a socket to tie it to an event loop. The stdin socket already existed, so whether the code wraps it in a stream or not should be an implementation detail.

(I haven't tried to review the PR, just answering the question I was mentioned for)

ipykernel/kernelbase.py
```diff
@@ -170,6 +172,26 @@ def __init__(self, **kwargs):
...
+        self.stdin_stream = ZMQStream(self.stdin_socket)
+
+        def handle_msg(msg):
+            idents, msg = self.session.feed_identities(msg, copy=False)
+            try:
+                msg = self.session.deserialize(msg, content=True, copy=False)
+            except Exception:
+                self.log.error("Invalid Message", exc_info=True)
+                return
+            self._stdin_msg = msg
```
tacaswell: Don't we need to lock here too?

ipykernel/kernelbase.py
```diff
@@ -899,6 +915,41 @@ def _input_request(self, prompt, ident, parent, password=False):
...
+        # Allow GUI event loop to update
+        if sys.version_info >= (3, 4):
+            is_main_thread = (threading.current_thread() is
+                              threading.main_thread())
+        else:
+            is_main_thread = isinstance(threading.current_thread(),
+                                        threading._MainThread)
+        if is_main_thread and self.eventloop:
+            self.eventloop(self)
+            return self._stdin_msg
```
tacaswell: and possibly here?

impact27: `_input_request_loop_step` is called from `_wait_input_request_reply` which is protected by a lock in `_input_request`, so I don't think a lock here would do anything.

ipykernel/kernelbase.py
impact27: The messages are processed by a single thread so a lock would not be useful here.

impact27:
I have added an option to enable this behaviour. I think it should be off by default for the following reason:

 - If I am debugging an application that uses the same eventloop as matplotlib, the debugger won't stop the program, as the eventloop will still be running!

Should matplotlib be called explicitly to avoid this problem as was the case earlier in this PR?

impact27:
I opened #465 as an alternative that only updates matplotlib.

impact27:
I tried to add a `%input_eventloop` magic to enable / disable this. What do you think of that?
The idea is that if you want matplotlib plots while debugging, you must explicitly activate it. If this is activated by default it might be very counter-intuitive for a user trying to debug a gui application.

jkablan:
Hi, I'm trying to use your updates to plot while debugging.  I've installed it by doing the following.  
1. Cloning the update Update_comms_matplotlib branch 
2. running `pip install -e . `

I then start ipython and attempt to run your magic command 

`%input_eventloop`

which returns the following.

`UsageError: Line magic function '%input_eventloop' not found.`

If I try plot anyways. I have to run `plt.pause(10)` for figures to show up.

Can you give me an idea of what I'm doing wrong? Also wondering if there's a timeline for integrating this back to master? Lack of this functionality is a serious productivity killer for those of us who are coming from MATLAB.

impact27:
> Can you give me an idea of what I'm doing wrong? Also wondering if there's a timeline for integrating this back to master? Lack of this functionality is a serious productivity killer for those of us who are coming from MATLAB.

`%input_eventloop` should work. If it doesn't, it might mean the installation failed somehow. You can try:
```
import ipykernel
print(ipykernel.__file__)
```
And check if the file is indeed the one you are expecting.

About the timeline, the problem is that there is a trade-off here: If this is implemented, any app that uses an event loop (such as matplotlib) can keep running while the debugger is stopped. This is good because you can plot things but bad if you are trying to debug a gui app. 

This trade-off means it is not trivial to merge this PR. My idea was to leave this behaviour disabled by default but give an option to enable it (with the `%input_eventloop` magic), which might make it acceptable.

Another solution for this problem would be to modify Pdb so it uses a custom method for blocking the code. This would be quite a big change and a lot of work.

jkablan:
Thanks for getting back to me so quickly.
`print(ipykernel.__file__)`
returns 
`C:\git\projects\temp\ipykernel\ipykernel\__init__.py`

This is indeed the directory I cloned your respo into. The head of my clone is pointed to 'd16d6e4e40f68f384d57a1a4458d4ef848e4ceb5' which is from about a week ago. Should I be using this commit? 
 
Introducing the magic command to allow the gui event loop to keep running seems like a very reasonable solution to me given the amount of work it would be to modify pdb. I guess I can't see what about this solution would prevent it from being merged.

impact27:
I am not convinced this is the right approach. This might introduce unwanted side effects.
In Spyder I solve this problem by:
```
import cmd
cmd.input = patched_input
```
and then running eventloop from this function. This means the eventloop is only run when called from pdb instead of any input calls. The main thing that would be nice to have is a way of interrupting the eventloop. The workaround I found is to send a dummy message on the shell channel, but having a `kernel.eventloop.stop()` function would be nice.

impact27:
I am going to close this for now as I am not convinced this is the right approach. What I am afraid of is that someone will use input() to try and block the execution. If the eventloop is still running, some code could be executed by the eventloop. This could make it hard to debug qt apps for example. Of course this is only an edge case but there should be at least an easy way of disabling this.

felipeespic:
I recall that this problem wasn't happening with IPython 7.7. Actually, I also remember that I didn't update IPython for a very long time just to avoid braking the integration between ipdb and Matplotlib.

felipeespic:
I just want to add that the issue with Matplotlib is not happening with **pdb**, only with **ipdb**.
For reference:
pdb in Python 3.8.10
ipdb 0.13.9 in IPython 8.6.0

PR #439, opened 2019-09-19
[Depends on #445] PR: Create _wait_input_request_reply method

impact27:
Move the code waiting for an input request reply to a method to make it easier to change this behaviour in a subclass.


impact27:
Merged with #438

PR #441, opened 2019-09-21
PR: Disable AppNap

impact27:
solves https://github.com/ipython/ipython/issues/11602

This is caused by AppNap: https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/AppNap.html
This can be desactivated by https://pypi.org/project/appnope/

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -154,12 +154,6 @@ def loop_wx(kernel):
 
     import wx
 
-    if _use_appnope() and kernel._darwin_app_nap:
-        # we don't hook up App Nap contexts for Wx,
-        # just disable it outright.
-        from appnope import nope
-        nope()
-
      # Wx uses milliseconds
     poll_interval = int(1000 * kernel._poll_interval)
 
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -17,6 +17,9 @@ from .comm import CommManager
 from .kernelbase import Kernel as KernelBase
 from .zmqshell import ZMQInteractiveShell
 
+if sys.platform == 'darwin':
+    import appnope
+
 try:
     from IPython.core.interactiveshell import _asyncio_runner
 except ImportError:
@@ -78,6 +81,10 @@ class IPythonKernel(KernelBase):
         comm_msg_types = [ 'comm_open', 'comm_msg', 'comm_close' ]
         for msg_type in comm_msg_types:
             self.shell_handlers[msg_type] = getattr(self.comm_manager, msg_type)
+
+        if sys.platform == 'darwin':
+            # Disable app-nap as the kernel is not a gui but can have guis
+            appnope.nope()
 
     help_links = List([
         {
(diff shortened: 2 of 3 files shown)


ccordoba12:
This could be done only at the event loops level and not for the entire application. At least that's how it was done in old IPython versions:

https://github.com/ipython/ipython/blob/a2685d78f2403f84e6cb915ae11a4f6033ccc5f6/IPython/lib/inputhook.py

impact27:
I disagree, the problem here is not limited to the back ends. If the user creates a gui app (qt5 / cocoa/ ...), macos will reclassify the kernel as a GUI application. Therefore, If the kernel (which is not a GUI application intrinsically) shows any window at any point, then macos will start to slow it down. Therefore AppNap should not apply to ipykernel. (As it uses window usage as a way to know if the app is used.)

impact27:
I am not sure what the error is. I see the error intermittently in spyder-kernels tests as well.
in `with setup_kernel(cmd) as client:`, somehow the connection file is empty? `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

impact27:
Retriggering the tests

ccordoba12:
@impact27, you should really avoid your practice of mixing PRs because it makes reviewing them a nightmare.

So please remove your last commit here and in PRs #438 and #439, and instead please try to make PR #445 pass, so it can be merged first.

impact27:
> @impact27, you should really avoid your practice of mixing PRs because it makes reviewing them a nightmare.
> 
> So please remove your last commit here and in PRs #438 and #439, and instead please try to make PR #445 pass, so it can be merged first.

I am sorry that this makes it hard for you to review. It is my fault for merging the PRs before #445 was ready.

I will mark all the PR that depends on another PR as [Waiting depencency merge] so you know that they must not be reviewed before the dependent PR is merged.

impact27:
The reason why I need to merge #445in the other PR is that AppVeyor gets stuck if #445 is not merged, so instead of having test suceeding in 10 min, there is failing tests in 3h.

ccordoba12:
> If the user creates a gui app (qt5 / cocoa/ ...), macos will reclassify the kernel as a GUI application

Even if there you haven't run `%gui qt5` before creating the app?

impact27:
I was able to trigger the problem purely with pyqt5

impact27:
#!/usr/bin/env python3

import time
# Uncomment either
from PySide2 import QtWidgets, QtCore
# from PyQt5 import QtWidgets, QtCore

app = QtWidgets.QApplication([''])
timer = QtCore.QTimer(app)
timer.setSingleShot(True)
timer.timeout.connect(lambda: app.quit())
timer.start(35000)
app.exec_()
start = time.time()
time.sleep(0.0005)
print('dt = ', time.time() - start)

ccordoba12:
> I was able to trigger the problem purely with pyqt5

Ok, if that's the case then this should be merged as it is.

blink1073:
Thanks!

PR #443, opened 2019-09-27
ensure control stream is flushed before processing shell messages

minrk:
previous flush was not handled at the right time with new async dispatch to ensure control stream priority

Affected case:

- shell request 1 takes a long time to process, is blocking (no async)
- shell request 2 arrives, is received and queued
- control request arrives later, but before shell request 1 is processed, and has not been queued thanks to no `awaits`

It's possible that shell request 2 is in our Python queue, but control request is still waiting one ioloop iteration
to be handled from zmq, and we want to make sure that we still process the control message first.

This is causing test failures in IPython parallel because task abort relies on proper queue priority handling.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -225,10 +225,6 @@ class Kernel(SingletonConfigurable):
     @gen.coroutine
     def dispatch_shell(self, stream, msg):
         """dispatch shell requests"""
-        # flush control requests first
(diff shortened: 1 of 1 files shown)


minrk:
@blink1073 Since this is causing failures in IPython parallel, I'd be happy to see this in 5.1.3 (#411), but I'm also okay if that ship has sailed.

PR #444, opened 2019-10-01
Allow pyside2 gui loop

### LINKED ISSUE #398: os.environ['QT_API'] gets overwritten in event-loop. No pyside2 integration.

bergtholdt:
Removed writing to the environment inside GUI loop.

Fixes #398 

Makes #399 obsolete.

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -126,7 +126,15 @@ def loop_qt4(kernel):
 def loop_qt5(kernel):
     """Start a kernel with PyQt5 event loop integration."""
     if os.environ.get('QT_API', None) is None:
-        os.environ['QT_API'] = 'pyqt5'
+        try:
+            import PyQt5
+            os.environ['QT_API'] = 'pyqt5'
+        except ImportError:
+            try:
+                import PySide2
+                os.environ['QT_API'] = 'pyside2'
+            except ImportError:
+                pass
     return loop_qt4(kernel)
 
 
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -125,7 +125,6 @@ def loop_qt4(kernel):
 @register_integration('qt', 'qt5')
 def loop_qt5(kernel):
     """Start a kernel with PyQt5 event loop integration."""
-    os.environ['QT_API'] = 'pyqt5'
     return loop_qt4(kernel)
 
 
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -134,7 +134,7 @@ def loop_qt5(kernel):
                 import PySide2
                 os.environ['QT_API'] = 'pyside2'
             except ImportError:
-                pass
+                os.environ['QT_API'] = 'pyqt5'
     return loop_qt4(kernel)
 
 
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -125,6 +125,8 @@ def loop_qt4(kernel):
 @register_integration('qt', 'qt5')
 def loop_qt5(kernel):
     """Start a kernel with PyQt5 event loop integration."""
+    if os.environ.get('QT_API', None) is None:
+        os.environ['QT_API'] = 'pyqt5'
     return loop_qt4(kernel)
 
 


ipykernel/eventloops.py
```diff
@@ -125,7 +125,6 @@ def loop_qt4(kernel):
 @register_integration('qt', 'qt5')
 def loop_qt5(kernel):
     """Start a kernel with PyQt5 event loop integration."""
-    os.environ['QT_API'] = 'pyqt5'
```
ccordoba12: If this is not set, then there's no way to select the right Python bindings to create the event loop (unless users set `QT_API` by hand).

Therefore, I'm -1 on this one.

bergtholdt: OK, changed it to set if unset. This also prevents it from being overwritten in every loop (as it was originally).

ccordoba12:
Ok, this is much better now, thanks! The only problem I see with this is that users still need to set `QT_API='pyside2'` by hand before running `%gui qt5`, even if PySide2 is the only Qt binding installed in your conda/virtual env.

@bergtholdt, do you think we could improve that situation?

bergtholdt:
OK I changed it to use pyqt5 as default and pyside2 as fall-back. That should not affect anything that worked before.

A little improvement, in the case that both backends are available, could be done to check if either module has been imported before calling %gui qt5 by checking sys.modules, but I would suspect that in that case the user sets the variable herself ("explicit is better than implicit").

ccordoba12:
> A little improvement, in the case that both backends are available, could be done to check if either module has been imported before calling %gui qt5 by checking sys.modules, but I would suspect that in that case the user sets the variable herself ("explicit is better than implicit").

Ok, let's go with this simple approach for now and see how users respond to it.

ipykernel/eventloops.py
```diff
@@ -125,7 +125,16 @@ def loop_qt4(kernel):
...
+    if os.environ.get('QT_API', None) is None:
+        try:
+            import PyQt5
+            os.environ['QT_API'] = 'pyqt5'
+        except ImportError:
+            try:
+                import PySide2
+                os.environ['QT_API'] = 'pyside2'
+            except ImportError:
+                pass
```
ccordoba12: I think we should set `QT_API='pyqt5'` at this point to have the same behavior as before?

ivanov: I'm super necromancing this PR - but are there circumstances where there's any effect to setting `QT_API='pyqt5'` even though both   `import PyQt5` and `import pyside2` have failed? If so, then we can add that in place of the path, but I'm thinking maybe this is OK as is?

ivanov:
Hey there, @bergtholdt, thank you for your contribution. I'm going through old PRs and this one has stalled.

I'm going to power cycle this one (close it and reopen it) just so we get the latest set of tests to rerun on it, and once I hear back on my the last remaining question to @ccordoba12 (or anyone else knowledgeable in QT stuff), then we can merge it.

ivanov:
@ccordoba12 and @stukowski - does 2d15ed6a4aad182fd0ba276cd7058c90155bfba3 look like I'm doing the thing you were talking about to keep with previous behavior? I'll wait to merge until I hear back or someone more competent shows up :smile:

ccordoba12:
> does 2d15ed6 look like I'm doing the thing you were talking about to keep with previous behavior?

Yeah, I think it does. I mean, it fallbacks to setting `QT_API` to `pyqt5` if we fail to import PySide2, which should preserve the previous behavior at the very least.

So I'm ok with merging this one now.

ivanov:
Perfect, thank you, Carlos!

ivanov:
and thank you @bergtholdt  for writing this up in the first place, apologies that it took so long to get merged in.

ccordoba12:
> Perfect, thank you, Carlos!

No prob, and thanks to you for persevering until this PR was merged.

PR #446, opened 2019-10-05
PR: Wait until connection file is valid

impact27:
Alternative to #445 that doesn't use flaky.

Wait until connection_file exists AND is a readable json file.

### DIFF
--- a/ipykernel/tests/test_embed_kernel.py
+++ b/ipykernel/tests/test_embed_kernel.py
@@ -54,6 +54,9 @@ def setup_kernel(cmd):
             and time.time() < tic + SETUP_TIMEOUT:
             time.sleep(0.1)
 
+        # Wait 100ms for the writing to finish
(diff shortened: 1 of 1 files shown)


impact27:
I'll retrigger the tests a few times to see how stable they are.

ccordoba12:
> I'll retrigger the tests a few times to see how stable they are.

This is the same thing as using flaky, isn't it?

impact27:
> > I'll retrigger the tests a few times to see how stable they are.
> 
> This is the same thing as using flaky, isn't it?

Well it looks like the tests in PR #445 are sometimes failing as well, even with flaky, but yes.

ccordoba12:
> Well it looks like the tests in PR #445 are sometimes failing as well, even with flaky, but yes.

Then this one should be closed and `flaky` should be added as a new dependency instead.

PR #456, opened 2019-11-12
workaround tornado+py38+windows compatibility issue

minrk:
set eventloop policy to the old default while tornado is not compatible with the new one

same as https://github.com/jupyter/notebook/pull/5047

cross-ref: https://github.com/tornadoweb/tornado/issues/2608

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -514,13 +514,45 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
         handler.setFormatter(formatter)
         logger.addHandler(handler)
 
+    def _init_asyncio_patch(self):
+        """set default asyncio policy to be compatible with tornado
+
+        Tornado 6 (at least) is not compatible with the default
+        asyncio implementation on Windows
+
+        Pick the older SelectorEventLoopPolicy on Windows
+        if the known-incompatible default policy is in use.
+
+        do this as early as possible to make it a low priority and overrideable
+
+        ref: https://github.com/tornadoweb/tornado/issues/2608
+
+        FIXME: if/when tornado supports the defaults in asyncio,
+               remove and bump tornado requirement for py38
+        """
+        if sys.platform.startswith("win") and sys.version_info >= (3, 8):
+            import asyncio
+            try:
+                from asyncio import (
+                    WindowsProactorEventLoopPolicy,
+                    WindowsSelectorEventLoopPolicy,
+                )
+            except ImportError:
+                pass
+                # not affected
+            else:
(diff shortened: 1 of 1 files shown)


bollwyvl:
Since this patch would need to be in _two_ places in `notebook` (in `notebookapp.py`, and in `launchnotebook.py`) and here in `ipykernel`, and very possibly in more places (anything with a threadpoolexecutor that also uses async... nbviewer, nbformat tests), should this not be hoisted up to `jupyter_core`?

It would likely also benefit us to get on github actions or azure pipelines so we are actually testing on windows... I wouldn't want to new start anything on appveyor at this point: the one-at-a-time free queue is just brutal, _especially_ for windows excursions.

minrk:
> should this not be hoisted up to jupyter_core?

No, I don't think so. That would change the jupyter-core dependencies to include tornado, bump the minimum jupyter-core requirement in both packages, etc. It's a tiny patch and I think is better to duplicate than import from a common dependency, which separates the solution from where the problem is encountered.

bollwyvl:
Fair points. Either way, will probably just have to keep making the rounds
of everything and shoring up windows ci to make sure about this. The
acceleration of the Python release cycle has the potential to make these
issues more acute in the future.

minrk:
Quite true. Thanks for expanding the CI coverage!

kevin-bates:
Thanks @minrk for the patch and @bollwyvl for the appveyor efforts - both have proven quite helpful to me.  We're going to need the patch in jupyter_server and jupyter_kernel_mgmt and both projects will require an updated ipykernel for their tests as well.  

Might there be an ETA on when an updated ipykernel can get released?  I'm currently referencing Min's branch in jupyter_kernel_mgmt's appveyor.yml (PR 34), so even a merge into master would be helpful.

jamesmyatt:
I assume this fix hasn't been released yet. Is there any timescale for this?

I'm hitting it now that notebook v6.0.3 has been released (https://github.com/jupyter/notebook/issues/4613).

kevin-bates:
Good point!  The NB 6.0.3 drop is going to cause some major grief in new (and default) configurations on Windows/py38 systems.  We probably should have ensured ipykernel was released prior to 6.0.3, but hindsight is 20/20 (hey, its the year of hindsight!).

@blink1073 - it looks like you've done the last few releases.  Any chance you (or someone with applicable privs) could promote this closer to the top of your "todos"?  If necessary, I could try to put together the change log entries if that helps.  I'm assuming this release would be 5.1.4 rather than 5.2.0.

minrk:
Apologies for the delay, 5.1.4 is out now with the fix.

PR #465, opened 2019-12-05
PR: [alternative to  #438] Update matplotlib figures while waiting for input

impact27:
This PR process events for any open matplotlib figure while waiting for an input.
This allow interaction with matplotlib figures while debugging.
Unlike #438, it doesn't process any other events that might come from a gui app, which would make debugging these apps difficult.


tacaswell:
Under the hood `flush_events` will spin the GUI event loop and drain all events, not just the ones specific to Matplotlib.

impact27:
Thanks I didn’t know that

PR #478, opened 2020-01-16
PR: Flush from process even without newline

impact27:
Solves a bug where calling `print(text, end='')` from `mp.Pool` doesn't print any text until the processing is finished.
eg:
```
import time
from multiprocessing import Process


def f(name):
    print('hello', end='')
    time.sleep(2)


if __name__ == '__main__':
    p = Process(target=f, args=('bob',))
    p.start()
    p.join()
```
only shows `hello` after 2 seconds and not immediately as expected.

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -406,6 +406,8 @@ class OutStream(TextIOBase):
                 # and this helps.
                 if '\n' in string:
                     self.flush()
+                else:
+                    self.pub_thread.schedule(self._flush)
             else:
                 self._schedule_flush()
 
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -296,6 +296,7 @@ class OutStream(TextIOBase):
         self.parent_header = {}
         self._master_pid = os.getpid()
         self._flush_pending = False
+        self._subprocess_flush_pending = False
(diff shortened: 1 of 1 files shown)


ipykernel/iostream.py
```diff
@@ -406,6 +406,8 @@ def write(self, string):
                 # and this helps.
                 if '\n' in string:
                     self.flush()
+                else:
+                    self.pub_thread.schedule(self._flush)
```
Carreau: I'm not too familiar with this code but should this be `self._schedule_flush()` to avoid excessive flush ?

impact27: That doesn't work. I am not 100% sure why but I think thwere is no `self._io_loop` in the subprocess.

impact27: Should `self.flush()` really be called on the line above? I am thinking that maybe I should replace
```
                if '\n' in string:
                    self.flush()
                else:
                    self.pub_thread.schedule(self._flush)
```
by:
`self.pub_thread.schedule(self._flush)`. (And maybe add a check on `self._flush_pending`)
But I am quite sure I can not use the IO thread.

impact27: Please see the latest commit

PR #480, opened 2020-01-26
Workaround tornado+py38+windows compatibility issue 

ericpre:
The example [inprocess_qtconsole.py](https://github.com/ipython/ipykernel/blob/master/examples/embedding/inprocess_qtconsole.py) is not working with python 3.8 on windows and this PR make it work.

Similar fix as #456: set eventloop policy to the old default while tornado is not compatible with the new one.

I don't follow how this works, so I don't know if this is the right fix (just tried to work out where to put the fix from #456) but at least it should highlight where the issue is! ;)

### DIFF
--- a/examples/embedding/inprocess_qtconsole.py
+++ b/examples/embedding/inprocess_qtconsole.py
@@ -2,8 +2,8 @@ from __future__ import print_function
 import os
 import sys
 
-from IPython.qt.console.rich_ipython_widget import RichIPythonWidget
-from IPython.qt.inprocess import QtInProcessKernelManager
+from qtconsole.rich_ipython_widget import RichIPythonWidget
+from qtconsole.inprocess import QtInProcessKernelManager
 from IPython.lib import guisupport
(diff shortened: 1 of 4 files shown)


minrk:
Thanks for this PR! I don't believe the patch belongs in the base IPythonKernel object, which is usually instantiated too late for this to work (the example is pretty minimal, in which case this is early enough, but most 'real' applications will have already selected and started the event loop, in which case this won't work). I think maybe putting a similar fix into the example app itself might be better. The patch in #456 is in the Application object, which is where setting up the eventloop should normally occur.

ericpre:
Thanks for the reply, I reverted and updated the example accordingly.

Carreau:
CI failures seem to be due to a faulty version of Pytest; a patch was issued and released; so I restarted the test; which should fix everything.

Carreau:
Test are passing, and this is only in examples; so merging. Thanks.

PR #490, opened 2020-03-02
Make pdb on Windows interruptible

itamarst:
This is the final fix for https://github.com/ipython/ipython/issues/10516

Without the fix:

1. The test fails.
2. If you open Jupyter on Windows, run `pdb.set_trace()`, and then _in same cell_ run some other code while pdb is still waiting for input, the interrupt button won't work.

With the fix:

1. The test passes.
2. `pdb.set_trace()` is finally interruptible on Windows, even in the scenario described above.


itamarst:
@blink1073 one final PR and pdb is interruptible on Windows!

PR #491, opened 2020-03-14
Produce better traceback when kernel is interrupted

Tronic:
The original code no longer worked as intended because modern Python versions require `from None` to suppress the previous context, and as a result a long traceback of ipykernel internals was being printed when pressing the Stop button.

This PR fixes the suppression and adds a debug message to make it easier to it easier to understand what happened.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -575,7 +575,7 @@ class Kernel(SingletonConfigurable):
         content = parent['content']
         code = content['code']
         cursor_pos = content['cursor_pos']
-        
+
         matches = yield gen.maybe_future(self.do_complete(code, cursor_pos))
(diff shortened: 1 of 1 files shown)


ccordoba12:
Could you post a screenshot of the before and after behaviors?

Tronic:
Before:


After:


For comparison, [an extension that I'm working on](https://github.com/Tronic/niceback):

ccordoba12:
Wow! This is a great improvement, thanks!

Tronic:
Cool. I see that this affects Spyder as well. IPython in bare terminal doesn't run that code but produces a neat traceback (no internals, just user code) by default.

ccordoba12:
Yeah, it'll be a very neat improvement in both the notebook and Spyder.

PR #498, opened 2020-04-01
Allow interrupting input() on Windows, as part of effort to make pdb interruptible

itamarst:
This is part of the solution to https://github.com/ipython/ipython/issues/10516.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -893,7 +893,9 @@ class Kernel(SingletonConfigurable):
                 # re-raise KeyboardInterrupt, to truncate traceback
                 raise KeyboardInterrupt("Interrupted by user") from None
             else:
-                # Use polling with sleep so KeyboardInterrupts can get through:
+                # Use polling with sleep so KeyboardInterrupts can get through;
+                # doing a blocking recv() means stdin reads are uninterruptible
+                # on Windows:
                 if (ident, reply) == (None, None):
                     time.sleep(0.001)
                 else:
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -886,20 +886,25 @@ class Kernel(SingletonConfigurable):
         # Await a response.
         while True:
             try:
-                ident, reply = self.session.recv(self.stdin_socket)
-            except Exception as e:
-                self.log.warning("Invalid Message:", exc_info=True)
+                # Use polling with select() so KeyboardInterrupts can get
+                # through; doing a blocking recv() means stdin reads are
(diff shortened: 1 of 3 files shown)


itamarst:
Not sure what's up with failure on Travis, unless I did something wrong I also get them on `master` too.

ipykernel/kernelbase.py
```diff
@@ -885,14 +886,20 @@ def _input_request(self, prompt, ident, parent, password=False):
...
             except KeyboardInterrupt:
                 # re-raise KeyboardInterrupt, to truncate traceback
                 raise KeyboardInterrupt("Interrupted by user") from None
             else:
-                break
+                # Use polling with sleep so KeyboardInterrupts can get through;
+                # doing a blocking recv() means stdin reads are uninterruptible
+                # on Windows:
+                if (ident, reply) == (None, None):
+                    time.sleep(0.001)
```
blink1073, line 900: I'm wary of spin loops like this; since it is a socket we can use [`select()`](https://docs.python.org/3/library/select.html#select.select).

itamarst: Unfortunately `select()` is also uninterruptible by KeyboardInterrupt on Windows. I decided to try using it anyway, because that way at least the loop will be more responsive while needing to wake up less. But sadly as far as I can tell this needs a loop of some sort if it's going to be uninterruptible 😞

blink1073: Shucks, thanks for looking into it.  I'm not sure what happened on the last travis build, I kicked it.

itamarst: I tried matplotlib 3.2.1 on my Linux machine and can't reproduce those issues. But on newer version of Python.

blink1073: IPython 7.14.0 was released 4 days ago, that might be it.

blink1073: This might be relevant: https://github.com/ipython/ipython/pull/12184/files

itamarst: I'm using same version of matplotlib as Travis, though. With updated ipython still can't reproduce... Can you try rerunning tests on master? At least that way can tell if it's in any way related to this PR.

blink1073: I just kicked https://travis-ci.org/github/ipython/ipykernel/builds/682166613

blink1073: ...and that failed in the same way

blink1073: one thing to try is pinning the ipython version to 7.13.0

PR #521, opened 2020-07-03
fix #520: run post_execute and post_run_cell on async cells

### LINKED ISSUE #520: ZMQInteractiveShell does not run post_execute and post_run_cell events on async cells
In the notebook repository, I submitted an issue indicating that some cells did not display matplotlib images and delayed their display to the next cell (https://github.com/jupyter/notebook/issues/5568). After some inspection, I noticed that it only occurs in Python >= 3.8 and that the code elements that cause the issue are comprehensions, which are being detected by IPython as async cells. So, I submitted another issue to the IPython repository related to the misdetection (https://github.com/ipython/ipython/issues/12422).

However, async cells should not prevent matplotlib from displaying in the right cells after the end of their executions. I inspected it further and realized that the problem with async cells is that they do not run `post_execute` and `post_run_cell` events.

Here is a small example with the bug that does not use comprehensions nor matplotib (so it is easier to create a test and if someone fixes the async detection in the IPython repository, it will not impact this bug report):

**In [1]:**
```python
COUNTER = 0
def count():
    global COUNTER
    COUNTER += 1
    print("Count", COUNTER)
ip = get_ipython()
ip.events.register('post_execute', count)
print('A')
```
A
Count 1

**In [2]:**
```python
import asyncio
await asyncio.sleep(0)
print('B')
```
B

**In [3]:**
```python
print('C')
```
C
Count 2

---------------------------------------------
I'm pretty sure this is the main cause of the bug that I reported in the notebook repository (https://github.com/jupyter/notebook/issues/5568). Should I close the issue there now or wait for a fix of the bug here?

This issue is also related to #514

JoaoFelipe:
More details on the issue: https://github.com/ipython/ipykernel/issues/520

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ after_success:
   - codecov
 matrix:
   include:
-    - python: 3.6
+    - python: 3.7
       env:
         - IPYTHON=master
   allow_failures:
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -292,7 +292,13 @@ class IPythonKernel(KernelBase):
                 coro_future = asyncio.ensure_future(coro)
 
                 with self._cancel_on_sigint(coro_future):
-                    res = yield coro_future
+                    res = None
+                    try:
+                        res = yield coro_future
+                    finally:
+                        shell.events.trigger('post_execute')
+                        if not silent:
+                            shell.events.trigger('post_run_cell', res)
             else:
                 # runner isn't already running,
                 # make synchronous call,


JoaoFelipe:
I guess it failed in the CI due to NEP 29 (stop support for 3.6)

blink1073:
Mind changing this line to 3.7? https://github.com/ipython/ipykernel/blob/8ec2c4689614342483b7b3d85fe2a8536eef1527/.travis.yml#L60

blink1073:
Released as part of 5.3.1

Issue #518, opened 2020-06-19
tkinter gui event loop integration in ipykernel


yesinnk:
python version: 3.7.7
ipython version: 7.15.0
ipykernel version: 5.3.0

I've been trying to integrate the tkinter event loop in the jupyter notebook following: 
[docs](https://ipython.readthedocs.io/en/stable/config/eventloops.html
)
I am running the following commands in order:

```
%gui tk
import example
canvas = example.create_canvas()
```

where the implementation of example.py: 

import tkinter as tk

```
def create_canvas():
    root = tk.Tk()
    root.resizable(False, False)
    root.title("Karel")
    frame = tk.Frame(root)
    frame.pack()
    canvas = tk.Canvas(frame, bg="white", width=500, 
        height=500)
    canvas.pack()
    return canvas
```

but when I execute `%gui tk` I get an error: 

`AttributeError: '_tkinter.tkapp' object has no attribute 'createfilehandler'

but when I downgrade the ipykernel version to 4.10 (everything works fine).

from what I can assume, it's a bug in the latest version of ipykernel and I was wondering if there was any work around or a fix to this issue.

Similar to: [issue](https://github.com/ipython/ipykernel/issues/358)

Issue #358, opened 2018-10-05
Tk event loop not working on Windows


mhochsteger:
Hi,

There seems to be a problem with the event loop magic for Tk on Windows.
Executing `%gui tk` in the kernel results in:
```
ERROR:tornado.application:Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x000000000
Traceback (most recent call last):
  File "c:\program files\python37\lib\site-packages\tornado\ioloop.py", line 758, in _run_callback
    ret = callback()
  File "c:\program files\python37\lib\site-packages\tornado\stack_context.py", line 300, in null_wrapper
    return fn(*args, **kwargs)
  File "c:\program files\python37\lib\site-packages\ipykernel\kernelbase.py", line 295, in advance_eventloop
    eventloop(self)
  File "c:\program files\python37\lib\site-packages\ipykernel\eventloops.py", line 246, in loop_tk
    app.tk.createfilehandler(stream.getsockopt(zmq.FD), READABLE, notifier)
AttributeError: '_tkinter.tkapp' object has no attribute 'createfilehandler'
```
Tested with
- Windows 7 x64
- Python 3.7.0 x64
- ipykernel 5.0.0

ipykernel 4.10 works fine. Also commenting out line 246 in eventloops.py seems to work.

Best,
Matthias

mhochsteger:
@minrk I guess commit e719892c6f56ac5185a51584f5ddd136fb77b603 is relevant here.

dsblank:
"This feature is not available on Windows": https://docs.python.org/2/library/tkinter.html#file-handlers

yesinnk:
Is there a replacement feature for windows that allows tkinter event loop integration into the ipython kernel?

PR #523, opened 2020-07-08
Restore timer based event loop as a Windows-compatible fallback.

JosiahKane:
tk support has been broken on Windows since the timer based event loop hook was replaced with a better but POSIX only event driven one. This PR leaves the POSIX one in place where that is available, but falls back to the timer where it is not.

Related:
#518
#358

Semi-reverts commit [e719892c](https://github.com/ipython/ipykernel/commit/e719892c6f56ac5185a51584f5ddd136fb77b603)


blink1073:
Released as 5.3.2.

PR #525, opened 2020-07-15
Fix QSocketNotifier in the Qt event loop not being disabled for the control channel

ccordoba12:
This was generating a flush of warnings in qtconsole and Spyder.

Fixes jupyter/qtconsole#332
Fixes spyder-ide/spyder#13275

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -36,6 +36,12 @@ def _notify_stream_qt(kernel, stream):
         if stream.flush(limit=1):
             notifier.setEnabled(False)
             kernel.app.quit()
+        else:
+            # Even if there's nothing to flush, we need to disable the
+            # notifier in order to connect a new one in the next
+            # execution. This applies to the control channel.
+            notifier.setEnabled(False)
+            kernel.app.quit()
 
     fd = stream.getsockopt(zmq.FD)
     notifier = QtCore.QSocketNotifier(fd, QtCore.QSocketNotifier.Read, kernel.app)


blink1073:
Hi @ccordoba12!  Is this good to go?

ccordoba12:
Yes, it is. I was waiting for confirmation from others, and @dalthviz (one of our core devs) and some of our users reported it solved the problem for them.

I removed the reference to Vispy because I didn't hear from @hmaarrfk nor @djhoese, but I hope this will solve it for them too.

ipykernel/eventloops.py
```diff
@@ -36,6 +36,12 @@ def process_stream_events():
         if stream.flush(limit=1):
             notifier.setEnabled(False)
             kernel.app.quit()
+        else:
+            # Even if there's nothing to flush, we need to disable the
+            # notifier in order to connect a new one in the next
+            # execution. This applies to the control channel.
+            notifier.setEnabled(False)
+            kernel.app.quit()
```
djhoese, line 44: Doesn't putting this in an `else` make the entire `if/else` block unnecessary (both True/False cases run the exact same code)?

ccordoba12: I though about that too, but the `if` also flushes the shell channel. So perhaps I should rewrite this to

```python
stream.flush(limit=1)
notifier.setEnabled(False)
kernel.app.quit()
```

?

ccordoba12:
> Is this meant to stop the repeated issuing of the QSocketNotifier warning or does it stop it from appearing completely?

It stops it completely because now we disable the previous notifier, so that a new one can be activated for the same Zmq socket.

The problem is that, according to [Qt docs](https://doc.qt.io/qt-5/qsocketnotifier.html#details):

> [I]t is not possible to install two socket notifiers of the same type (Read, Write, Exception) on the same socket.

So, since flushing the control channel didn't return anything (because it's an extra channel that for now it's only used by the JupyterLab debugger, I think), we were not deactivating the previous notifier, which triggered the warning.

blink1073:
Released as 5.3.3.

ccordoba12:
Thanks for your help @blink1073! Do you want me to push another PR to improve the code according to @djhoese's suggestions?

I had them prepared but I was busy with other things, sorry.

ccordoba12:
Pinging @anaconda-pkg-build about this one. It'd be really great if you could include this version in the upcoming Anaconda release. Thanks!

jacky4eyes:
Hi, 

I'm not sure whether this is the right place for reporting issues, but my qt in jupyter doesn't work any more after I upgraded ipykernel today from 5.3.0 to 5.3.3. Downgrading it back to 5.3.0 solves my problem, so I believe there's probably some incompatibility in this 5.3.3 version.

Here is what I got:

Test this code in jupyter notebook:
```
%matplotlib qt
x = [1,2,3]
y = [5,4,5]
plt.plot(x,y)
```


Before upgrade to ipykernel==5.3.3
image

After:
image

pchristian4481:
Installed ipykernel==5.3.3 and reinstalled spyder 4.1.4 and that seemed to fix the socket issue, so far.

Thanks

blink1073:
Hi @jacky4eyes, can you please open a new issue?  I'm also curious if installing `5.3.2` fixes the problem, which would help narrow it down.

Wuhuuu2:
Hi all,
the last post is quite some time ago, but I have to say that I still get this error message.
All i need to create the error:
"QSocketNotifier: Multiple socket notifiers for same socket <changing number> and type Read"
```python
fig = plt.figure()
plt.show()
```
However there is an interesting thing about it:
If I copy this code into a new console, the message does not show up.
If I run this code from inside a file, like test.py it does show up (I excecute it in a new console).
Now, if I execute my test.py file, but either comment everything or prevent execution with import sys and sys.exit(), just a new console opens. Now running the uncommented code from test.py inside this previously new created console, the error does not show up. 
I feel thats interestingly strange. Maybe it helps you find the source again.

# Name                       Version        Build  Channel
ipython                        7.22.0           py39hd4e2768_0
ipython_genutils            0.2.0            pyhd3eb1b0_1
python                          3.9.4            h6244533_0
python-dateutil             2.8.1            pyhd3eb1b0_0
python-jsonrpc-server   0.4.0            py_0
python-language-server    0.36.2      pyhd3eb1b0_0
python-slugify               5.0.0            pyhd3eb1b0_0
python_abi                     3.9              1_cp39    conda-forge
ipykernel                        5.3.4            py39h7b7c402_0
matplotlib                      3.3.4            py39haa95532_0
matplotlib-base              3.3.4            py39h49ac443_0

hmaarrfk:
I have alot of questions about your setup.

It seems you are maybe mixing anaconda base with conda-forge stuff and are likely using old versions of certain packages. 

I'm not too sure, but it would help if we could get more organized details about your setup.

PR #526, opened 2020-07-17
Update logo to reflect IPython.

Carreau:
Especially in a context where Python can be used with Xeus, it is
important to have a distinction. This updates the logos shown in (for
example), JupyterLab to be clearer, but still convey "Python".


Carreau:
cc @SylvainCorlay

Issue #528, opened 2020-07-21
Qt slow in Spyder when using ipykernel 5.3.3 (5.3.2 OK)


8bit-Dude:
I am using Spyder 4.1.4 on Miniconda3-py37-4.8.3.
I noticed that matplotlib was unsually slow in IPython when using the Qt backend.
Downgrading ipykernel with conda to version 5.3.0 can however fix the issue.

blink1073:
@ccordoba12 might this be related to https://github.com/ipython/ipykernel/pull/525?

DiegoF90:
I also have the same issue, matplotlib seems to be very slow with the new Spyder 4.1.4.
I am not sure if it is related to #525, but I had already done the change suggested by @ccordoba12 [here](https://github.com/ipython/ipykernel/commit/4a05435220c730605ab454001ac3ecc364299736#diff-773c85e45dc4aacc95746531e6f97afe) and faced this issue later.

DiegoF90:
I do not know if this is related or f any use but, I was trying to use Tkinter as the graphics backend to see if the speed improved, but after restarting Spyder to apply this changes I got:

```python-traceback

ERROR:tornado.application:Exception in callback functools.partial(<function Kernel.enter_eventloop.<locals>.advance_eventloop at 0x000001D8E9EA4AF0>)
Traceback (most recent call last):
  File "C:\Users\1520797\AppData\Local\conda\conda\envs\gmv\lib\site-packages\tornado\ioloop.py", line 743, in _run_callback
    ret = callback()
  File "C:\Users\1520797\AppData\Local\conda\conda\envs\gmv\lib\site-packages\ipykernel\kernelbase.py", line 314, in advance_eventloop
    eventloop(self)
  File "C:\Users\1520797\AppData\Local\conda\conda\envs\gmv\lib\site-packages\ipykernel\eventloops.py", line 238, in loop_tk
    app.tk.createfilehandler(stream.getsockopt(zmq.FD), READABLE, notifier)
AttributeError: '_tkinter.tkapp' object has no attribute 'createfilehandler'

```

Which is the file referenced from which the problem seems to come from in #525 ?

DiegoF90:
[...]

PR #530, opened 2020-07-22
Revert "Fix QSocketNotifier in the Qt event loop not being disabled for the control channel

blink1073:
Reverts #525 in light of #528


blink1073:
cc @ccordoba12 I think we should make a release with this since #525 was a breaking change, then we can readdress the warnings.

ccordoba12:
Please don't simply revert my previous PR. The warnings are equally annoying. I'll submit a new PR with my proposed solution in #528 in a couple of hours.

PR #531, opened 2020-07-22
Only run Qt eventloop in the shell stream

### LINKED ISSUE #528: Qt slow in Spyder when using ipykernel 5.3.3 (5.3.2 OK)

ccordoba12:
Fixes #528.

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -36,12 +36,6 @@ def _notify_stream_qt(kernel, stream):
         if stream.flush(limit=1):
             notifier.setEnabled(False)
             kernel.app.quit()
-        else:
-            # Even if there's nothing to flush, we need to disable the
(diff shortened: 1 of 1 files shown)


ccordoba12:
@blink1073, this is my proposed solution for #528. Sorry the delay.

I tested it on my end and found no issues. However, I'll post a message on #528 to see if other users can test it too.

blink1073:
Sounds like we're good, releasing now.

blink1073:
https://pypi.org/project/ipykernel/5.3.4/#files

ccordoba12:
Thanks @blink1073 for releasing a new version with this patch.

PR #535, opened 2020-07-30
Check and adjust permissions before kernel spec is written

RobertRosca:
If the permissions for the `resources` directory do not allow the owner to write into it (e.g. as is the case in some directories using ACL, or if `site-packages` is set to read-only) then the install phase will fail with an error saying that `kernel.json` could not be written as the copied `resources` directory is read-only.

This PR fixes that by adding a permission check to `write_kernel_spec`. If `resources` has no write permissions it sets only the owner permission (leaving group/other unchanged) to 7 so that `kernel.json` can be written into it.

Also added in a test which creates a `resources` directory with read-only permissions to check that this has worked correctly. For the test to work `write_kernel_spec` now has a new `resources` optional argument, which is set to `RESOURCES` by default.

@takluyver :smile:

### DIFF
--- a/ipykernel/kernelspec.py
+++ b/ipykernel/kernelspec.py
@@ -77,9 +77,6 @@ def write_kernel_spec(path=None, overrides=None, extra_arguments=None):
 
     # change permission if resources directory is not read-write able
     if not os.access(path, os.W_OK | os.R_OK):
-        warnings.warn(
-            UserWarning("resources is not writable, adjusting permissions before creating kernel")
-        )
         # changes permissions only for owner, do not touch group/other
         os.chmod(path, os.stat(path).st_mode | 0o700)
         for f in os.listdir(path):
--- a/ipykernel/kernelspec.py
+++ b/ipykernel/kernelspec.py
@@ -59,7 +59,7 @@ def get_kernel_dict(extra_arguments=None):
     }
 
 
-def write_kernel_spec(path=None, overrides=None, extra_arguments=None, resources=RESOURCES):
+def write_kernel_spec(path=None, overrides=None, extra_arguments=None):
     """Write a kernel spec directory to `path`
 
     If `path` is not specified, a temporary directory is created.
@@ -72,7 +72,7 @@ def write_kernel_spec(path=None, overrides=None, extra_arguments=None, resources
         path = os.path.join(tempfile.mkdtemp(suffix='_kernels'), KERNEL_NAME)
 
     # stage resources
-    shutil.copytree(resources, path)
+    shutil.copytree(RESOURCES, path)
 
     # change permission if resources directory is not read-write able
     if not os.access(path, os.W_OK | os.R_OK):
@@ -96,10 +96,10 @@ def write_kernel_spec(path=None, overrides=None, extra_arguments=None, resources
 def install(kernel_spec_manager=None, user=False, kernel_name=KERNEL_NAME, display_name=None,
             prefix=None, profile=None):
     """Install the IPython kernelspec for Jupyter
-    
+
     Parameters
     ----------
-    
+
     kernel_spec_manager: KernelSpecManager [optional]
         A KernelSpecManager to use for installation.
         If none provided, a default instance will be created.
@@ -118,7 +118,7 @@ def install(kernel_spec_manager=None, user=False, kernel_name=KERNEL_NAME, displ
 
     Returns
     -------
-    
+
     The path where the kernelspec was installed.
     """
     if kernel_spec_manager is None:
@@ -153,12 +153,12 @@ from traitlets.config import Application
 class InstallIPythonKernelSpecApp(Application):
     """Dummy app wrapping argparse"""
     name = 'ipython-kernel-install'
-    
+
     def initialize(self, argv=None):
         if argv is None:
             argv = sys.argv[1:]
         self.argv = argv
-    
+
     def start(self):
         import argparse
         parser = argparse.ArgumentParser(prog=self.name,
--- a/ipykernel/kernelspec.py
+++ b/ipykernel/kernelspec.py
@@ -59,19 +59,29 @@ def get_kernel_dict(extra_arguments=None):
     }
 
 
-def write_kernel_spec(path=None, overrides=None, extra_arguments=None):
+def write_kernel_spec(path=None, overrides=None, extra_arguments=None, resources=RESOURCES):
     """Write a kernel spec directory to `path`
-    
+
     If `path` is not specified, a temporary directory is created.
     If `overrides` is given, the kernelspec JSON is updated before writing.
-    
+    if 'resources' is given, copies resources from non-standard location.
+
     The path to the kernelspec is always returned.
     """
     if path is None:
         path = os.path.join(tempfile.mkdtemp(suffix='_kernels'), KERNEL_NAME)
-    
+
     # stage resources
-    shutil.copytree(RESOURCES, path)
+    shutil.copytree(resources, path)
+
+    # change permission if resources directory is not read-write able
+    if not os.access(path, os.W_OK | os.R_OK):
+        # changes permissions only for owner, do not touch group/other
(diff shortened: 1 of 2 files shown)


ipykernel/kernelspec.py
```diff
@@ -59,27 +59,37 @@ def get_kernel_dict(extra_arguments=None):
     }
 
 
-def write_kernel_spec(path=None, overrides=None, extra_arguments=None):
+def write_kernel_spec(path=None, overrides=None, extra_arguments=None, resources=RESOURCES):
```
takluyver, line 62: I'm ambivalent about this - on the one hand, I don't like adding function parameters just for tests, but on the other, I don't like using mocking unless it's necessary. So leave this as is unless someone else asks for a change.

takluyver: (Just realised 'as is' is ambiguous - I mean as you have already done it in this PR)

ipykernel/tests/test_kernelspec.py
```diff
@@ -88,6 +88,25 @@ def test_write_kernel_spec_path():
     shutil.rmtree(path)
 
 
+def test_write_kernel_spec_permissions():
+    read_only_resources = tempfile.mkdtemp()
+    shutil.copytree(RESOURCES, read_only_resources, dirs_exist_ok=True)
```
takluyver, line 93: The `dirs_exist_ok` parameter is new in Python 3.8, and ipykernel will need to support earlier versions for a while. The easy way to do that is just to copy to a subdirectory of your new temp directory.

ipykernel/kernelspec.py
RobertRosca: Woo now I know how to use the mock patch thing :D never used it before.

Ran into a small problem though. I want to test that the mock patch worked correctly, and that the path the patched `RESOURCES` variable points to has the correct permissions set.

But the `RESOURCES` variable comes from `from ipykernel.kernelspec import RESOURCES` at the top, and I can't think of an easy way to get the patched variable without doing `import ipykernel.kernelspec` so that I can call `ipykernel.kernelspec.RESOURCES` directly in the `with mock.patch(...)` block.

At the same time I also think that this permission-changing behaviour should be a bit more verbose and raise a warning. So I did the whole "two birds with one stone" thing and added a warning to `write_kernel_spec` whenever the permissions are modified, and then the test checks that this warning is raised.

To be honest, I'm like 50/50 split on if the warning is useful or pointless, so if you think the warning shouldn't be there I'm happy to remove it.

ipykernel/tests/test_kernelspec.py
RobertRosca: Implemented your suggestion

ipykernel/kernelspec.py
takluyver: I think it shouldn't warn - read-only files in site-packages doesn't indicate that anything is going wrong or needs to be fixed. And people don't like unnecessary warnings.

If you want to ensure that it's picking up your test copy of the resources dir, why not stick an extra file in there, and then check that it exists in the resulting directory?

RobertRosca: Aha, good idea, implemented it like you said with a marker file.

My concern that `RESOURCES` didn't get mocked properly is probably silly, my thinking is more that the way the source resources directory is specified for `write_kernel_spec` might change at some point and just silently break this test.

takluyver:
I think the test can be a bit neater using the pytest `tmp_path` feature, like this:

```python
def test_write_kernel_spec_permissions(tmp_path):
```

Then pytest takes care of creating a temporary directory, gives you a [pathlib](https://docs.python.org/3/library/pathlib.html) Path object to work with, and cleans up afterwards.

RobertRosca:
Oops, didn't see your comment, I changed it to use the fixture as you suggested.

I originally wrote it using the pathlib methods instead of `os.path.join` like this:

```python
    read_only_resources = tmp_path.joinpath("_RESOURCES")
    shutil.copytree(RESOURCES, read_only_resources)

    # file used to check that the correct (mocked) resource directory was used
    with open(read_only_resources.joinpath('touch'), 'w') as f:
        pass
```

But, at least to me, the `.joinpath` thing seems a bit harder to read than `os.path.join` for something simple like this, so I'd rather leave it as it is.

ipykernel/kernelspec.py
```diff
@@ -61,35 +61,45 @@ def get_kernel_dict(extra_arguments=None):
 
 def write_kernel_spec(path=None, overrides=None, extra_arguments=None):
     """Write a kernel spec directory to `path`
-    
+
     If `path` is not specified, a temporary directory is created.
     If `overrides` is given, the kernelspec JSON is updated before writing.
-    
+    if 'resources' is given, copies resources from non-standard location.
+
```
ivanov, line 68: this should be removed, since resources were added, but then removed as a parameter to write_kernel_spec in this PR.

ipykernel/kernelspec.py
```diff
@@ -61,35 +61,45 @@ def get_kernel_dict(extra_arguments=None):
...
     shutil.copytree(RESOURCES, path)
+
+    # change permission if resources directory is not read-write able
+    if not os.access(path, os.W_OK | os.R_OK):
+        # changes permissions only for owner, do not touch group/other
+        os.chmod(path, os.stat(path).st_mode | 0o700)
+        for f in os.listdir(path):
+            file_path = os.path.join(path, f)
+            os.chmod(file_path, os.stat(file_path).st_mode | 0o600)
+
```
ivanov, line 84: I believe the relevant piece of this code has already been merged in #593 (and unfortunately earlier in #377, which was overlooked). I think that we don't need the directory to have the execute bit set, only have it be writeable. We don't need all of the files to be writeable - since those are icons - they just need to be readable - which gets taken care of by copytree just above this section, since if they weren't readable from the RESOURCES directory, that would have failed.

I might have overlooked something subtle, though, but either way this would need to be rebased on master, as it conflicts with what ended up making it to ipykernel 5.5.0 in #593.

ipykernel/tests/test_kernelspec.py
```diff
@@ -88,10 +90,41 @@ def test_write_kernel_spec_path():
     shutil.rmtree(path)
 
 
+@pytest.mark.skipif(sys.platform == 'win32',
+                    reason="does not run on windows")
+def test_write_kernel_spec_permissions(tmp_path):
```
ivanov, line 95: This function should now probably be the main focus of this PR. I believe the implementation from #593 runs on Windows (at least it does for me on Python 3.9.2), so no need to skip that platform.

ivanov (changes requested):
Thank you for pitching in, @RobertRosca, and sorry that this stalled. As I mention on a code inline comment - I believe the functional equivalent of the proposed change has already been merged in #593 (and unfortunately also overlooked earlier in #377). 

If you have the desire and time, I think it would make sense to isolate this PR to a test of that functionality, so we don't accidentally lose it in the future. We also probably want remove all those whitespace changes from the commit to keep things tidy and isolated to the new test case

 Let us know if you'd like to proceed or if we should press on by ourselves. Either way is fine, it's really up to you :)

Happy hacking! :bowtie:

RobertRosca:
> Thank you for pitching in, @RobertRosca, and sorry that this stalled. As I mention on a code inline comment - I believe the functional equivalent of the proposed change has already been merged in #593 (and unfortunately also overlooked earlier in #377).

Hey, no problem! That's alright, I understand these things can drop through the cracks (like you reply did for me :wink:) so it's no problem.

> If you have the desire and time, I think it would make sense to isolate this PR to a test of that functionality, so we don't accidentally lose it in the future. We also probably want remove all those whitespace changes from the commit to keep things tidy and isolated to the new test case

Sure, that sounds for me.

> Let us know if you'd like to proceed or if we should press on by ourselves. Either way is fine, it's really up to you :)

I'm happy to carry on with this, but given it's been a while since you asked I want to check if this is all still needed.

Cheers!

Carreau:

PR #537, opened 2020-08-17
Remove most of the python 2 compat code

rahulporuri:
Note to reviewer : I have tried making one and only one type of change in each of the commits so reviewing this PR commit-by-commit will be straightforward.

This PR removes most of the machinery in the codebase that supported python 2/3 compatibility.

- remove `__future__` import statements
- remove coding/encoding cookies from files
- remove try/except statements around imports that have changed between python 2 and 3
- use str instead of py3compat.unicode_type
- use str instead of py3compat.string_types
- use dict.items() instead of py3compat.iteriterms(dict)
- Remove a python 2 only test
- Remove py3compat.PY3 conditionals
- Use metaclass kwarg instead of py3compat.with_metaclass
- Use builtins instead of py3compat.builtin_mod
- Use `func.__closure__` instead of py3compat.get_closure(func)
- Remove py3compat.*_py2 functions which are no-op on python 3

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
 #
 # IPython Kernel documentation build configuration file, created by
 # sphinx-quickstart on Mon Oct  5 11:32:44 2015.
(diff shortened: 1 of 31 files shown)


rahulporuri:
@blink1073 i think I rebased correctly and didn't mess anything up.

PR #541, opened 2020-08-23
Add env parameter to kernel installation

fesaille:
This delivers a basic support of **env** parameter for setting kernel environment variables.

[Optional environment variables](https://jupyter-client.readthedocs.io/en/stable/kernels.html) can be set in kernels, this modification sets multiple environments variables on installation.

If the proposal is accepted, env. var. correctness can be added e.g. based on [Open Group Base Specifications](https://pubs.opengroup.org/onlinepubs/9699919799/)


fesaille:
Added a test based on pytest, hope it is ok (nose seems to be unmaintained since years).

blink1073:
> Added a test based on pytest, hope it is ok (nose seems to be unmaintained since years).

Yep, we only have `nose` around for a few utility functions.  Thanks for adding a test!

PR #547, opened 2020-09-09
Fix stack levels for ipykernel's deprecation warnings and stop using some deprecated APIs

jack1142:
This PR:
- fixes stack levels for ipykernel's deprecation warnings
- stops using deprecated traitlets api for observing
- uses the ipyparallel interfaces over the deprecated ipykernel variants when available
- uses tornado's IOLoop when pyzmq version is 17 or greater

This greatly minimizes the amount of deprecation warnings from ipykernel.


minrk:
Great, thanks!

PR #561, opened 2020-11-21
appnope segfaults on Apple Silicon

geohot:
I haven't root caused why, but this PR disables it for arm64 chip. Otherwise on M1 the kernels segfault on startup.

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -17,9 +17,9 @@ from traitlets.config.application import Application
 def _use_appnope():
     """Should we use appnope for dealing with OS X app nap?
 
-    Checks if we are on OS X 10.9 or greater.
+    Checks if we are on OS X 10.9 or greater and not on Apple Silicon.
     """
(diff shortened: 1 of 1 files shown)


mwidjaja1:
Can confirm, this change fixes Jupyter for me on an ARM Macbook Air!

erykoff:
Note that this solution is just a band-aid on a much deeper wound, see #562 .  Something is not working with ctypes calls on Apple Silicon that is touched by appnope.

geohot:
As far as I can tell, it's just appnope itself that's broken, nothing deeper. Fix here:

https://github.com/minrk/appnope/pull/7

mwidjaja1:
I don't think this change is necessary anymore because as noted in the issue @geohot linked, the issue has been resolved in appnope. I updated appnope today, reverted this MR in my copy of iPython, and I was able to run Jupyterlab natively. 

https://github.com/minrk/appnope/pull/7#issuecomment-735236121

geohot:
Yes, should be good with appnope 0.1.1, which will already be installed. Closing PR.

PR #564, opened 2020-12-10
add github actions, bail on asyncio patch for tornado 6.1

### LINKED ISSUE #563: handle tornado 6.1 on windows vs python 3.8
The good news: tornado 6.1 obviates the need for the nasty event loop hack.

The bad news: if the event loop hack is applied, it causes intermittent hanging at some deep dark level, and needs to be killed with non-python fire.

I've sprinkled a hard removal of the hack in a bunch of `jupyter_server` stuff, but feel like we might need to be a bit more cautious here.

I'm thinking of just adding `if tornado.version_info >= (6, 1): continue` to all of the hacks, as that API has been stable back to our bottom pin.

Also, it's about time to get off the travis-ci bus. GitHub actions seems like the right play, we've been fairly successful with it across a number of other repos.

PR incoming, but it might take a while (i'll try to get it green on a personal PR first).

bollwyvl:
This:
- removes appveyor and travis
- adds github actions
  - tests (py3.{6,7,8,9} + pypy) x {win,lin,mac}
    - couldn't even really get 3.5 going, not sure what's up
  - codecov
  - wheel caching
- adds skips for some osx/pypy tests that just aren't working
  - we already skip them in conda-forge downstream
- adds a version check for the tornado asyncio patch(es) on windows

Links:
- fixes #563


blink1073 (approved):
Looks good, thank you!

Zsailer:
This might warrant a quick patch release here. Things are a bit flaky in jupyter_server's CI and I think it's related to this issue.

Zsailer:
BTW, thank you @bollwyvl!

bollwyvl:
Thanks so much for the quick review! ☺️

PR #567, opened 2020-12-11
Invalid syntax in ipykernel/log.py

chronitis:
The `DeprecationWarning` in this file is missing a comma, so the file can't be loaded. Not used in the normal configuration (since it's deprecated, but while the file is present it should probably be importable).

Found while packaging for Debian since all files get byte-compiled on install, which picked this up.


blink1073 (approved):
Thanks!

blink1073:
https://pypi.org/project/ipykernel/5.4.1/

Issue #571, opened 2020-12-11
Re-introduce #539 after understanding language_info KeyError


glentakahashi:
See https://github.com/ipython/ipykernel/pull/570 and https://github.com/ipython/ipykernel/pull/539

Not sure why language_info is returning null for an aborted request, and if it happens consistently or only if the kernel runs code that has an error right before a language_info request is fired off.

If its the latter case, I think the expected behavior should be that `kernel_info_request` requests should be exempt from the "abort_on_error" functionality.

PR #572, opened 2020-12-14
Re-introduce #539, fix abort to not abort non-execute requests

glentakahashi:
Fixes https://github.com/ipython/ipykernel/issues/571

As described in the original issue (https://github.com/ipython/ipykernel/pull/539), the old `stop_on_error_timeout ` functionality didn't actually work asynchronously, and would cause new requests to block until the error was finished. The PR #539 fixed this to work async without changing any of the other logic.

This caused occasional issues in other libraries that depend on the `kernel_info` request, when the notebook had cells with python errors in them due to the race condition of sending off the request.

This new PR re-introduces the async `stop_on_error_timeout` fix, but changes the logic to abort /only/ `execute_request`'s. This means that running multiple cells in a row where the first cell has an error will still cause the remaining cell executions to abort, but you can still continue to do things like `kernel_info`, `inspect`, `completions`, etc.

It also adds a couple of tests for those.

Also tested against papermill in https://github.com/nteract/papermill/issues/519 to verify that the fix allows papermill to still function.


glentakahashi:
Hmm, i'm pretty sure the failing test is a flake, but I don't know how to re-trigger a step in github actions. Do I just have to push up an empty no-op commit or do I lack permissions?

blink1073:
Thanks for adding the test and for testing against `papermill`!  I kicked CI.  Closing and re-opening the PR works I think, otherwise you can rebase and force push to force a rebuild.

blink1073 (approved):
Thanks!  A different CI build failed this time, so it looks like flakiness.

PR #573, opened 2020-12-16
Deacrease lag time for eventloop

impact27:
While using interactive matplotlib figure, the figure will freeze for 1 seconds very often. This get rid of this freeze.


blink1073 (approved):
Thanks!  I verified that the CPU usage is not affected.

PR #578, opened 2021-01-11
Rework wait_for_ready logic

davidbrochart:
This implements the same logic [as in jupyter_client](https://github.com/jupyter/jupyter_client/commit/06a32c4f91aba9a8ee6ee76474215aa96e5cd1c1).
Closes https://github.com/jupyter/jupyter_client/issues/601

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,6 +97,7 @@ setup_args = dict(
             'pytest-cov',
             'flaky',
             'nose',  # nose because there are still a few nose.tools imports hanging around
+            'jedi<=0.17.2'
         ],
     },
     classifiers=[
--- a/ipykernel/inprocess/blocking.py
+++ b/ipykernel/inprocess/blocking.py
@@ -76,10 +76,21 @@ class BlockingInProcessKernelClient(InProcessKernelClient):
     def wait_for_ready(self):
         # Wait for kernel info reply on shell channel
         while True:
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -97,6 +97,7 @@ def run(self):
             'pytest-cov',
             'flaky',
             'nose',  # nose because there are still a few nose.tools imports hanging around
+            'jedi<=0.17.2'
```
ccordoba12, line 100: I think this needs to be fixed in IPython, not here. Pinging @bollwyvl and @Carreau about it.

davidbrochart: Eventually yes, but the tests were failing without that.

Carreau: I'm aware of this, but I'm swamped and haven't had a chance to look at IPython since mid december, I know there are a bunch of issues and PRs I need o review/fix/merge.

blink1073 (approved):
Thanks!

setup.py
jasongrout: @davidbrochart - can we add a comment to the effect here, so that we know later when we can remove the pin?

davidbrochart: Done in #579

blink1073:
@meeseeksdev please backport to 5.4.x

blink1073:
https://pypi.org/project/ipykernel/5.4.3/

PR #585, opened 2021-01-27
Run control channel in separate thread

### LINKED ISSUE #447: Handle the Control channel in a separate thread
Hello,

I’d created a [similar discussion in Jupyter Google Groups](https://groups.google.com/forum/#!topic/jupyter/2jriNY6H5pg) but looks that responsible people from IPython kernel does not participate in groups.

Brief explanation: I tried to implement a tunnel for the [pydev debugger](https://www.pydev.org/manual_adv_remote_debugger.html) through notebook’s websocket and using Comm. Unfortunately, Comm and cell execution shares the same thread, so when debugger suspends the cell’s code it also suspends the tunnel. Anyway, I'd succeeded to implement it by moving a Control channel in a separate thread using monkey-patching but it doesn't look like a production-ready solution.

I propose to move handling of the Control channel in a separate thread and add an analog of Comm for that thread. It could be used for tunneling streams of a debugger or for collecting metrics at runtime.

A rough algorithm of handling each message in a separate thread:
* If the message can be handled in the Control thread then handle it immediately. Example of methods that can be handled here: `kernel_info_request`, `complete_request`, Comm for Control.
* Otherwise the message is being re-queued into main thread’s event loop and then it being handled like in a current implementation.

Do you see any obvious pitfalls around handling the Control channel in a separate thread? If I’ll prepare a patch with acceptable code quality, will you merge it?

SylvainCorlay:
This makes the control channel run in a separate thread from the shell channel, so that control messages can be processed concurrently to shell requests.

 - this is now the recomended approach in the [Jupyter protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html#introduction).
 
    >  *For a smoother user experience, we recommend running the control channel in a separate thread from the shell channel, so that e.g. shutdown or debug messages can be processed immediately without waiting for a long-running shell message to be finished processing (such 
as an expensive execute request).*

 - this is required for ipykernel to support the Jupyter Debug Protocol in ipykernel: we need to be able to remove and add breakpoints while code is running (or even more importantly, process messages such as "continue", "step into", which are by nature to be called during execution.

Fixes #447

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -36,6 +36,7 @@ from jupyter_client.connect import ConnectionFileMixin
 
 # local imports
 from .iostream import IOPubThread
+from .control import ControlThread
 from .heartbeat import Heartbeat
 from .ipkernel import IPythonKernel
 from .parentpoller import ParentPollerUnix, ParentPollerWindows
@@ -124,6 +125,7 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
     stdin_socket = Any()
     iopub_socket = Any()
     iopub_thread = Any()
+    control_thread = Any()
 
     ports = Dict()
 
@@ -276,6 +278,17 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
         self.stdin_port = self._bind_socket(self.stdin_socket, self.stdin_port)
         self.log.debug("stdin ROUTER Channel on port: %i" % self.stdin_port)
 
+        if hasattr(zmq, 'ROUTER_HANDOVER'):
+            # set router-handover to workaround zeromq reconnect problems
+            # in certain rare circumstances
+            # see ipython/ipykernel#270 and zeromq/libzmq#2892
+            self.shell_socket.router_handover = \
+                self.stdin_socket.router_handover = 1
+
+        self.init_control(context)
+        self.init_iopub(context)
+
+    def init_control(self, context):
         self.control_socket = context.socket(zmq.ROUTER)
         self.control_socket.linger = 1000
         self.control_port = self._bind_socket(self.control_socket, self.control_port)
@@ -285,11 +298,10 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp,
             # set router-handover to workaround zeromq reconnect problems
             # in certain rare circumstances
             # see ipython/ipykernel#270 and zeromq/libzmq#2892
-            self.shell_socket.router_handover = \
-                self.control_socket.router_handover = \
-                self.stdin_socket.router_handover = 1
+            self.control_socket.router_handover = 1
 
-        self.init_iopub(context)
+        self.control_thread = ControlThread(self.control_socket)
+        self.control_thread.start()
 
     def init_iopub(self, context):
         self.iopub_socket = context.socket(zmq.PUB)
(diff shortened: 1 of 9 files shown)


minrk:
Nice! How do you handle maintaining the `parent_header` state? Processing a control request needs to set parent header for its own side effects (only status?), but must not conflict with the parent header set by the now concurrently outstanding shell request.

Make sure to add a test for this, like:

1. shell request produces multiple outputs (e.g. `for i in range(5): print(i); sys.stdout.flush(); time.sleep(1)`)
2. control request begins and ends in the middle of shell request, ideally ensuring it overlaps at least one or two shell output messages
3. verify that all shell output is properly routed and not captured by control, status messages are handled correctly, etc.

SylvainCorlay:
> Nice! How do you handle maintaining the parent_header state? 

Yes, I think we need to have two parent_header attributes depending on the channel like in xeus-python.

> now concurrently outstanding shell request.

The debugger tests will essentially be of that nature. Since we are working on plugging the debugger (porting code from xeus-python) on top of this PR, we should probably keep this open until the debugger works in practice.

ipykernel/kernelapp.py
```diff
@@ -437,13 +449,13 @@ def init_signal(self):
...
         """Create the Kernel object itself"""
         shell_stream = ZMQStream(self.shell_socket)
-        control_stream = ZMQStream(self.control_socket)
+        control_stream = ZMQStream(self.control_socket, self.control_thread.io_loop)
 
         kernel_factory = self.kernel_class.instance
 
         kernel = kernel_factory(parent=self, session=self.session,
                                 control_stream=control_stream,
-                                shell_streams=[shell_stream, control_stream],
```
minrk, line 446: Can we leave shell_streams as a list? ipyparallel relies on this API and even if the default length is 1, there's no harm in preserving the type for compatibility. It should also make this patch smaller, since there's no type change to work around.

SylvainCorlay:
Update: test cases are passing, but somehow, the suite hangs forever in the end.
Something must still be wrong with the shutdown.

ipykernel/kernelapp.py
SylvainCorlay: Okay, it makes sense if this is an API that people depend upon.

(I thought it was simplifying the code to drop the list).

SylvainCorlay: I added a deprecated `shell_streams` list property. Let me know if you think that it is fine.

ipykernel/kernelbase.py
```diff
@@ -681,9 +672,17 @@ def shutdown_request(self, stream, ident, parent):
...
+        # Flush to ensure reply is sent before stopping loop
+        self.control_stream.flush(zmq.POLLOUT)
+
+        self.log.debug('Stopping control ioloop')
+        control_io_loop = self.control_stream.io_loop
+        control_io_loop.add_callback(control_io_loop.stop)
+
+        self.log.debug('Stopping shell ioloop')
+        shell_io_loop = self.shell_stream.io_loop
+        shell_io_loop.add_callback(shell_io_loop.stop)
```
SylvainCorlay, line 685: I have been instrumenting the code to investigate the random failure, and it seems that `shell_io_loop.add_callback(shell_io_loop.stop)` does not always stop the shell ioloop for some reason.

Note: this function is called from the control thread, but `add_callback` is supposedly thread safe.

ipykernel/kernelapp.py
```diff
@@ -330,7 +342,7 @@ def close(self):
             if socket and not socket.closed:
                 socket.close()
         self.log.debug("Terminating zmq context")
-        self.context.term()
+        self.context.destroy()
```
SylvainCorlay, line 345: I will look at what sockets `destroy` closes which may cause `term` to hang.

SylvainCorlay:
No random failure anymore. We have systematic failure on `Windows + Python 3.6` and `Windows + Python 3.7`.

blink1073 (approved):
Nice work, thank you!

Carreau:
I think this has broken Jupyter_client ci; there are a bunch of errors wrt control channel.

ipykernel/kernelbase.py
```diff
@@ -60,7 +58,16 @@ def _update_eventloop(self, change):
...
-    shell_streams = List()
+    shell_stream = Instance(ZMQStream, allow_none=True)
+
+    @property
+    def shell_streams(self):
+        warnings.warn(
+            'Property shell_streams is deprecated in favor of shell_stream',
+            DeprecationWarning
+        )
+        return [shell_stream]
```
Carreau, line 69: missing self.

PR #590, opened 2021-02-09
Add configure_inline_support and call it in the shell

martinRenou:
This removes a circular dependency between ipykernel and IPython. See https://github.com/ipython/ipython/pull/12814

### DIFF
--- a/ipykernel/pylab/backend_inline.py
+++ b/ipykernel/pylab/backend_inline.py
@@ -13,6 +13,7 @@ from matplotlib import colors
 from matplotlib._pylab_helpers import Gcf
 
 from IPython.core.getipython import get_ipython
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
LGTM, thanks!

martinRenou:
Thanks for merging @blink1073 We need to check that this works properly with the current IPython version. I've only tested it coupled with ipython/ipython#12814. Sorry, this should have been a draft PR :S

blink1073:
Thanks for the heads up!  I'll make sure not to cut any releases until then.

martinRenou:
Looks like it works properly 👍🏼 

It will configure the inline backend twice, but I guess we can live with that until IPython removes it.

blink1073:
Great, thanks for following up!

Issue #588, opened 2021-02-01
Extract inline backend from ipykernel and matplotlib magics from IPython


martinRenou:
There is currently a big special case for the Matplotlib library in ipykernel and IPython.

IPython provides a `%matplotlib`magic that allows choosing the current Matplotlib backend that is being used. For the sake of modularity, and in order to get rid of the special case, we should maybe extract this magic out of IPython and put it in Matplotlib (when Matplotlib is installed, it installs this magic).

ipykernel provides the Matplotlib inline backend, this should maybe be extracted from ipykernel and put in a separate package (Matplotlib?). This new module should maybe contain the [configure_inline_support](https://github.com/ipython/ipython/blob/master/IPython/core/pylabtools.py#L363) method that currently lives in IPython, and should be called when `%matplotlib inline` is invoked.

Those two improvements would help clean up the current ipykernel-IPython-Matplotlib relationship.

@SylvainCorlay @Carreau @tacaswell

Carreau:
@martinrenou do you consider this finished now that IPython rely on `matplotlib-inline` ?

martinRenou:
Yes! Let's close it

SylvainCorlay:
However, we should eventually get back to core IPython and remove the hardcoded dict "backend name" -> "module".

PR #591, opened 2021-02-12
Use matplotlib-inline

martinRenou:
This goes in the direction of #588

Opening as a draft for now, until there is some consensus on under which organization we put this matplotlib-inline (and how we should name it). 

It is temporarily here: https://github.com/martinRenou/matplotlib-inline

I believe this should go under http://github.com/matplotlib or http://github.com/ipython


ipykernel/pylab/backend_inline.py
```diff
@@ -3,246 +3,4 @@
...
-    opposed to dark). Based on ITU BT.601 luminance formula (see
-    https://stackoverflow.com/a/596241)."""
-    rgbaArr = colors.to_rgba_array(color)
-    return rgbaArr[:,:3].dot((.299, .587, .114)) > .5
-
-def _is_transparent(color):
-    """Determine transparency from alpha."""
-    rgba = colors.to_rgba(color)
-    return rgba[3] < .5
+from matplotlib_inline.backend_inline import * # analysis: ignore
```
martinRenou, line 6: We might want to add a deprecation warning here.

martinRenou:
This PR is coupled with https://github.com/ipython/ipython/pull/12817, but it doesn't need it for having a working inline backend.

SylvainCorlay:
@martinRenou I rebased the PR now that IPython 7.23 is out with the new inline backend.

martinRenou:
Thanks! I'll see if this PR needs updating

blink1073 (approved):
Thanks!

PR #593, opened 2021-02-18
kernelspec: ensure path is writable before writing kernel.json.

### LINKED ISSUE #592: Cannot python3 -m ipykernel install when ipykernel is installed in read-only directory
When running
`python3 -m ipykernel install --user --name=<my-venv>`, I see:

> [Errno 13] Permission denied: '/tmp/tmpok8jp1dc_kernels/python3/kernel.json'

I pinpointed the problem to the following line:

https://github.com/ipython/ipykernel/blob/da44c7c7b8d60016e68408fefb438b511d2aa283/ipykernel/kernelspec.py#L78

I have the issue that the directory pointed to by `RESOURCES` is read-only (by design), which leads to `path` being read-only as well, after being populated by `shutil.copytree`. This subsequently leads to some permission denied errors when trying to write kernel.json. 

This is easily remedied by something like the following:
```python
import stat
mode_flags = stat.S_IWUSR
    m = os.stat(path).st_mode
    if not m & mode_flags:
        os.chmod(path, m | mode_flags)
```

If this is an acceptable (upstream-able) fix for my admittedly weird situation, I can work on a proper PR. 
Thanks!

jellelicht:
Fixes #592

### DIFF
--- a/ipykernel/kernelspec.py
+++ b/ipykernel/kernelspec.py
@@ -7,6 +7,7 @@ import errno
 import json
 import os
 import shutil
+import stat
 import sys
 import tempfile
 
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!  I'll make a release with this change tomorrow.

ivanov:
Thanks for this, @jellelicht and @blink1073 -  I just wanted to flag that we had a stale PR with an equivalent fix sitting around for about two years in #377 - so hopefully we can do better job catching such duplicates in the future.

PR #595, opened 2021-02-22
Change to markdown for changelog

afshin:
This makes it easier to maintain. Converted using pandoc and some manual updates.


blink1073 (approved):
Thanks!

PR #599, opened 2021-02-27
Do not override matplotlib params

Mottl:
`InlineBackendConfig` has some hardcoded parameters which prevents loading of user-defined parameters from `~/.matplotlib/matplotlibrc` or `~/.config/matplotlib/matplotlibrc`. Matplotlib has already good (or the same) default parameters which `InlineBackendConfig` redefines.

This commit removes unnecessary overrides of `matplotlib.rcParams`

### DIFF
--- a/ipykernel/pylab/config.py
+++ b/ipykernel/pylab/config.py
@@ -39,23 +39,10 @@ class InlineBackendConfig(SingletonConfigurable):
 class InlineBackend(InlineBackendConfig):
     """An object to store configuration of the inline backend."""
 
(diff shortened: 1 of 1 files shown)


Amicable-beep (commented):
I'm very new here. I don't understand.

Mottl:
> I'm very new here. I don't understand.

This commit fixes reading of user-defined parameters in `matplotlibrc`. Current implementation prevents setting `dpi`, `figsize` and some others in `matplotlibrc`.

PR #600, opened 2021-03-08
Some removal of ipython_genutils.py3compat.

Carreau:
As in some of the other Jupyter Repositories; this start to remove
ipython_genutils because of packaging problem in linux distribution
(getting rid of nose).

This is a step toward that.

Note that this changes is slightly stricter in some area where function
would previously accept string/bytes and now only accept one; but as far
as I can tell are generally only used with one of those since we are
Python 3 only.

### DIFF
--- a/ipykernel/tests/test_jsonutil.py
+++ b/ipykernel/tests/test_jsonutil.py
@@ -13,7 +13,6 @@ import nose.tools as nt
 
 from .. import jsonutil
 from ..jsonutil import json_clean, encode_images
-from ipython_genutils.py3compat import unicode_to_str
 
 class MyInt(object):
     def __int__(self):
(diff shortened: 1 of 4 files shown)


ipykernel/tests/test_jsonutil.py
```diff
@@ -80,14 +79,8 @@ def test_encode_images():
     encoded2 = json_clean(encode_images(encoded))
     assert encoded == encoded2
     
-    # test that we don't double-encode base64 str
-    b64_str = {}
-    for key, encoded in encoded.items():
-        b64_str[key] = unicode_to_str(encoded)
-    encoded3 = json_clean(encode_images(b64_str))
-    assert encoded3 == b64_str
```
Carreau, line 88: unicode to str is no-op, and then this code becomes strictly identical to the one with encoded2.

SylvainCorlay:
@Carreau are you still working on this?

PR #601, opened 2021-03-09
Replace import item from ipython_genutils to traitlets.

Carreau:
This is to start removing ipython_genutils dependency,
the functionality is identical except this one checks its parameter is a
str.


blink1073 (approved):
Thanks!

PR #604, opened 2021-03-16
Move Changelog to Standard Location

afshin:
- Move the change log to the standard top level `CHANGELOG.md` file.
- Copy the file into the docs source tree when building.
- Adds a test for the docs build.


blink1073 (approved):
Thanks!

Issue #540, opened 2020-08-20
DeprecationWarning has unclear origin: `should_run_async` will not call `transform_cell` automatically in the future


courtarro:
This deprecation appears when I have warnings enabled, as of some time within the past week or so, and it is unclear what its origin is. How would I go about finding what component is causing this warning to appear?

For the record, the following code causes the problem to appear:

```python
import warnings
import pandas as pd
warnings.simplefilter('always')
pd.read_csv('filename.csv')
```

I could turn off warnings, obviously, but I need them for other situations. Hiding the issue doesn't really fix it, of course. I have the standard contrib notebook extensions installed also.

Carreau:
Sorry those are just internal APIs.I transfered this issue to IPykernel repository.

hellodanylo:
Another workaround is to downgrade IPython to 7.10 until this issue is resolved in IPyKernel

jckling:
Same problem here

mattpodolak:
I got this error as well, it seems like if it's triggered once it will print out in ever cell afterward.

It first occurred when printing out values from a list:

```
post_to_comments_list[:10]
```

All the following cells triggered this warning as well, including this cell:

```
tst_arr = [1, 2, 3]
```

asford:
Does this need a owner? Would a fix for this fit in the 6.0 release timeline?

Carreau:
#605 is not a proper fix, it re-introduces the exact bug that was painfully fixed in CLI IPython.

Carreau:
Sorry it does not reintroduce it, the bug is still there in IPykernel ; it just mute the error message.

gansanay:
I am still getting this warning at each cell in jupyterlab. python==3.9.2, ipykernel==5.5.3, ipython==7.21.0.

cam4ani:
me too

Carreau:
Do you have a file and line number for this warning ? it would be helpful to know which lines trigger it.

stridera:
```
.venv/lib/python3.8/site-packages/ipykernel/ipkernel.py:283: DeprecationWarning: `should_run_async` will not call `transform_cell` automatically in the future. Please pass the result to `transformed_cell` argument and any exception that happen during thetransform in `preprocessing_exc_tuple` in IPython 7.17 and above.
  and should_run_async(code)
```

gboeing:
@Carreau this warning is still appearing with ipykernel=5.5.3, ipython=7.22.0, python=3.9.2:

```
[...]

PR #605, opened 2021-03-18
Fix Handling of shell.should_run_async

afshin:
Fixes ipython/ipykernel#540

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -282,9 +282,16 @@ class IPythonKernel(KernelBase):
             # default case: runner is asyncio and asyncio is already running
             # TODO: this should check every case for "are we inside the runner",
             # not just asyncio
+            preprocessing_exc_tuple = None
+            try:
+                transformed_cell = self.shell.transform_cell(code)
+            except Exception:
+                transformed_cell = code
+                preprocessing_exc_tuple = sys.exc_info()
+
             if (
                 _asyncio_runner
-                and should_run_async(code)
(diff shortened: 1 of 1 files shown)


blink1073 (commented):
Thanks!

jccurtis:
Yes thank you! Is there a planned release with this fix?

blink1073:

blink1073:
Done!  https://pypi.org/project/ipykernel/5.5.1/

blink1073:
I deleted https://pypi.org/project/ipykernel/5.5.1 because we would have to upgrade our ipython dependency to 7.17, which doesn't make sense in a patch release.  cf #607.  This will have to wait for the next minor or major release, whichever comes first.

Carreau:
Noooooooooo. 

That reintroduce the exact bug the change of API was meant to prevent. We have transformers that are stateful; they destroy data if called twice.

`self.shell.transform_cell(code)` should not be called twice in an execution loop. The warning is there for a reason.

afshin:
> That reintroduce the exact bug the change of API was meant to prevent. We have transformers that are stateful; they destroy data if called twice.
> 
> `self.shell.transform_cell(code)` should not be called twice in an execution loop. The warning is there for a reason.

Ah, did we miss passing `transformed_cell` and `preprocessing_exc_tuple` [here](https://github.com/ipython/ipykernel/blob/703510e0fb73dd49edb9154386737c2968d23b6f/ipykernel/ipkernel.py#L316) and perhaps [here](https://github.com/ipython/ipykernel/blob/703510e0fb73dd49edb9154386737c2968d23b6f/ipykernel/ipkernel.py#L331)?

PR #611, opened 2021-03-22
Flush control stream upon shutdown

SylvainCorlay:
The control stream was not properly flushed before shutting down the ZMQ context.

This fixes CI in `jupyter_client` with the pre-release.


JohanMabille (approved):

blink1073 (approved):
👍🏼

PR #612, opened 2021-03-23
bump Python to 3.8 in readthedocs.yml

minrk:
3.5 is EOL and docs builds have been failing.


blink1073:
Thanks Min!  I just added myself to the email notification list as well.

PR #616, opened 2021-03-25
Specify ipykernel in kernelspec

SylvainCorlay:
I think that it makes sense to specify (ipykernel) in the kernelspec to differentiate from xeus-python, or other potential python kernels (such as Slicer).

Eventually I would also like to use the proper Python logo for xeus-python.

Screenshot from 2021-03-25 16-24-25


blink1073 (approved):
👍🏼

Issue #609, opened 2021-03-21
Execute Requests Immediately Following and Error are Aborted in 5.5


MSeal:
In debugging https://github.com/nteract/testbook/issues/88, which was traced to an issue in testbook behavior causing empty execute response messages with ipykernel 5.5 in several tests, I believe we have tacked down the change in behavior to [this commit](https://github.com/ipython/ipykernel/commit/45afaab2d0779455b1bc9fbd2cc216af8bc21abc). It appears to have changed the error abort interaction such that any additional execute requests within `stop_on_error_timeout` after an error occurs are aborted.

The docs state `Requests that arrive within this [stop_on_error_timeout] window after an error will be cancelled.`. What was the intention for this feature and why did it change behavior in 5.5? The message sent are also odd if this was intentional behavior because there's no indicator in the execute_result that the intended execution was aborted. I feel an error message or status of sorts would be expected unless I am missing something here?

I believe this is unintended behavior, as `5.4.3` does not have this behavior and the `5.5.0` changelog has no mention of it. Additionally notebooks with `allow-error` flags could also hit this issue unexpectedly, depending on how the notebook executor was queuing / processing the requests. Testbook certainly hits this because it's often testing that something failed or trying optimistic execution paths. For now I think the only way around the issue is to wait after failed executions or set the `stop_on_error_timeout` to 0.

I've drafted the simplest example I could to demonstrate behavior with `jupyter_client`, please let me know if any additional details are needed / if this was intended and incidental behavior change:

```
In [1]: from jupyter_client.manager import start_new_kernel

In [2]: km, kc = start_new_kernel(kernel_name='python3')

# This is a dumbed down version of what nbclient does to poll until a cell has finished executing
In [3]: def poll_output_msg(parent_msg_id):
   ...:     msgs = []
   ...:     while True:
   ...:         msg = kc.iopub_channel.get_msg()
   ...:         if msg['parent_header'].get('msg_id') == parent_msg_id:
   ...:             msgs.append(msg)
   ...:         if msg['msg_type'] == 'status' and msg['content']['execution_state'] == 'idle':
   ...:             break
   ...:     return msgs
   ...: 

In [4]: def execute(req):
   ...:     return poll_output_msg(kc.execute(req))
   ...: 

# The first execution will result an a NameError, the second will not print post-error because it is executed immediately after
In [5]: a, b = execute("error_req"), execute("print('post-error')")
```

Here's the message contents showing the first cell a's error

```
In [6]: a
[...]

PR #618, opened 2021-03-29
Changed default timeout to 0.0 seconds for stop_on_error_timeout

MSeal:
Resolves https://github.com/ipython/ipykernel/issues/609

Tested on classic, lab, papermill, testbook, and nteract. The behavior matches 5.4 kernel behavior by default now, with the ability for each interface to set the default abort time if they wish. Testing this with a higher value override worked as expected with requests aborting within the given window of time, though none of the interfaces showed that an abort occurred, they simple show no result. Also `stop_on_error: false` works to ignore the setting (nteract has this on by default for some reason).

@glentakahashi I implemented the suggested change, but when testing if you short circuit the abort future, the run-all in classic and lab will no properly abort already queued cells that were sent with `Run All Cells`. It does work appropriately for papermill/nbclient/testbook since those interfaces wait for each cell completion to send the next cell to execute. Unlike the headless libraries, the browser notebook interfaces will send all the cells at once (in back-to-back `execute_request`s) when told to `Run All`, relying on the kernel queue to handle execution. If the abort isn't fired with an immediate delay the queue will not be dropped.

The reason for headed interfaces to do this is to partially protect users from not executing those cells if they lose network connection to the server. The ideal solution would be for the notebook server to do the queuing so that A) latency between kernel and request is minimized and B) the server can control continued execution when clients disconnect rather than relying on the kernel to know caller intent. But that's beyond the scope of the change here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changes in IPython kernel
 
+* Set `stop_on_error_timeout` default to 0.0 matching pre 5.5.0 default behavior with correctly working flag from 5.5.0.
+
 ## 5.5
 
 ### 5.5.0
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -120,7 +120,7 @@ class Kernel(SingletonConfigurable):
     _poll_interval = Float(0.01).tag(config=True)
 
     stop_on_error_timeout = Float(
-        0.1,
+        0.0,
         config=True,
         help="""time (in seconds) to wait for messages to arrive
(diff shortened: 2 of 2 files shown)


glentakahashi (approved):
That makes sense on the short circuit. Seems fine to be based on our conversation! I'm not a real collaborator though, so will need a real review

blink1073 (approved):
Thanks!

blink1073:

blink1073:
Released: https://pypi.org/project/ipykernel/5.5.2/

PR #627, opened 2021-03-31
Update Trio mode for compatibility with Trio >= 0.18.0

mehaase:
The Trio package renamed their trio.hazmat module to trio.lowlevel.


blink1073 (approved):
Thanks!

PR #632, opened 2021-04-09
Use native coroutines instead of tornado coroutines

SylvainCorlay:
- This fixes https://github.com/ipython/ipython/issues/11565 because tornado's wrapping of asyncio is a bit aggressive with creating Task objects which causes them to have a different `contextvars.Context`.
- However, this is a major change. I think this deserves a careful review - and be considered a backward incompatible change.
- If/when this goes in, we should also consider completely getting rid of tornado in ipykernel.

cc @afshin @minrk @Carreau.

Note: tests are passing except for the in-process kernel tests, for which we will need to make more functions async.

### DIFF
--- a/ipykernel/inprocess/client.py
+++ b/ipykernel/inprocess/client.py
@@ -11,6 +11,8 @@
 # Imports
 #-----------------------------------------------------------------------------
 
+import asyncio
+
 # IPython imports
 from traitlets import Type, Instance, default
 from jupyter_client.clientabc import KernelClientABC
@@ -173,8 +175,8 @@ class InProcessKernelClient(KernelClient):
         stream = kernel.shell_stream
         self.session.send(stream, msg)
         msg_parts = stream.recv_multipart()
-        kernel.dispatch_shell(msg_parts)
-
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(kernel.dispatch_shell(msg_parts))
         idents, reply_msg = self.session.recv(stream, copy=False)
         self.shell_channel.call_handlers_later(reply_msg)
 
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -566,10 +566,12 @@ class Kernel(SingletonConfigurable):
             self.execution_count += 1
             self._publish_execute_input(code, parent, self.execution_count)
 
-        reply_content = await self.do_execute(
+        reply_content = self.do_execute(
             code, silent, store_history,
             user_expressions, allow_stdin,
         )
+        if inspect.isawaitable(reply_content):
+            reply_content = await reply_content
 
         # Flush output before sending the reply.
         sys.stdout.flush()
@@ -593,7 +595,7 @@ class Kernel(SingletonConfigurable):
         if not silent and reply_msg['content']['status'] == 'error' and stop_on_error:
             await self._abort_queues()
 
-    async def do_execute(self, code, silent, store_history=True,
+    def do_execute(self, code, silent, store_history=True,
                    user_expressions=None, allow_stdin=False):
         """Execute user code. Must be overridden by subclasses.
         """
@@ -604,11 +606,14 @@ class Kernel(SingletonConfigurable):
         code = content['code']
         cursor_pos = content['cursor_pos']
 
-        matches = await self.do_complete(code, cursor_pos)
+        matches = self.do_complete(code, cursor_pos)
+        if inspect.isawaitable(matches):
+            matches = await matches
+
         matches = json_clean(matches)
         self.session.send(stream, "complete_reply", matches, parent, ident)
 
-    async def do_complete(self, code, cursor_pos):
+    def do_complete(self, code, cursor_pos):
         """Override in subclasses to find completions.
         """
         return {'matches' : [],
@@ -620,17 +625,20 @@ class Kernel(SingletonConfigurable):
     async def inspect_request(self, stream, ident, parent):
         content = parent['content']
 
-        reply_content = await self.do_inspect(
+        reply_content = self.do_inspect(
             content['code'], content['cursor_pos'],
             content.get('detail_level', 0),
         )
+        if inspect.isawaitable(reply_content):
+            reply_content = await reply_content 
+
         # Before we send this object over, we scrub it for JSON usage
         reply_content = json_clean(reply_content)
         msg = self.session.send(stream, 'inspect_reply',
                                 reply_content, parent, ident)
         self.log.debug("%s", msg)
 
-    async def do_inspect(self, code, cursor_pos, detail_level=0):
+    def do_inspect(self, code, cursor_pos, detail_level=0):
         """Override in subclasses to allow introspection.
         """
         return {'status': 'ok', 'data': {}, 'metadata': {}, 'found': False}
@@ -638,14 +646,16 @@ class Kernel(SingletonConfigurable):
     async def history_request(self, stream, ident, parent):
         content = parent['content']
 
-        reply_content = await self.do_history(**content)
+        reply_content = self.do_history(**content)
+        if inspect.isawaitable(reply_content):
+            reply_content = await reply_content
 
         reply_content = json_clean(reply_content)
         msg = self.session.send(stream, 'history_reply',
                                 reply_content, parent, ident)
         self.log.debug("%s", msg)
 
-    async def do_history(self, hist_access_type, output, raw, session=None, start=None,
+    def do_history(self, hist_access_type, output, raw, session=None, start=None,
                    stop=None, n=None, pattern=None, unique=False):
         """Override in subclasses to access history.
         """
@@ -698,7 +708,9 @@ class Kernel(SingletonConfigurable):
         self.log.debug("%s", msg)
 
     async def shutdown_request(self, stream, ident, parent):
-        content = await self.do_shutdown(parent['content']['restart'])
+        content = self.do_shutdown(parent['content']['restart'])
+        if inspect.isawaitable(content):
+            content = await content
         self.session.send(stream, 'shutdown_reply', content, parent, ident=ident)
         # same content, but different msg_id for broadcasting on IOPub
         self._shutdown_message = self.session.msg('shutdown_reply',
@@ -715,7 +727,7 @@ class Kernel(SingletonConfigurable):
         shell_io_loop = self.shell_stream.io_loop
         shell_io_loop.add_callback(shell_io_loop.stop)
 
-    async def do_shutdown(self, restart):
+    def do_shutdown(self, restart):
         """Override in subclasses to do things when the frontend shuts down the
         kernel.
         """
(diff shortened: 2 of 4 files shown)


ipykernel/kernelbase.py
```diff
@@ -214,7 +211,7 @@ def process_control(self, msg):
             self.log.error("UNKNOWN CONTROL MESSAGE TYPE: %r", msg_type)
         else:
             try:
-                yield gen.maybe_future(handler(self.control_stream, idents, msg))
+                await handler(self.control_stream, idents, msg)
```
minrk, line 214: This is a backward-incompatible change to *force* all handlers to be async. We could keep the maybe_future functionality with [inspect.isawaitable](https://docs.python.org/3/library/inspect.html#inspect.isawaitable):

```suggestion
                result = handler(self.control_stream, idents, msg)
                if inspect.isawaitable(result):
                    await result
```

ipykernel/kernelbase.py
```diff
@@ -810,16 +790,14 @@ def _topic(self, topic):
...
         self.shell_stream.flush()
         self._aborting = True
-
-        def stop_aborting(f):
+        def stop_aborting():
             self.log.info("Finishing abort")
             self._aborting = False
-
-        self.io_loop.add_future(gen.sleep(self.stop_on_error_timeout), stop_aborting)
+        await asyncio.sleep(self.stop_on_error_timeout)
```
minrk, line 799: This changes behavior substantially - the original behavior was to stop aborting after a short delay, to flush the shell queue, essentially.

This change means that whoever calls `_abort_queues` blocks until this is done. If that is in the same thread as the shell channel message handling, then nothing really happens to the queue, as messages waiting in the queue don't get cancelled. (This affects the stop_on_error behavior in cases like restart & run all)

To preserve the same behavior, use `call_later(self.stop_on_error_timeout, stop_aborting)`

minrk (commented):
Makes sense. One minor suggestion that we preserve the `maybe_future` functionality to allow non-async methods, since most are in fact not async. That's probably the biggest breaking change here.

Since asyncio doesn't have its own `maybe_future` we can be more simplistic and just use:

```python
result = maybe_async_api()
if inspect.isawaitable(result):
    result = await result
```

The main difference this has is that `concurrent.futures.Future` is not awaitable. Everything tornado returns is, though (>=5.0, at lest). I think that's okay, though.

ipykernel/kernelbase.py
SylvainCorlay: > `call_later(self.stop_on_error_timeout, stop_aborting)`

That's what I originally did, however this was breaking the tests.

ipykernel/kernelbase.py
SylvainCorlay: I like this solution.

Quick thought: should we try to avoid a situation in which some handlers are async and others not?

SylvainCorlay:
I could fix the in-process kernel tests.

ipykernel/inprocess/client.py
```diff
@@ -173,8 +175,8 @@ def _dispatch_to_kernel(self, msg):
         stream = kernel.shell_stream
         self.session.send(stream, msg)
         msg_parts = stream.recv_multipart()
-        kernel.dispatch_shell(msg_parts)
-
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(kernel.dispatch_shell(msg_parts))
```
SylvainCorlay, line 179: - `dispatch_shell` is now an async function.
 - before, it was a tornado coroutine, that was called synchronously in this function.

ipykernel/kernelbase.py
SylvainCorlay: > This change means that whoever calls _abort_queues blocks until this is done.

Actually I don't think that this is true. It will only block for the amount of time specified in `stop_on_error_timeout`.

ipykernel/inprocess/client.py
SylvainCorlay: cc @minrk. Let me know what you think about this.

(I think the previous behavior was not correct)

SylvainCorlay:
Also, should I remove other uses of tornado in the code base in this PR?

ipykernel/kernelbase.py
minrk: > should we try to avoid a situation in which some handlers are async and others not?

I don't think it's important, as long as all handlers are *allowed* to be async.

ipykernel/kernelbase.py
minrk: > It will only block for the amount of time specified in stop_on_error_timeout.

That's true, but it prevents the flushing that should be occurring while we wait for this timeout. The point of this timeout is to abort any requests we receive during this window. What does the failure look like when the original behavior is preserved?

I suspect this might be working in tests because the control thread doesn't block the main thread by default.

Aside: What would be even better would be to check the request date header and compare with clock time, but that's vulnerable to hard-to-diagnose clock drift, so probably unreliable.

ipykernel/inprocess/client.py
minrk: This assumes asyncio isn't running, which I guess is true?

To preserve the tornado behavior would be to use `asyncio.ensure_future(kernel.dispatch_shell(msg_parts))` which assumes the loop *is* (or at least will be) running.

Maybe add a check for if the current loop is running, and call `run_until_complete` if it's not, and `ensure_future` if it is.

The main difference between tornado and asyncio coroutines is that calling a tornado coroutine actually begins the call, whereas asyncio only creates a coroutine without starting any execution at all. `ensure_future` puts the coroutine on the loop.

minrk (commented):
> should I remove other uses of tornado in the code base in this PR?

I think it's sensible to do it in stages, since this is a concrete discrete change.

It looks like the main thing left is our use of queues. <del>However, asyncio's queue is [not threadsafe](https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue), unlike tornado's, and the whole point of our use of the queues is inter-thread communication. So I think it makes sense to deal with that in its own PR, and keep this one to just `async def`.<del> Edit: tornado's queues are also not threadsafe. They are only called via `add_callback`. But I can imagine there being subtle threadsafety issues, since we *do* use them across threads.

ipykernel/inprocess/client.py
SylvainCorlay: > This assumes asyncio isn't running, which I guess is true?

Absolutely. The loop is not running.

> The main difference between tornado and asyncio coroutines is that calling a tornado coroutine actually begins the call,

Yes - and the inprocess kernel is taking advantage of it, but never starting the loop! I think it probably a buggy behavior for the inprocess kernel.

ipykernel/kernelbase.py
SylvainCorlay: Fixed. I don't know what was wrong earlier with that.

Carreau:
That looks great; with this, capturing of IO and some of the cleanup that has been done recently it would make a compelling release. We may even use the occasion to drop a couple deprecated things.

SylvainCorlay:
> That looks great; with this, capturing of IO and some of the cleanup that has been done recently it would make a compelling release. We may even use the occasion to drop a couple deprecated things.

:+1: And debugger support!

blink1073 (approved):
👍🏼

ipykernel/kernelbase.py
SylvainCorlay: I added your suggestion for both shell and control handlers, raison a deprecation warning when the handlers is not awaitable.

ipykernel/inprocess/client.py
SylvainCorlay: This is the last outstanding review comment, but I am not quite sure how to act upon it. The event loop is not running in the in-process kernel, however, we could be in a context of a larger application that has an event loop and we can't nest them...

Is the use case you are trying to solve?

In any case, unless we can nest event loop, the behavior will not be exactly the same... Hence I am not sure we should do anything about it.

SylvainCorlay:
> It looks like the main thing left is our use of queues. However, asyncio's queue is not threadsafe, unlike tornado's, and the whole point of our use of the queues is inter-thread communication. So I think it makes sense to deal with that in its own PR, and keep this one to just async def. Edit: tornado's queues are also not threadsafe. They are only called via add_callback. But I can imagine there being subtle threadsafety issues, since we do use them across threads.

Tornado queues are not thread safe, but it does not mean you can't use them with threads. It only means you can't use their APIs from a different thread where the ioloop runs - except for add_callback which is documented as being threadsafe.

ipykernel/kernelbase.py
```diff
@@ -570,11 +577,9 @@ def execute_request(self, stream, ident, parent):
             self.execution_count += 1
             self._publish_execute_input(code, parent, self.execution_count)
 
-        reply_content = yield gen.maybe_future(
-            self.do_execute(
-                code, silent, store_history,
-                user_expressions, allow_stdin,
-            )
+        reply_content = await self.do_execute(
```
minrk, line 580: Might not be async

ipykernel/kernelbase.py
```diff
@@ -597,26 +602,25 @@ def execute_request(self, stream, ident, parent):
...
 
-    @gen.coroutine
-    def complete_request(self, stream, ident, parent):
+    async def complete_request(self, stream, ident, parent):
         content = parent['content']
         code = content['code']
         cursor_pos = content['cursor_pos']
 
-        matches = yield gen.maybe_future(self.do_complete(code, cursor_pos))
+        matches = await self.do_complete(code, cursor_pos)
```
minrk, line 618: Might not be async

ipykernel/kernelbase.py
```diff
@@ -625,45 +629,41 @@ def do_complete(self, code, cursor_pos):
...
         """
         return {'status': 'ok', 'data': {}, 'metadata': {}, 'found': False}
 
-    @gen.coroutine
-    def history_request(self, stream, ident, parent):
+    async def history_request(self, stream, ident, parent):
         content = parent['content']
 
-        reply_content = yield gen.maybe_future(self.do_history(**content))
+        reply_content = await self.do_history(**content)
```
minrk, line 653: Also here

ipykernel/kernelbase.py
```diff
@@ -308,7 +312,14 @@ def dispatch_shell(self, msg):
             except Exception:
                 self.log.debug("Unable to signal in pre_handler_hook:", exc_info=True)
             try:
-                yield gen.maybe_future(handler(self.shell_stream, idents, msg))
+                result = handler(self.shell_stream, idents, msg)
+                if inspect.isawaitable(result):
+                    await result
+                else:
+                    warnings.warn(
+                    "Message handlers should be awaitable",
```
minrk, line 320: I don't think there's a reason to deprecate this. It's fine and in fact preferable a lot of the time for these to not be async when they don't need to be.

minrk (commented):
👍 in general. I'd only add that I think we *shouldn't* deprecate synchronous handlers, so I think we should skip the deprecation message, and shouldn't convert existing `do_` methods and message handlers to async if they aren't already. There isn't really a benefit to that that I can see, and it does have a (tiny) performance cost.

ipykernel/inprocess/client.py
minrk: Adding some notes from testing - master definitely doesn't work if the eventloop is running, but it's less clear because tornado's generator-based coroutines execute up to the first yield (which is always the whole thing if the loop isn't running), whereas asyncio never runs anything when the loop isn't running. The code before accidentally assumes the loop isn't runnig, but the change here makes that explicit. It's not a new regression, but it make clearer that something that's never worked won't work. Something we can fix if people need it, but not in scope for this PR.

SylvainCorlay:
I addressed all of @minrk comments.

The windows python 3.7 test failure is the same as in master and is probably due to the release of importlib-metadata a few days a go.

blink1073:
Kicking CI

minrk:
Nice!

SylvainCorlay:
Many thanks to @minrk for the careful review.

blink1073:
Wow, good work y'all.  🎉

gsakkis:
I just upgraded ipykernel to 5.5.5 for this fix but it doesn't seem to be working for me:
JupyterLab
Am I missing something?

SylvainCorlay:
Yep, will be in 6.0.

PR #634, opened 2021-04-12
Use channel get_msg helper method

davidbrochart:
Use e.g. `kc.get_shell_msg()` instead of `kc.shell_channel.get_msg()`, because the latter is now always async in jupyter_client. The former is a higher-level method that runs `get_msg()` until complete for a blocking client.


davidbrochart:
The failing tests timed out, and they are using jupyter_client 6.1.12, so I am not sure what is the reason.

davidbrochart:
Trigger CI

davidbrochart:
This is now ready to review/merge.

blink1073 (approved):
Thanks!

PR #638, opened 2021-04-14
make deprecated shell_streams writable

minrk:
shell_streams is deprecated, but keep it working with deprecation warnings instead of breaking it

and handle debugpy_stream being undefined, since allow_none is True

This is part of avoiding breakage in ipyparallel to smooth the transition.

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -162,7 +162,10 @@ class IPythonKernel(KernelBase):
 
     def start(self):
         self.shell.exit_now = False
-        self.debugpy_stream.on_recv(self.dispatch_debugpy, copy=False)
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
LGTM!

minrk:
Added one more handling of optional new functionality: when control_thread is undefined, control messages would never be handled. This starts handling control messages, on whichever loop they are defined.

SylvainCorlay:
> Added one more handling of optional new functionality: when control_thread is undefined, control messages would never be handled. This starts handling control messages, on whichever loop they are defined.

I like that the add_callback is now called in the start function.

blink1073 (approved):
Thanks!

PR #644, opened 2021-04-18
Build docs only on ubuntu : add jobs to check docstring formatting.

Carreau:
This makes sure that docstring auto formatting won't change anything. Runs only on ubuntu + 3.9


blink1073 (approved):
Thanks!

PR #651, opened 2021-04-26
Fixes debugging with native coroutines

SylvainCorlay:
This function should not be async.


SylvainCorlay:
From a failed rebase of #632. Will merge when all is green and tag another alpha release.

PR #653, opened 2021-05-02
Import configure_inline_support from matplotlib_inline if available

ccordoba12:
- This correctly applies the rcParams declared in the inline backend configuration when using `%matplotlib inline`.
- This is necessary for IPython 7.23+, which requires `matplotlib_inline` now.
- I also think a new release in the `5.5` series is required to work with that IPython version.

Pinging @SylvainCorlay and @blink1073 about this one.

### DIFF
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -624,7 +624,10 @@ class ZMQInteractiveShell(InteractiveShell):
     def enable_matplotlib(self, gui=None):
         gui, backend = super(ZMQInteractiveShell, self).enable_matplotlib(gui)
 
-        from ipykernel.pylab.backend_inline import configure_inline_support
+        try:
+            from matplotlib_inline.backend_inline import configure_inline_support
+        except ImportError:
+            from ipykernel.pylab.backend_inline import configure_inline_support
 
         configure_inline_support(self, backend)
 


blink1073 (approved):
Thanks!

martinRenou:
This should be handled by https://github.com/ipython/ipykernel/pull/591

blink1073:
Ah, I see, thanks @martinRenou!

martinRenou:
> I also think a new release in the 5.5 series is required to work with that IPython version.

I remember trying and I saw it working, so it *should* be fine. Although I introduced an issue that https://github.com/ipython/ipython/pull/12940 fixes, but this code should not be triggered, ipykernel was already using its own `configure_inline_support` function IIRC.

ccordoba12:
> I remember trying and I saw it working, so it should be fine, although maybe not the configuration

It is working, I confirmed it in Spyder.

> but this code should not be triggered, ipykernel was already using its own configure_inline_support function IIRC.

I don't see a problem with it being triggered for now. I mean, the problem is that the inline backend is currently semi-broken and this will fix it until ipykernel 6 is released.

blink1073:
@ccordoba12 if you can make a PR to `5.5.x` I an cut a release.

ccordoba12:
Ok, thanks for understanding @blink1073! Will do right away.

PR #658, opened 2021-05-07
flush control queue prior to handling shell messages

minrk:
preserves control channel priority over shell channel

Always flush the control queue before processing shell messages. When control thread wasn't running, shell messages would often get handled before control messages that had arrived before they started processing, due to how the queues are flushed from zmq to processing queues and the removal of synchronization in #585.

This adds a flush immediately before any shell message is handled, which does:

- `control_stream.flush()` pops messages from zmq to `control_queue`
- float a threadsafe Future on `control_queue` to allow waiting for any pending messages to process (I first tried `control_queue.qsize()`, but that doesn't work because it goes to 0 when control *starts* processing the last request, not when it finishes)

This is much less of an issue when the control thread is running because control messages are usually quick, so the control queue is *usually* empty. So we could do this only when control thread is not running. However, that's not necessarily a safe assumption, and the behavior is now more correct in both cases, though it was usually fine with control_thread and always wrong without it.

with this, all ipyparallel tests are passing for me (#635)

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -4,6 +4,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 import asyncio
+import concurrent.futures
 from datetime import datetime
 from functools import partial
 import itertools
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
This looks good to me!

PR #659, opened 2021-05-07
stop control thread before closing sockets on it

minrk:
closing sockets while they are still in use by another thread is a recipe for SIGABRT.

skip `close(all_fds=True)` in favor of explicit cleanup of sockets

I believe this fixes the nbclient failures in test_many_parallel_notebooks mentioned [here](https://github.com/ipython/ipykernel/pull/635#issuecomment-827826766).

I'm not sure why that test would have triggered it and not others, but it was readily reproducible for me and doesn't happen anymore after this.

### DIFF
--- a/ipykernel/control.py
+++ b/ipykernel/control.py
@@ -15,7 +15,16 @@ class ControlThread(Thread):
         self.pydev_do_not_trace = True
         self.is_pydev_daemon_thread = True
 
-    def run(self): 
+    def run(self):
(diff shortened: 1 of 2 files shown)


davidbrochart (commented):
I don't know about the failures in ipykernel, but it fixed nbclient:
https://github.com/jupyter/nbclient/pull/146

davidbrochart:
Great, thanks!

PR #660, opened 2021-05-07
only store the parent header in parent_headers

minrk:
the old parent_header was actually the parent *message*, but we don't need to keep doing that

#657 makes `get_parent_header` a public API, so let's be more precise about actually returning the header


SylvainCorlay:
Thanks for this!

> Alternative here is to keep storing the full parent message and add a new get_parent method that's more symmetrical with set_parent

I don't have an opinion on this. Symmetric is nice...

minrk:
Symmetrical version in #661

PR #661, opened 2021-05-07
Add Kernel.get_parent to match set_parent

minrk:
the now-deprecated `_parent_header` was actually the parent message, but we don't need to keep having that mismatch in a new public API.

instead of `get_parent_header`, since it's actually returning the parent *message* not the parent header

Alternative to #660, which stores and returns just the header instead of the whole request.

We've only ever *used* the header, so either one ought to work fine

I think I have a slight preference for this one over #660, but I don't think it makes much difference. If we find that we ever need public API access to the content or metadata of the request, only this one will do.

closes #660

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -133,17 +133,17 @@ class Kernel(SingletonConfigurable):
 
     # track associations with current request
     _allow_stdin = Bool(False)
-    _parent_headers = Dict({"shell": {}, "control": {}})
+    _parents = Dict({"shell": {}, "control": {}})
     _parent_ident = Dict({'shell': b'', 'control': b''})
 
     @property
     def _parent_header(self):
         warnings.warn(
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
LGTM. Indeed, symetry is nicer.

SylvainCorlay (approved):

ipykernel/kernelbase.py
```diff
@@ -975,7 +975,7 @@ def raw_input(self, prompt=''):
         return self._input_request(
             str(prompt),
             self._parent_ident["shell"],
-            self.get_parent_header("shell"),
+            self.get_parent("shell"),
```
SylvainCorlay, line 978: The value returned by `get_parent("shell")` is the full parent message or just the header?

SylvainCorlay: Ah, the answer to my question was in the changes in the alternative pr. Merging!

PR #665, opened 2021-05-10
fix: Backwards compat with older versions of zmq

mlucool:
zmq.ROUTING_ID was added in libzmq 4.2.3 as an alias for IDENTITY.


mlucool:
>@mlucool your commit author does not match you github email; not sure if that's matter to you.

I fixed it, but it was mostly correct.

> I'm guessing this should be backported to the 5.5.x branch, or 5.x branch.

This only is for the debugger, which I don't think is in 5.x.

SylvainCorlay:
cc @JohanMabille

JohanMabille:
> This only is for the debugger, which I don't think is in 5.x.

Indeed the debugger is in 6.x only.

PR #669, opened 2021-05-13
Keep preferring SelectorEventLoop on Windows

minrk:
Selector is still preferable when ~all out events are on zmq sockets, which must use the add_reader methods missing from Proactor. Tornado 6.1 puts these events in a background thread when Proactor is used, which we can avoid by using Selector.

cf https://github.com/jupyter/notebook/pull/6052

reverts that part of #564

I don't understand the hang mentioned in #563 - if/where that happens and is caused by this patch, it is presumably multiple event loops with events, but only one running (waiting on async events scheduled on a non-running loop will never finish). That should be traceable and avoidable, though, if it can be reproduced.


blink1073 (approved):
Thanks!

blink1073:

PR #678, opened 2021-05-25
Do not override matplotlib params

jbc913:


martinRenou:
@jbc913 the inline matplotlib back-end does not live in ipykernel anymore. You might want to apply your changes to https://github.com/ipython/matplotlib-inline instead

PR #683, opened 2021-06-04
Make the isatty method of OutStream return True

### LINKED ISSUE #268: OutStream.isatty
I'm using jupyter notebook and `sys.stdout` is `ipykernel.iostream.OutStream`.
And `sys.stdout.isatty()` is False.

Since the output is interactive/terminal-like, I expect it to be true. Also see [comments in cpython](https://github.com/python/cpython/blob/master/Lib/_pyio.py#L469).

Can this be changed? This should be simply two lines of code.

### LINKED ISSUE #680: `ipykernel.iostream.OutStream.isatty()` should be `True`.
I was redirected from [qtconsole](https://github.com/jupyter/qtconsole). I posted [this Issue](https://github.com/jupyter/qtconsole/issues/486) over there. 

A stream's method isatty should return whether it is an 'interactive' stream. I think that means whether it is an interactive console. 

Although *qtconsole* is an interactive stream their `sys.stdout.isatty()` returns `False`. This is because *qtconsole* uses `ipykernel.iostream.OutStream` as standard streams for `sys.stdout` and `sys.stderr`. This class inherits the method `isatty()` from `io.TextIoBase` which isn't supposed to be an interactive stream by default.

https://github.com/ipython/ipykernel/blob/fda115626dd434f49a4d8f8e55080d5fd2c2e2b9/ipykernel/iostream.py#L276

One of the current main developers of `qtconsole` told me:

> Please open this issue in the ipykernel repo because we can't solve it here and a change as the one you're suggesting will also affect the Jupyter notebook and JupyterLab.

***
This Issue relates to #268 .

peendebak:
The ` ipykernel.iostream.OutStream` represents an interactive stream and should there have a `isatty` method returning `True`.
Also see https://github.com/ipython/ipykernel/issues/268, https://github.com/ipython/ipykernel/issues/680, https://github.com/willmcgugan/rich/pull/1221

Fixes #680 
Fixes #268

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -383,11 +383,11 @@ class OutStream(TextIOBase):
 
     def isatty(self):
         """Return a bool indicating whether this is an 'interactive' stream.
-        
+
         The standard ipykernel streams are assumed to be interactive.
         """
         return True
-    
+
     def _setup_stream_redirects(self, name):
         pr, pw = os.pipe()
         fno = getattr(sys, name).fileno()
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -333,7 +333,7 @@ class OutStream(TextIOBase):
             the file descriptor by its number. It will spawn a watching thread,
             that will swap the give file descriptor for a pipe, read from the
             pipe, and insert this into the current Stream.
-        isatty: bool (default, False)
+        isatty : bool (default, False)
             Indication of whether this stream has termimal capabilities (e.g. can handle colors)
 
         """
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -320,7 +320,7 @@ class OutStream(TextIOBase):
             self._exc = sys.exc_info()
 
     def __init__(
-        self, session, pub_thread, name, pipe=None, echo=None, *, watchfd=True, isatty=True,
+        self, session, pub_thread, name, pipe=None, echo=None, *, watchfd=True, isatty=False,
     ):
         """
         Parameters
@@ -333,7 +333,7 @@ class OutStream(TextIOBase):
             the file descriptor by its number. It will spawn a watching thread,
             that will swap the give file descriptor for a pipe, read from the
             pipe, and insert this into the current Stream.
-        isatty: bool (default, True)
+        isatty: bool (default, False)
             Indication of whether this stream has termimal capabilities (e.g. can handle colors)
 
         """
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -381,6 +381,13 @@ class OutStream(TextIOBase):
             else:
                 raise ValueError("echo argument must be a file like object")
 
+    def isatty(self):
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
This seems reasonable to me, thank you!   I'll leave this open until Monday in case another maintainer knows of a reason not to enable this.

real-yfprojects:
You should add Closes #680 and closes #268 .
Also thank you for creating the pull request. I didn't  know this was the faster way to go.

ccordoba12:
Although this would be quite useful to support Rich and Colorama in Spyder and the notebook, I don't know if it could have side effects on IPython's display system. I mean, there should be a reason why the `OutStream` is expected to behave as a `TextIO` object.

@minrk, @takluyver, @Carreau, what do you think?

Carreau:
I think this is a bad idea. A lot of things assume that `isatty() == True`, mean you can put the terminal in raw mode and/or support cursor motion up/down to redraw previous lines.

Things like pagers ( less, more) and many others will start to misbehave if you set tty to true.

ccordoba12:
Thanks for the input @Carreau! What if we make this change configurable then? That way, users that know what they are doing could enable this functionality for Rich and Colorama. In Spyder we could also expose this option in our Preferences system and inform users to activate it if we detect they are working with those libraries.

dsblank:
It is too bad that there is a single flag that signals if an output can handle simple line control (like backspace), colors, and full cursor movements. We have a commercial product and almost 100% of the non-interactive output works well when istty() returns True.

Carreau:
No objections to making in configurable..I believe there might be something
like terminal/termcap that should let us hurry more in depth which
capabilities the terminal has, but I don't think it's standard or has
python binding.

dsblank:
> No objections to making in configurable

Agreed, that sounds like a good approach!

peendebak:
@dsblank @Carreau @blink1073 @ccordoba12 I will make the option configurable, so that projects using the `ipykernel` can pick whichever mode they want.

The only remaining question is what the default value will be for this option. The safe option would be `False`. On the other hand,  `True` might be the better option for a larger number of cases.

ccordoba12:
> The only remaining question is what the default value will be for this option. The safe option would be False. On the other hand, True might be the better option for a larger number of cases.

It needs to be False to be on the safe side and not to change the current behavior.

peendebak:
@blink1073 The additional commits are under another github account. Could you approve the CI?

blink1073:
> Could you approve the CI?

Done!

blink1073:
Looks like tests need updating again

blink1073:
The linter flagged one thing: https://github.com/ipython/ipykernel/pull/683/checks?check_run_id=2777466679#step:8:21

peendebak:
> The linter flagged one thing: https://github.com/ipython/ipykernel/pull/683/checks?check_run_id=2777466679#step:8:21

@blink1073  Thanks, I fixed the formatting.

blink1073:
Thanks for the contribution @peendebak!

PR #684, opened 2021-06-07
call metadata methods on abort replies

minrk:
ensures aborted messages have consistent metadata with real replies


blink1073 (approved):
Thanks!  I kicked CI

PR #685, opened 2021-06-14
Return len of item written to OutStream

### LINKED ISSUE #682: `ipykernel.iostream.OutStream.write()` fails to return the number of characters written
The current implementation of `ipykernel.iostream.OutStream.write()` ([source](https://github.com/ipython/ipykernel/blob/fda115626dd434f49a4d8f8e55080d5fd2c2e2b9/ipykernel/iostream.py#L479)) does not adhere to the Python `TextIOBase` [API](https://docs.python.org/3/library/io.html#io.TextIOBase.write), which calls for `write()` to return the number of characters written (the current implementation implicitly returns `None`).  This causes problems for clients that verify that the data was reported to be correctly written to the output stream (see, e.g., Pyomo/pyomo#1997)

As IPython handles the actual output asynchronously, I believe that an appropriate fix would just be to to return `len(string)` from `OutStream.write()`.

Carreau:
Closes #682

Remove the piece of logic that handle `not isinstance(str)` it is a
leftover from Python 2, in pure Python, sys.stdout.write only accepts
str, therefore we have no reason not to do the same.

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -487,7 +487,21 @@ class OutStream(TextIOBase):
             self.session.send(self.pub_thread, 'stream', content=content,
                 parent=self.parent_header, ident=self.topic)
 
-    def write(self, string):
+    def write(self, string: str) -> int:
+        """Write to current stream after encoding if necessary
+
+        Returns
+        -------
+        len : int
+            number of items from input parameter written to stream.
+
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!

Carreau:
Thanks !

PR #692, opened 2021-06-16
Remove pin on Jedi because that was already fixed in IPython

ccordoba12:
Just in case this is used by somebody.


Carreau:
+1 restarted the test it was a timeout of GHA.

blink1073 (approved):
Thanks!

PR #704, opened 2021-07-01
Fix Tk and asyncio event loops

ccordoba12:
I noticed this when running the Spyder test suite against the newly released ipykernel 6.


blink1073 (approved):
Thanks!

PR #708, opened 2021-07-02
Make json_clean a no-op for jupyter-client >= 7

### LINKED ISSUE #706: Remove json_clean
cc. @SylvainCorlay @minrk @afshin 
Also @ibdafna, you might be interested.

### Problem

The [`json_clean`](https://github.com/ipython/ipykernel/blob/master/ipykernel/jsonutil.py#L71) function is responsible for some slowness in ipykernel, both in comms (jupyter widgets) and in the kernel protocol.

This function loops over all communication messages and [makes copies of all lists-like](https://github.com/ipython/ipykernel/blob/master/ipykernel/jsonutil.py#L123-L128) and [dicts](https://github.com/ipython/ipykernel/blob/master/ipykernel/jsonutil.py#L130-L142) in there. This is fine when it comes to execution requests, but with comms it can be very slow as people are passing entire dataframes through them (see beakerx_tabledisplay and ipydatagrid for example).

After the object has been cleaned by `json_clean`, it is [serialized to a JSON string by jupyter_client](https://github.com/jupyter/jupyter_client/blob/47cfe68c075fc7f31c0d0892ada29509910369b2/jupyter_client/session.py#L94-L100) inside the `send` and `serialize` methods by [`self.pack`](https://github.com/jupyter/jupyter_client/blob/47cfe68c075fc7f31c0d0892ada29509910369b2/jupyter_client/session.py#L341). 

### Solution?

We should remove this `json_clean` method, and instead, we should use the official JSON library API to do the "cleaning" **while** serializing the message, directly in `jupyter_client`. This is already achieved for dates (see [this](https://github.com/jupyter/jupyter_client/blob/47cfe68c075fc7f31c0d0892ada29509910369b2/jupyter_client/session.py#L97) and [this](https://github.com/jupyter/jupyter_client/blob/47cfe68c075fc7f31c0d0892ada29509910369b2/jupyter_client/jsonutil.py#L93)) and could be done for other objects (bytes, etc).

### More context

I have to say that I tried making the `json_clean` function a no-op locally, and it worked perfectly with most of the Notebooks I tried (Folium, Plotly, Altair, Interactive Widgets, ipydatagrid, IPython.display)

Only the Matplotlib inline backend was not working anymore, due to incorrect serializing of bytes (which could be done in jupyter_client, see the proposed solution).

martinRenou:
Fix #706

This should not be merged before a potential release of jupyter_client with https://github.com/jupyter/jupyter_client/pull/664, otherwise, Matplotlib will not work anymore in Jupyter Notebook/Lab

### DIFF
--- a/ipykernel/datapub.py
+++ b/ipykernel/datapub.py
@@ -66,6 +66,6 @@ def publish_data(data):
         DeprecationWarning,
         stacklevel=2
     )
-    
+
     from ipykernel.zmqshell import ZMQInteractiveShell
     ZMQInteractiveShell.instance().data_pub.publish_data(data)
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -7,9 +7,14 @@ from zmq.utils import jsonapi
 from tornado.queues import Queue
 from tornado.locks import Event
 
-from .compiler import (get_file_name, get_tmp_directory, get_tmp_hash_seed)
-
 from IPython.core.getipython import get_ipython
+
+try:
+    from jupyter_client.jsonutil import json_default
+except ImportError:
+    from jupyter_client.jsonutil import date_default as json_default
+
+from .compiler import (get_file_name, get_tmp_directory, get_tmp_hash_seed)
 
 # This import is required to have the next ones working...
 from debugpy.server import api  # noqa
@@ -170,7 +175,12 @@ class DebugpyClient:
     def _send_request(self, msg):
         if self.routing_id is None:
             self.routing_id = self.debugpy_stream.socket.getsockopt(ROUTING_ID)
-        content = jsonapi.dumps(msg)
+        content = jsonapi.dumps(
+            msg,
+            default=json_default,
+            ensure_ascii=False,
+            allow_nan=False,
+        )
         content_length = str(len(content))
         buf = (DebugpyMessageQueue.HEADER + content_length + DebugpyMessageQueue.SEPARATOR).encode('ascii')
         buf += content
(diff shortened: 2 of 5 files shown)


JohanMabille:
Please notice that this function is also used by the [debugger](https://github.com/ipython/ipykernel/blob/master/ipykernel/debugger.py#L423). It should be fixed before merging this.

martinRenou:
Thanks for the review @JohanMabille. I hadn't rebased.

martinRenou:
This should be resolved now. Testing message dumps as jupyter_client does it.

minrk:
👍 to the change in general. Can we get matplotlib working? Or can it not be done here because the mpl backend was moved to its own package?

What about including a backport of the json_default function from https://github.com/jupyter/jupyter_client/pull/664?

martinRenou:
Thanks a lot for your review @minrk ! 

> Can we get matplotlib working? 

This needs https://github.com/jupyter/jupyter_client/pull/664 which does the base64 serializing https://github.com/jupyter/jupyter_client/pull/664/files#diff-fda072ac1fb480ce6ce75694ef6e4144e811d8443191be1c4f233a9ee3fd02a1R108

I am not sure we can fix it here, as ipykernel is not responsible for the serialization.

> What about including a backport of the json_default function from jupyter/jupyter_client#664?

You mean here in ipykernel?

martinRenou:
This requires `jupyter_client>=7`, I've updated the requirement in `setup.py` accordingly

blink1073 (approved):
Thank you!

martinRenou:
Thanks!!

PR #709, opened 2021-07-02
[bugfix] fix in setup.py (comma before appnope)

jstriebel:
The two dependency-strings for `matplotlib-inline` and `appnope` were merged which lead to errors such as `Invalid constraint (matplotlib-inline (<0.2.0appnope,>=0.1.0) ; platform_system == "Darwin") found in ipykernel-6.0.0rc0 dependencies`.


blink1073 (approved):
Thank you!

PR #710, opened 2021-07-02
Remove CachingCompiler's filename_mapper

martinRenou:
This seems to be some leftover from old debugger implementation

cc. @JohanMabille


blink1073 (approved):
Thanks!

PR #714, opened 2021-07-09
Add upper bound to dependency versions

martinRenou:
This might save us from some troubles if a major release of one of those dependencies arrives (thinking of jupyter-client 7.0 coming)

If you are fine with it, I will open a paired PR to the conda-forge feedstock


ccordoba12:
This is a very good idea to me. Thanks for doing that @martinRenou!

setup.py
```diff
@@ -76,10 +76,10 @@ def run(self):
     install_requires=[
         'importlib-metadata<4;python_version<"3.8.0"',
         'debugpy>=1.0.0',
```
davidbrochart, line 78: Maybe also pin debugpy?
```suggestion
        'debugpy>=1.0.0,<2.0',
```

davidbrochart:
Thanks! I'll update the feedstock.

PR #717, opened 2021-07-12
Minimal flake8 config

Carreau:
This disable most common issues that are not critical like whitespace,
at least for now, so that we can fix all the issues and put flake8 in
CI with an exist status.

Then we can decide to renable (or not) those checks.


blink1073 (approved):
Fantastic, thank you for all this cleanup!

PR #721, opened 2021-07-12
Do not use bare except

Carreau:
Part of the #717 PR-group.

Bare excepts replaced by except Exception/BaseException.

Here I'd like to have a second opinion in each case whether Exception/BaseException is the right thing.


blink1073 (approved):
Thanks!

PR #722, opened 2021-07-12
Remove unused variables

Carreau:
Unused variables are a code smell.
Removing them make it clearer we expect the functions called to be
called for their side-effects.

Part of the #717 PR-group.


blink1073 (approved):
👍🏼

PR #723, opened 2021-07-12
Replace non-existing function.

Carreau:
This was forgotten as part of an earlier pass, when safe_unicode (py2
compat) was removed. safe_unicode is undefiled now and would lead to
crash.

Part of the #717 PR-group.


blink1073 (approved):
Thanks!

PR #726, opened 2021-07-13
[WIP] Fix exception raised by `OutStream.write`

### LINKED ISSUE #725: The `OutStream.write` method raises `ValueError` instead of `TypeError`
From comments on the commit that added the code, I believe the intent is for `OutStream.write` to ducktype like `sys.stdout.write`, but by inspection, I believe to correctly ducktype, `OutStream.write` should raise `TypeError`.

E.g.
```
$ python
Python 3.8.8 | packaged by conda-forge | (default, Feb 20 2021, 16:12:38)
[Clang 11.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.stdout.write(b'')
  File "<stdin>", line 1, in <module>
TypeError: write() argument must be str, not bytes
```

This is confusing some libraries we are using in conjunction with `ipykernel` specifically `eliot` for logging.  I believe I see the fix [here](https://github.com/ipython/ipykernel/blob/master/ipykernel/iostream.py#L501), and a PR will be incoming.

SimonKrughoff:
In order to conform with the behavior of `sys.stdout.write` this method should return a `TypeError` rather than a `ValueError`.  This PR also adds an 'f' to the to make the f-string evaluate the `type(string)` and return more helpful information in the exception message.

After some fiddling, I was not able to get the nosetests to pass on my local machine by following the instructions in the READM.md for this repo.  The contributing docs indicated that a pass in actions would be sufficient, but I'm open to suggestions for what I might be doing wrong.

The PR is intended to fix #725

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -498,7 +498,7 @@ class OutStream(TextIOBase):
         """
 
         if not isinstance(string, str):
-            raise ValueError(
+            raise TypeError(
                 "TypeError: write() argument must be str, not {type(string)}"
             )
 
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -499,7 +499,7 @@ class OutStream(TextIOBase):
 
         if not isinstance(string, str):
             raise TypeError(
-                "TypeError: write() argument must be str, not {type(string)}"
+                f"write() argument must be str, not {type(string)}"
             )
 
         if self.echo is not None:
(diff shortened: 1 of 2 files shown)


thomasaarholt:
Agreed! Although I think you could improve that error message further by removing the superfluous "TypeError" part of string of the raised error :)

SimonKrughoff:
I've rebased and removed the extra `TypeError:`.  Thanks for the suggestion!

thomasaarholt:
It looks like all the tests here are passing. Maybe @Carreau would like to take a look? It's only 3 months since the ValueError was introduced :)

Carreau (approved):
That's looks good to me. I didn't had much time to work on IPykernel recently, 
I'll try to work on solution to either free some of my time, or free someone else time to do so.

Issue #730, opened 2021-07-15
Exception on ipc file cleanup resulting in socket leak


amorenoz:
Running an ipython kernel with `--transport=ipc` and stopping it (via sending an `quit` command though a jupyter-client), generates the following  exception:

```
Traceback (most recent call last):
  File "/home/amorenoz/.local/share/virtualenvs/insights-cmd-nQG3uYIv/lib/python3.9/site-packages/ipykernel/kernelapp.py", line 242, in cleanup_connection_file
    self.cleanup_ipc_files()
  File "/home/amorenoz/.local/share/virtualenvs/insights-cmd-nQG3uYIv/lib/python3.9/site-packages/jupyter_client/connect.py", line 426, in cleanup_ipc_files
    for port in self.ports:
TypeError: %i format: a number is required, not str
```

The socket files are not cleaned up and leaked

PR #731, opened 2021-07-15
RFC: kernelapp: rename ports variable to avoid override

amorenoz:
As described in #730, it seems ConnectionFileMixin is expecting `ports` to be List[int].

https://github.com/jupyter/jupyter_client/blob/5742d84ca2162e21179d82e8b36e10baf0f8d978/jupyter_client/connect.py#L386

IPKernelApp is overriding such attribute and changing its type, which generates the error described.

This is the simplest fix I could come up with. Since I am not familiar with the project, please take this as a suggestion.

#Fixes 730

Signed-off-by: Adrian Moreno <amorenoz@redhat.com>


blink1073 (approved):
Thank you!

Issue #645, opened 2021-04-21
Master build failures with Python 3.7


SylvainCorlay:
There may have been some new PyPI release of our dependencies causing these failures.

```
argcomplete 1.12.2 has requirement importlib-metadata<4,>=0.23; python_version == "3.7", but you have importlib-metadata 4.0.1.
```

ltalirz:
Hi @SylvainCorlay  , since we need importlib-metadata >=4,  this fix forced us to use an earlier version of ipykernel.

I just wanted to point out that the latest v1.12.3 patch release of argcomplete has bumped the importlib-metadata upper limit to <5:
https://github.com/kislyuk/argcomplete/blob/develop/setup.py#L8

I'll open a PR to bump this

PR #738, opened 2021-08-12
bump importlib-metadata limit for python<3.8

ltalirz:
Related to #645

The importlib-metadata version was limited to <4 due to a corresponding restriction in the argcomplete package. 
The latest argcomplete patch release has bumped the limit to <5.

Note: ipykernel does not depend on argcomplete directly. 
I could also add an explicit dependency `argcomplete>=1.12.3` to select latest version if that is preferred

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,7 @@ setup_args = dict(
     python_requires='>=3.7',
     install_requires=[
         'importlib-metadata<5;python_version<"3.8.0"',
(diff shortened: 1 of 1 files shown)


blink1073:
> I could also add an explicit dependency argcomplete>=1.12.3 to select latest version if that is preferred

👍🏼

blink1073 (approved):
Thank you!

ltalirz:
Thanks for the quick merge!

PR #741, opened 2021-08-13
Add Support for Message Based Interrupt

afshin:
Implements handling of `interrupt_request` [message](https://jupyter-client.readthedocs.io/en/stable/messaging.html#kernel-interrupt), which is now possible due to the control channel being on a separate thread.
Does not implement handling on Windows, which is less likely to be used for remote kernels.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -10,6 +10,7 @@ from functools import partial
 import itertools
 import logging
 import inspect
+import os
 from signal import signal, default_int_handler, SIGINT
 import sys
 import time
@@ -198,7 +199,7 @@ class Kernel(SingletonConfigurable):
         'inspect_request', 'history_request',
         'comm_info_request', 'kernel_info_request',
         'connect_request', 'shutdown_request',
-        'is_complete_request',
+        'is_complete_request', 'interrupt_request',
         # deprecated:
         'apply_request',
     ]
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Excellent, thank you!

blink1073:
cc @SylvainCorlay @kevin-bates

SylvainCorlay:
Thanks for this!
Happy to see ipykernel move closer to fully implementing the protocol ;)

ccordoba12:
I think this is not enough because it's missing a call to perform the request in jupyter_client, just like there's one for shutdown:

https://github.com/jupyter/jupyter_client/blob/9c0b4c0e28a9c17d9c14123522415b2048bd9522/jupyter_client/client.py#L762-L781

Otherwise, how is it possible to perform an interrupt from the kernel client?

kevin-bates:
Hi @ccordoba12 - the interrupt machinery comes from the [`KernelManager`](https://github.com/jupyter/jupyter_client/blob/cbc29c9ae91b9a40c61b4b96937efaae4ea8c809/jupyter_client/manager.py#L536-L539) by virtue of the kernelspec indicating support for message-based interrupts.

Hmm - should we make signal-based interrupts the `else` so that kernelspecs that don't specify a mode attempt at least that form of interrupt?

blink1073:
According to the [docs](https://jupyter-client.readthedocs.io/en/stable/messaging.html#kernel-interrupt), it is opt-in "For this to work, the kernels kernelspec must set interrupt_mode to message".  I think the question is whether the `ipykernel` default spec should be message based.

kevin-bates:
ok - I understood the question to be one of missing logic - sorry about that.

My question was more general and I just found the answer in that ['signal' will be used if `interrupt_mode` is not specified](https://github.com/jupyter/jupyter_client/blob/cbc29c9ae91b9a40c61b4b96937efaae4ea8c809/jupyter_client/kernelspec.py#L38), which, now that I recall, have seen before. 

I'll let the maintainers decide what the desired mode should be. :smile:

ccordoba12:
> the interrupt machinery comes from the KernelManager by virtue of the kernelspec indicating support for message-based interrupts.

Yeah, but if you connect to a kernel started elsewhere (for example, locally to a kernel in a server, which Spyder users do a lot), then you don't have access to a kernel manager, just a kernel client. And in that case it'd be really nice to be able to send an interrupt message using the control channel (like you can now to perform a shutdown with the method I referenced above).

I thought this PR was about that, but it seems I was mistaken, sorry.

kevin-bates:
I'm sorry, I probably steered this discussion in the wrong direction and I apologize.  Please continue.

ccordoba12:
No worries @kevin-bates. Let's ping @davidbrochart to see what he thinks about this one.

@davidbrochart, do you think it's worth it to add a message-based interrupt in jupyter-client?

SylvainCorlay:
I think it is worth it just for the sake of completeness.

ccordoba12:
Agreed. I'll open a PR in jupyter-client so that (hopefully) it can be part of its 7.0 version.

PR #748, opened 2021-08-17
Remove Nose skipIf in favor of pytest

Carreau:
We could totally remove the skipif, but the we would need to have a min version of jupyter_client.

cc @SylvainCorlay as I believe you are the one who introduced this and comm_info.


blink1073 (approved):
Thank you!

PR #753, opened 2021-08-27
Implemented deep variable inspection

JohanMabille:
This implements the deep variable inspection after executing a cell. The output is consistent with that of the variable request, after the code has stopped on a breakpoint.

inspect_deep_variables


blink1073 (approved):
Thank you!

PR #756, opened 2021-08-31
Add dependency on IPython genutils.

### LINKED ISSUE #755: IPykernel is broken with IPython 7.27.0+
That release removed its dependency on `ipython_genutils`, but IPykernel is still using it in several places. Since it was not declared directly in `setup.py`, the package is broken when installed in a clean env.

@Carreau, could you take a look at this one?

Carreau:
this was an implicit dependency via IPython, which does not requires it
anymore and traitlets, which also does not requires it anymore since
today.

Closes #755

### DIFF
--- a/ipykernel/connect.py
+++ b/ipykernel/connect.py
@@ -7,7 +7,6 @@ import json
 import sys
 from subprocess import Popen, PIPE
 
-from ipython_genutils.path import filefind
 
 import jupyter_client
 from jupyter_client import write_connection_file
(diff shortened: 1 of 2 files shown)


Carreau:
@blink1073 It look like we'll need another release.

blink1073:
Yep, will do

Carreau:
Thanks ! I've seen the releases are automatic now !

I guess you could also trigger Jupyter releaser when a tag is pushed :-)

blink1073:
For security and accountability reasons, the releases are manual using a specific user's scoped PYPI token

blink1073:
(manual in that I am running the workflows from my fork of releaser)

blink1073:
Done!  And from my phone. 😀

https://github.com/ipython/ipykernel/releases/tag/v6.3.1

PR #761, opened 2021-09-01
Update some warnings with instructions and version number.

Carreau:
I came across this when trying to reproduce an unrelated bug with
someone in an outdated environment, but at least having the version
number will make us more confident when e remove the conditional branch
later.


Carreau:
@minrk There seem to be unexpected crash with a NameError `should_crash` when testing downstream ipyparallel 7.0.0b2, not sure if you want to have a look.

minrk:
Weird, I'll look into it. Looking at the test, that would normally be caused by cloudpickle somehow being invoked.

PR #766, opened 2021-09-06
Don't assume kernels have loops.

Carreau:
In particular the in-process kernels don't.

This does not fix all the issues in particular the quit() and exit()
autocaller that become no-op, but that's another story


blink1073 (approved):
Thank you!

Issue #712, opened 2021-07-08
Please make the dependency on debugpy optional


mgorny:
DebugPy is a packaging nightmare, and doesn't seem to support non-x86 platforms. Could you please make the dependency optional?

JohanMabille:
That would break the debugging feature of ipykernel. I'm not sure we want to require users to do extra manipulations to have the debugger in ipykernel because of technical difficulty to package debugpy. Besides, I'm not sure to understand what you mean by "DebugPy doesn't seem to support non-x86 platforms":  DebugPy has been packaged for ppc and aarch64 on conda-forge and seems to work.

mgorny:
Well, right now forcing debugpy means we can't package new ipykernel at all due to bug #715. Though I suppose that's a problem of its own, and needs to be fixed independently of debugpy.

frenzymadness:
It seems that the latest stable debugpy is not compatible with Python 3.10, although some commits exist in the main branch.

frenzymadness:
I've tried to update ipykernel in Fedora to the latest stable release but I've failed because of debugpy. Debugpy itself is not yet compatible with Python 3.10 even when you use the content of the main branch. According to upstream, the reason is that the bundled pydevd does not support 3.10 fully yet: https://github.com/microsoft/debugpy/issues/714
But this is not the biggest issue I have. The problem is that in Fedora, we do not provide JupyterLab as an RPM package - we only have Jupyter Notebook. And Jupyter Notebook does not support debuggers yet so trying to package and maintain debugpy just to be able to update ipykernel when our users have no benefit from the debugger implementation does not seem to be the best way how to spend my time.
Right now, I can either keep the latest version where debugpy is not required or try to disable the debugger implementation in ipykernel.

Should I try to prepare a PR making the dependency on debugpy optional?

martinRenou:
It might be simpler and safer to use pip, or even better miniconda, to install your Python packages, instead of dnf.

That way you don't mess with your system. You could even start using JupyterLab :)

hroncok:
Some of our Fedora packages (integrated into the distro) depend on ipykernel. Using pip or miniconda is not an option. Lumír has provided some feedback about why is this new dependency on DebugPy undesired and problematic for us and saying "just use pip/miniconda" does not bring any value to this discussion.

martinRenou:
Is the issue with Python 3.10 the only blocker? It might be worth just waiting for debugpy to be released with Python 3.10 support?

[...]

PR #767, opened 2021-09-07
Make ipykernel work without debugpy

frenzymadness:
debugpy is an optional dependency because only frontends with
debugging support (Jupyter lab) can really use its features.

Fixes: https://github.com/ipython/ipykernel/issues/712

There is also one new test in the CI to check that all tests are passing when debugpy is not installed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,3 +110,44 @@ jobs:
     - name: Check Docstrings
       run: |
         velin . --check --compact
+  test_without_debugpy:
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu]
+        python-version: [ '3.9' ]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+    - name: Install Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: 'x64'
+    - name: Upgrade packaging dependencies
+      run: |
+        pip install --upgrade pip setuptools wheel --user
+    - name: Get pip cache dir
+      id: pip-cache
+      run: |
+        echo "::set-output name=dir::$(pip cache dir)"
+    - name: Cache pip
+      uses: actions/cache@v1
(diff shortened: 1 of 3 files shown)


blink1073 (approved):
LGTM, thank you!  I'll leave this open for a day in case another maintainer has input

frenzymadness:
I'm testing this also in Fedora where 25 out of 28 packages built fine with the patched version without debugpy. The three packages fail to build for different reasons.

minrk:
Thanks!

fcollonval:
Should `debugpy` be removed from install requirements following this change?

frenzymadness:
That's a good question for maintainers. I can imagine that it is an extra dependency and then jupyter notebook can depend on ipykernel without debugby and jupyterlab can depend on something like ipykernel[debug] because as far as I understand it, Jupyter lab is the only frontend with implemented debugging.

martinRenou:
This approach might not work in the current state of things.

JupyterLab does not depend on ipykernel directly, it depends on `nbclassic` which itself depends on `notebook` which depends on ipykernel.
Also this king of installation "flavors" has no counterpart in conda which seems to be the preferred way of installing the jupyter packages in the ecosystem.

I would also like to say that jupyterlab (and notebook) should ideally not depend on ipykernel, it's a language agnostic tool after all.

minrk:
> Should debugpy be removed from install requirements following this change?

> In particular if debugpy is kept a mandatory dependency, it is considered as bad practice to have a set of packages breaking pip check.

No, it shouldn't be removed in this repo. If a downstream packager wants to split this dependency, it may, but at the same time, it should also make sure to patch the package metadata when doing so to remove the dependency and satisfy `pip check`.

frenzymadness:
> > Should debugpy be removed from install requirements following this change?
> 
> > In particular if debugpy is kept a mandatory dependency, it is considered as bad practice to have a set of packages breaking pip check.
> 
> No, it shouldn't be removed in this repo. If a downstream packager wants to split this dependency, it may, but at the same time, it should also make sure to patch the package metadata when doing so to remove the dependency and satisfy `pip check`.

And that's exactly what we do in Fedora. Except `pip check` does not make sense for the system site-packages.

PR #773, opened 2021-09-10
debugpy is now a build requirement

### LINKED ISSUE #770: Debugger always disabled with ipykernel 6.4.0?
Using the latest `6.4.0` version of `ipykernel`, it looks like the debugger is always disabled.

This was first noticed on Binder using this gist: https://gist.github.com/jtpio/6ce26381703355e0ef1da4af742b7f72

But can be reproduced locally:

```bash
mamba create -n tmp python -y
conda activate tmp

# install the latest versions
python -m pip install -U ipykernel debugpy jupyterlab

# start JupyterLab
jupyter lab
```

The spec seems to be installed with `debugger: false` on disk:

```
$ cat ${CONDA_PREFIX}/share/jupyter/kernels/python3/kernel.json
{
 "argv": [
  "python",
  "-m",
  "ipykernel_launcher",
  "-f",
  "{connection_file}"
 ],
 "display_name": "Python 3 (ipykernel)",
 "language": "python",
 "metadata": {
  "debugger": false
 }
}
```

And the kernelspecs return `debugger: false` for `ipykernel` at startup:

image

Dependencies:

```bash
$ pip list | grep -E "debugpy|ipykernel"
debugpy             1.4.1
ipykernel           6.4.0
```

minrk:
to get the right debugger metadata in the wheel kernelspec

due to the runtime check added [here](https://github.com/ipython/ipykernel/pull/767/files#diff-c4f6702e977e08e5d1aee47d489ef372ff37913e7e282dbaf269d050f58d56a7R57). The wheel kernelspec currently has an inaccurate `debugger: false` in its metadata, so should probably do a 6.4.1

closes #770

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,7 @@ build-backend = "setuptools.build_meta"
 requires=[
   "setuptools",
   "wheel",
+  "debugpy",
   "ipython>=5",
   "jupyter_core>=4.2",
   "jupyter_client",


blink1073 (approved):
Thanks!  I'll cut a release with this change

PR #778, opened 2021-10-06
Add `ipython_genutils` dependency to 5.5.x branch.

carlosperate:
https://github.com/ipython/ipykernel/issues/759

This was an implicit dependency via Traitlets, but it was removed in version 5.1 and needs to be explicitly included now.


blink1073:
Looks like we need to add it to build reqs as well: https://github.com/ipython/ipykernel/blob/b7736840272b66de15f2f9301205d257286cc9e8/pyproject.toml#L3

carlosperate:
Oh wait, I didn't realised that was a `pyproject.toml` file, you are probably right and that could be enough!

PR #796, opened 2021-11-08
Unpin IPython, and remove some dependencies on it.

Carreau:
This makes it annoying to test whether IPython breaks IPykernel.
I'm almost always on IPython master branch and I'd like to know when/if
I break things.

Prompted by https://github.com/ipython/ipython/pull/13252,
which is not breaking ipykernel, but it's better to decouple where we
can.

### DIFF
--- a/ipykernel/inprocess/tests/test_kernel.py
+++ b/ipykernel/inprocess/tests/test_kernel.py
@@ -13,8 +13,8 @@ from ipykernel.inprocess.blocking import BlockingInProcessKernelClient
 from ipykernel.inprocess.manager import InProcessKernelManager
 from ipykernel.inprocess.ipkernel import InProcessKernel
 from ipykernel.tests.utils import assemble_output
-from IPython.testing.decorators import skipif_not_matplotlib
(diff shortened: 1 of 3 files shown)


blink1073:
We had added the pin as a failsafe in case there are breaking API changes, I'd feel better about removing it if `ipython` had a downstream test on [`ipykernel`](https://github.com/jupyter/jupyter_client/blob/f453b51eeeff9e905c583b7da3905c0e35cfbdf0/.github/workflows/downstream.yml#L27).

Carreau:
Thanks for the pointer, the pyargs and --upgrade-strategy=only-if-needed are useful. 

Should I test both master an released or just released ?

blink1073:
I'd say just master, since then we can update ipykernel prior to the release to be compatible.

Carreau:
I just merged https://github.com/ipython/ipython/pull/13266 that test upstream master.

blink1073:
Nice, thanks!

blink1073 (approved):

PR #801, opened 2021-11-17
fix #800 the temp file name generated by murmur2 in python and ts is not same

### LINKED ISSUE #800: The temp file name generated by murmur2 in python and ts is not the same
This leads to not able to create a breakpoint when there is Chinese or other language in the notebook

I clicked the line in 
`print("中文")`
but not able to see it.

image

eastonsuo:
fix #800


blink1073 (approved):
Thank you!

PR #805, opened 2021-11-20
Add usage_request and usage_reply based on psutil

echarles:
This PR aims to start a discussion to add a new message type usage_request/usage_reply as discussed in the last JupyterLab meeting https://github.com/jupyterlab/team-compass/issues/128#issuecomment-971829367

On JupyterLab https://github.com/jupyterlab/jupyterlab/pull/11285 is needed.

cc/ @Carreau

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -872,10 +872,7 @@ class Kernel(SingletonConfigurable):
         # The first time cpu_percent is called it will return a meaningless 0.0 value which you are supposed to ignore.
         if cpu_percent != None and cpu_percent != 0.0:
             reply_content['cpu_percent'] = cpu_percent
-        reply_content['virtual_memory'] = psutil.virtual_memory()
-        reply_content['virtual_memory_dict'] = dict(psutil.virtual_memory()._asdict())
-        reply_content['virtual_memory_percent'] = psutil.virtual_memory().percent
-        reply_content['available_memory_percentage'] = psutil.virtual_memory().available * 100 / psutil.virtual_memory().total
+        reply_content['virtual_memory'] = dict(psutil.virtual_memory()._asdict())
         reply_msg = self.session.send(stream, 'usage_reply', reply_content,
                                       parent, ident)
         self.log.debug("%s", reply_msg)
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -16,6 +16,7 @@ import sys
 import time
 import uuid
 import warnings
+import psutil
 
 try:
     # jupyter_client >= 5, use tz-aware now
@@ -204,7 +205,7 @@ class Kernel(SingletonConfigurable):
         'apply_request',
     ]
     # add deprecated ipyparallel control messages
-    control_msg_types = msg_types + ['clear_request', 'abort_request', 'debug_request']
+    control_msg_types = msg_types + ['clear_request', 'abort_request', 'debug_request', 'usage_request']
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -856,7 +857,23 @@ class Kernel(SingletonConfigurable):
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
         reply_content = json_clean(reply_content)
+        reply_content['hello'] = 'hello'
         reply_msg = self.session.send(stream, 'debug_reply', reply_content,
+                                      parent, ident)
+        self.log.debug("%s", reply_msg)
+
+    async def usage_request(self, stream, ident, parent):
+        content = parent['content']
+        reply_content = self.do_debug_request(content)
+        if inspect.isawaitable(reply_content):
+            reply_content = await reply_content
+        reply_content = json_clean(reply_content)
+        reply_content['cpu_percent'] = psutil.cpu_percent()
+        reply_content['virtual_memory'] = psutil.virtual_memory()
+        reply_content['virtual_memory_dict'] = dict(psutil.virtual_memory()._asdict())
+        reply_content['virtual_memory_percent'] = psutil.virtual_memory().percent
+        reply_content['available_memory_percentage'] = psutil.virtual_memory().available * 100 / psutil.virtual_memory().total
+        reply_msg = self.session.send(stream, 'usage_reply', reply_content,
                                       parent, ident)
         self.log.debug("%s", reply_msg)
 
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -862,11 +862,7 @@ class Kernel(SingletonConfigurable):
         self.log.debug("%s", reply_msg)
 
     async def usage_request(self, stream, ident, parent):
-        content = parent['content']
-        reply_content = self.do_debug_request(content)
-        if inspect.isawaitable(reply_content):
-            reply_content = await reply_content
-        reply_content = json_clean(reply_content)
+        reply_content = {}
         cpu_percent = psutil.cpu_percent()
         # https://psutil.readthedocs.io/en/latest/index.html?highlight=cpu#psutil.cpu_percent
         # The first time cpu_percent is called it will return a meaningless 0.0 value which you are supposed to ignore.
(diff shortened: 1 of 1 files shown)


ipykernel/kernelbase.py
```diff
@@ -855,10 +856,26 @@ async def debug_request(self, stream, ident, parent):
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
         reply_content = json_clean(reply_content)
+        reply_content['hello'] = 'hello'
```
mlucool, line 859: Was this left in by accident?

ipykernel/kernelbase.py
```diff
@@ -855,10 +856,26 @@ async def debug_request(self, stream, ident, parent):
...
                                       parent, ident)
         self.log.debug("%s", reply_msg)
 
+    async def usage_request(self, stream, ident, parent):
+        content = parent['content']
+        reply_content = self.do_debug_request(content)
+        if inspect.isawaitable(reply_content):
+            reply_content = await reply_content
+        reply_content = json_clean(reply_content)
+        reply_content['cpu_percent'] = psutil.cpu_percent()
```
mlucool, line 870: Should we ignore 0 values? By doing this we'd keep this part of the API hidden from the front ends who may think 0 is a real value.

https://github.com/ipython/ipykernel/pull/805/files

> Warning the first time this function is called with interval = 0.0 or None it will return a meaningless 0.0 value which you are supposed to ignore.

ipykernel/kernelbase.py
```diff
@@ -855,10 +856,26 @@ async def debug_request(self, stream, ident, parent):
...
+        reply_content['virtual_memory'] = psutil.virtual_memory()
```
mlucool, line 871: It seems all of `virtual_memory` `virtual_memory_dict`, `virtual_memory_percent` and `available_memory_percentage` can be derived by just using `virtual_memory`.

Proposal - only send:
```
reply_content['virtual_memory'] = dict(psutil.virtual_memory()._asdict())
```

ipykernel/kernelbase.py
echarles: Thx. 0c8fcba479a60d299cfea3489300beb0217b4cd9 removes this.

ipykernel/kernelbase.py
echarles: Thx. Fixed with c574a01a344656e0ac915e88dcdc54a9543b92ac

ipykernel/kernelbase.py
echarles: Thx. Fixed with 889481b15413755f1cebfe433a78fead6559f67f

echarles:
A JupyterLab extension example/experiment using this features is available on https://github.com/datalayer-experiments/jupyter-usage

echarles:
@Carreau This is ready for review

ipykernel/kernelbase.py
```diff
@@ -859,6 +860,22 @@ async def debug_request(self, stream, ident, parent):
                                       parent, ident)
         self.log.debug("%s", reply_msg)
 
+    async def usage_request(self, stream, ident, parent):
+        content = parent['content']
+        reply_content = self.do_debug_request(content)
```
JohanMabille, line 865: why calling this method specific do `debug_request` here?

echarles: Not needed. Latest commit removes it. Thx @JohanMabille

ipykernel/kernelbase.py
```diff
@@ -204,7 +208,7 @@ def _parent_header(self):
         'apply_request',
     ]
     # add deprecated ipyparallel control messages
-    control_msg_types = msg_types + ['clear_request', 'abort_request', 'debug_request']
+    control_msg_types = msg_types + ['clear_request', 'abort_request', 'debug_request', 'usage_request']
```
Carreau, line 211: at some point I think this should be a `Enum`, not a request to do that here and now. Maybe just a comment for later.

ipykernel/kernelbase.py
```diff
@@ -859,6 +863,41 @@ async def debug_request(self, stream, ident, parent):
...
+
+    async def usage_request(self, stream, ident, parent):
+        reply_content = {}
+        if psutil is None:
+            return reply_content
+        current_process = psutil.Process()
+        all_processes = [current_process] + current_process.children(recursive=True)
+        process_metric_value = lambda process, name, attribute: self.get_process_metric_value(
+                process, name, attribute
+        )
```
Carreau, line 888: is the lambda necessary, can't you just pass `self.get_process_metric_value` below ?

ipykernel/kernelbase.py
```diff
@@ -859,6 +863,41 @@ async def debug_request(self, stream, ident, parent):
...
+            # psutil.Process methods will either return...
+            metric_value = getattr(process, name)()
+            if attribute is not None:  # ... a named tuple
+                return getattr(metric_value, attribute)
+            else:  # ... or a number
+                return metric_value
+        # Avoid littering logs with stack traces
+        # complaining about dead processes
+        except BaseException:
+            return 0
```
Carreau, line 878: Do you wan to use something "None" or do you want process metrics to always return a number ?
I don't particularly care, just wondering if you want to a way to propagate that it's not actually 0, but a no-data ?

ipykernel/kernelbase.py
echarles: That lambda function is used after by `sum([process_metric_value(process, 'cpu_percent', None) for process in all_processes])` to sum the metric values on all the processes.

As it is written now, it is a sugar and is reused twice. I could rewrite that code block, it is just a matter of code styling I think.

ipykernel/kernelbase.py
echarles: That lambda function is used after by sum([process_metric_value(process, 'cpu_percent', None) for process in all_processes]) to sum the metric values on all the processes.

As it is written now, it is a sugar and is reused twice. I could rewrite that code block, it is just a matter of code styling I think.

ipykernel/kernelbase.py
echarles: Going for None. Thx.

ipykernel/kernelbase.py
echarles: I have opened https://github.com/ipython/ipykernel/issues/833 as follow-up

ipykernel/kernelbase.py
Carreau: Ok, sorry if what i'm saying is unclear. 

What's the difference with:

```diff
- process_metric_value = lambda process, name, attribute: self.get_process_metric_value(
                process, name, attribute
        )
+ process_metric_value = self.get_process_metric_value
```

You don't capture any values, you don't cache, and don't change  arguments order.

echarles: oh, true. It was clear but I have not seen the obvious simplification. Thx for helping here :) That change has been committed and works fine.

Carreau: I think there is a lambda calcul term for that is is a β-reduction ?

https://en.wikipedia.org/wiki/Lambda_calculus#%CE%B2-reduction

echarles: > I think there is a lambda calcul term for that is is a β-reduction ?

I trust you on that :)

Carreau:
Rebased on master to see if it fixes the test failures. Those should have been fixed on main branch.

echarles:
Super! Thx @Carreau

jasongrout:
For clarification, for those reading this, `usage_request` and `usage_reply` are non-standard kernel messages right now (i.e., they are not in the kernel message spec, so they are specific implementation details of ipykernel).

jasongrout:
The related JupyterLab PR, https://github.com/jupyterlab/jupyterlab/pull/11285, is now closed:

> I will close this one as we are not going to go further with this specific implementation. A potential more generic follow-up can be discussed in https://github.com/jupyterlab/jupyterlab/issues/12117

If this message is not going to be used, should we revert this PR? @echarles?

echarles:
The usage request/reply is consumed by https://github.com/Quansight/jupyterlab-kernel-usage

PR #807, opened 2021-11-22
Add explicit encoding to open calls in debugger

dlukes:
This should help get rid of chatty EncodingWarnings for users on 3.10 with PYTHONWARNDEFAULTENCODING enabled (such as myself) :)


blink1073:
cc @minrk, it looks like adding encoding affects `ipyparallel`.

blink1073:
Thanks for the contribution @dlukes!

minrk:
This didn't affect ipyparallel, there just appears to be a flaky test in the latest release.

PR #809, opened 2021-11-27
Send omit_sections to IPython to choose which sections of documentation you don't want

fasiha:
This PR brings to ipykernel a recent IPython PR https://github.com/ipython/ipython/pull/13343 which allows us to customize IPython's responses to `info` and `inspect` messages by specifying which documentation sections we *don't* want. For example, `numpy.sin` returns the following sections:
- Call signature
- Type
- String form
- File
- Docstring
- Class docstring

and with the new `omit_sections` keyword argument, we can for example specify that we are not interested in the `File`, `String form`, and `Type` sections, thereby saving runtime and bandwidth by not generating those documentation sections.

In this present PR, we enable ipykernel's `inspect` message to take advantage of `omit_sections`. If the client provides `omit_sections` in its request, we'll pass it on to IPython.

**Nota bene** if merged, this PR should be released with an ipykernel that depends on IPython 8.0, the milestone @Carreau has set in https://github.com/ipython/ipython/pull/13343.

---

To test this using JupyterLab for example:
1. install ipython master
2. install this PR's branch of ipykernel (https://github.com/fasiha/ipykernel/tree/blocklist-for-inspector)
3. install https://github.com/fasiha/jupyterlab/tree/use-omit_sections branch of jupyterlab and start it: this branch has [configured](https://github.com/fasiha/jupyterlab/blob/6c20cefe215f28f54ad8d9b1405c544773fce2aa/packages/inspector/src/kernelconnector.ts#L45) its inspector to omit all sections except docstrings
4. run the following in a notebook cell:
```py
import numpy
numpy.sin
numpy.exp
```
5. open JupyterLab's contextual help menu and click on "exp" and "sin", etc.

Original ipykernel:

Screen Shot 2021-11-26 at 7 19 14 PM

This PR with customized JupyterLab to omit numerous sections:

Screen Shot 2021-11-26 at 7 19 48 PM

---

Recommended triage: `enhancement`?

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -5,7 +5,6 @@ import builtins
 from contextlib import contextmanager
 from functools import partial
 import getpass
-import re
 import signal
 import sys
 
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -459,7 +459,7 @@ class IPythonKernel(KernelBase):
                 'metadata': {_EXPERIMENTAL_KEY_NAME: comps},
                 'status': 'ok'}
 
-    def do_inspect(self, code, cursor_pos, detail_level=0):
+    def do_inspect(self, code, cursor_pos, detail_level=0, omit_sections=()):
         name = token_at_cursor(code, cursor_pos)
 
         reply_content = {'status' : 'ok'}
@@ -469,7 +469,8 @@ class IPythonKernel(KernelBase):
             reply_content['data'].update(
                 self.shell.object_inspect_mime(
                     name,
-                    detail_level=detail_level
+                    detail_level=detail_level,
+                    omit_sections=omit_sections,
                 )
             )
             if not self.shell.enable_html_pager:
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -5,6 +5,7 @@ import builtins
 from contextlib import contextmanager
 from functools import partial
 import getpass
+import re
 import signal
 import sys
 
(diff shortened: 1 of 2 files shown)


ipykernel/ipkernel.py
```diff
@@ -469,7 +469,8 @@ def do_inspect(self, code, cursor_pos, detail_level=0):
             reply_content['data'].update(
                 self.shell.object_inspect_mime(
                     name,
-                    detail_level=detail_level
+                    detail_level=detail_level,
+                    omit_sections=omit_sections,
```
blink1073, line 473: If we guard this with a `TypeError` catch and then retry without `omit_sections` then we maintain backwards compatibility.

Carreau: You can also be explicit and check against `IPython.version_info` tuple, this will make it obvious which codepath to cleanup when we stop supporting IPython 7.x. If we use try/except we may catch too many error and it will not be obvious in 2/3 years what's happening here.

fasiha: @Carreau and @blink1073 thanks for the tips!!! I have prepared a commit to look at `version_info`, and will push it shortly. A small question: if we do it that way, ipykernel master to not use this until IPython v8 is released—are we ok with that? (Meanwhile the `TypeError` approach would let both modules interoperate in master.)

fasiha: Pushed 0bdfd0bf9ae8e1dd61802ac90f16a9d8d4ed74af, very much open to your feedback 🙇🙏!

Carreau: The version number of master is already 8.0.0.dev,  so it will (should) take this branch, and I'm likely to release 8.0 end of december/early 2022. The only case where this would fail is if someone is using and older commit of IPython master branch, and ipykernel after this is merged. 

In that case I believe the person is likely a developer and would now how to fix it/update their repositories.

fasiha: Ahh of course, makes total sense! Sorry for being slow 😅, I realize now I was looking at the Conda-installed copy of IPython's `release.py`, not master. I pushed this improvement, thanks so much for your patience and very kind assistance 🙏!

Carreau: No problem, we don't expect you to be aware of all the informations and when what is released and what is our future plan.

Carreau (approved):

blink1073:
Thanks again @fasiha!

PR #814, opened 2021-12-01
Added missing excpeionPaths field to debugInfo reply

JohanMabille:
This field is now mandatory in the [Jupyter Kernel Protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html#additions-to-the-dap) and is required in order to move https://github.com/jupyterlab/jupyterlab/pull/10298 forward.


blink1073 (approved):
Thanks!

PR #816, opened 2021-12-01
Add support for the debug modules request

echarles:
This PR adds the support for the Modules request https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Modules as discussed on https://github.com/jupyterlab/jupyterlab/pull/11566#discussion_r759661311

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -594,15 +594,16 @@ class Debugger:
         return reply
 
     async def modules(self, message):
-        modules = sys.modules.values()
+        modules = list(sys.modules.values())
+        startModule = message.get('startModule', 0)
+        moduleCount = message.get('moduleCount', len(modules))
         mods = []
-        for module in modules:
+        for i in range(startModule, moduleCount):
+            module = modules[i]
             filename = getattr(getattr(module, "__spec__", None), "origin", None)
             if filename and filename.endswith(".py"):
                 mods.append({module.__name__: filename})
 
-        reply = {"body": {"modules": mods, "totalModules": len(modules)}}
-        return reply
         reply = {"body": {"modules": mods, "totalModules": len(modules)}}
         return reply
 
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -1,3 +1,4 @@
+import sys
 import os
 import re
 import threading
@@ -265,7 +266,8 @@ class Debugger:
 
     # Requests that can be handled even if the debugger is not running
     static_debug_msg_types = [
-        'debugInfo', 'inspectVariables', 'richInspectVariables'
+        'debugInfo', 'inspectVariables', 
+        'richInspectVariables', 'modules'
     ]
 
     def __init__(self, log, debugpy_stream, event_callback, shell_socket, session):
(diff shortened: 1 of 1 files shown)


ipykernel/debugger.py
```diff
@@ -562,6 +564,24 @@ async def richInspectVariables(self, message):
...
+            m = str(module)
+            if ".py'" in m:
+                a = {}
+                x = re.findall(r"'(.*?)'", m)
+                a[x[0]] = x[1]
+                mods.append(a)
+        reply = {
+            'body': {
+                'modules': mods,
+                'totalModules': len(modules)
```
JohanMabille, line 580: I think we the debugger should handle the `startModule` and `moduleCount` arguments speficied in the [DAP](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Modules) and filter the result accordingly.

JohanMabille (changes requested):

ipykernel/debugger.py
```diff
@@ -562,6 +564,24 @@ async def richInspectVariables(self, message):
         reply["success"] = True
         return reply
 
+    async def modules(self, message):        
+        modules = [sys.modules[name] for name in sys.modules]
```
Carreau, line 568: Any reasons not to use `sys.modules.values()` here ?

ipykernel/debugger.py
```diff
@@ -562,6 +564,24 @@ async def richInspectVariables(self, message):
...
+        mods = []
+        for module in modules:
+            m = str(module)
```
Carreau, line 571: Are you looking for `mod.__spec__.origin`  ? (`__spec__` may be non so you would need `getattr(..., '__spec__', None)`, so not sure it is easier.

Carreau (approved):
No objections from me.

ipykernel/debugger.py
Carreau: I came  up with the following which is a bit less regexp-y:

```python
    async def modules(self, message):        
        modules = sys.modules.values()
        mods = []
        for module in modules:
            filename = getattr(getattr(module, "__spec__", None), "origin", None)
            if filename and filename.endswith(".py"):
                mods.append({mod.__name__: filename})

        reply = {"body": {"modules": mods, "totalModules": len(modules)}}
        return reply
```

One thing which is not clear to me is why would `mod.__name__` be different than the key in `sys.modules`, and is that roughly trying to do:

```
mods = [{k:m.__spec__.origin} for k,m in sys.modules.items()]
```

ipykernel/debugger.py
```diff
@@ -263,7 +264,8 @@ class Debugger:
 
     # Requests that can be handled even if the debugger is not running
     static_debug_msg_types = [
-        'debugInfo', 'inspectVariables', 'richInspectVariables'
+        'debugInfo', 'inspectVariables', 
+        'richInspectVariables', 'modules'
     ]
```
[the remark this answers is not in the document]
echarles: Tracked in https://github.com/ipython/ipykernel/issues/833

ipykernel/debugger.py
echarles: sys.modules.values() works. thx.

ipykernel/debugger.py
echarles: I have implemented your suggestion in 8777e68e56712fef254ea72c70dca147b6e56a6d

Not sure how `sys.modules` should behave. I came to that code block in an empirical way.

ipykernel/debugger.py
echarles: I have implemented the `startModule` and `moduleCount` in 95711b80b47537927bb2359daf43ea39cf2f0ad2

echarles:
Current comments have been addressed.
I have rebased on `main` and undrafted.
CI is green.

This is ready for review/merge.

JohanMabille:
LGTM

echarles:
Thx @JohanMabille and @Carreau for the review and merge.

Issue #763, opened 2021-09-03
Call to process_one(wait=False) raises exception when there is nothing to process


Kirill888:
I think code here is trying to catch wrong exception:

https://github.com/ipython/ipykernel/blob/3c6037fbed4431a6997414918bfc6a65b13070e7/ipykernel/kernelbase.py#L442-L445

My guess is that `tornado.queues.QueueEmpty` was the intended exception, and not `asyncio.QueueEmpty`

blink1073:
Fixed in #817

PR #817, opened 2021-12-02
RF: Handle both asyncio.QueueEmpty and tornado.queues.QueueEmpty

pauldmccarthy:
Howdy, I just came across the error reported in #763. I believe that the fix is trivial. My guess is that using an ioloop from either `tornado` or `ioloop` is possible (or is a future possibility), so both exception types should be catered for.

Thanks!


blink1073 (approved):
Thanks!

Issue #799, opened 2021-11-17
<debug_request> publish the <status:idle> while an <execute_request> is running


mizililo:
In ipykernel 5.5.6，both <control_stream> and <shell_stream> put msgs into <msg_queue> and process the msg with <schedule_dispatch>. Although both <dispatch_control> and <dispatch_shell> publish the <status: busy/idle>，there are no conflicts.

When it comes to ipykernel 6.5.0, <control_stream> put msgs into <control_queue> and process the msg with <dispatch_control>, while shell_stream still works in the old way. This means <dispatch_control> and <dispatch_shell> will publish the <status: busy/idle> in the same time.

I found this problem in jupyterlab when i executed a cell(sleep 600s) in an ipykernel with <metadata: {debugger: True}>，i switched to another ipynb and switched back. The cell was still running, but the kernel execution_state became idle. Then i found the ipykernel received multi <debug_requests> and published the status to idle.

As https://jupyter-client.readthedocs.io/en/latest/messaging.html said "Debug requests and replies are sent over the control channel to prevent queuing behind execution requests". It confuses me that why control_requests affect the kernel's execution_state? Should we use another state like control_state?

minrk:
Until the debugger, all requests on both control and shell were handled in the say way - where one request could be processed at a time, and thus the kernel was always 'busy' or 'idle', regardless of which queue the task came from. Control being a separate queue does _not_ mean that control requests don't block in the same way that shell requests do, nor that they are concurrent. Just that they jump to the head of the line when the kernel is ready to handle the next request, whichever queue it may come from. However, debug state being concurrent does complicate things. I'm not quite sure what the right answer is there.

mizililo:
> Until the debugger, all requests on both control and shell were handled in the same way.
In previous ipykernel like 5.5.6, it's true and all requests are put into <msg_queue> by <schedule_dispatch> and processed in order by <dispatch_queue> -> <process_one>
While in latest version, msgs from <crontrol_stream> are put into <control_queue> by <dispatch_control> and processed by <poll_control_queue> -> <poll_control_queue>. This Actually Make shell_request / control_request Work Concurrently.
I think the pollers(<poll_control_queue> and <dispatch_queue>) do cause the bug.

mizililo:
In short, previous versions use one poller to get msg from msg_queue, while the latest version uses two pollers to get msg from two queues, and this causes two dispatchers to process and publish the status concurrently

JohanMabille:
[...]

PR #819, opened 2021-12-03
fix ipykernel #799 by removing publish_status in process_control 

mizililo:
because control_channel shouldn't publish execution_state

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -266,7 +266,6 @@ class Kernel(SingletonConfigurable):
 
         # Set the parent message for side effects.
         self.set_parent(idents, msg, channel='control')
-        self._publish_status('busy', 'control')
 
         header = msg['header']
         msg_type = header['msg_type']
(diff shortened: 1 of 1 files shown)


JohanMabille:
I think we should wait before merging this. I am currently writing a JEP to move the `kernel_info_request` to the control channel to fix https://github.com/jupyterlab/jupyterlab/issues/10174, and the jupyter server expects both a reply to the `kernel_info_reply` and a message on IOPub to cosider it has connected to the kernel. Merging this PR would break the new behavior.

Notice there is a on-going [JEP](https://github.com/jupyter/enhancement-proposals/pull/65) to replace PUB socket with XPUB socket, so that clients get an iopub message upon connection and they don't have to nudge the kernel anymore.

mizililo:
OK

Carreau:
I think according to the JEPwe can close this; plus it's been 5 years, feel free to start with a new PR if this is in correct.

PR #822, opened 2021-12-09
IPython's start_kernel is deprecated.

Carreau:
Technically it has been deprecated since 5.0 but indirectly as it
imports from IPython.kernel with itself is a shim module that exposes
ipykernel.

So it is useless to call into IPython if it's for calling back into
ipykernel.

Though the start_kernel itself was not emitting a deprecation warning,
which will be the case in 8.0, and hopefully we can finish removing the
shim modules later.

### DIFF
--- a/ipykernel/tests/test_start_kernel.py
+++ b/ipykernel/tests/test_start_kernel.py
@@ -1,14 +1,19 @@
 from .test_embed_kernel import setup_kernel
 from flaky import flaky
+from textwrap import dedent
 
 TIMEOUT = 15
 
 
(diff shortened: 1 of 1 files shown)


Carreau:
Note: I will likely remove IPython.kernel in 8.0 (ipython/ipython#13386), but will leave and update `IPython.start_kernel` with a deprecation warning in 8.0, to directly refer to ipykernel.

blink1073 (approved):
Thanks!

Carreau:
Thanks !!

PR #827, opened 2021-12-22
Clean python 2 artifacts. Fix #826

### LINKED ISSUE #826: Python 2.x artifacts present in code
`super(..., self).` instead of `super()`
subclassing of `object`
`IOError` instead of `OSError`

penguinolog:
* `IOError` is `OSError` in python3
* `io.open` -> `open`
* `return str()` -> `return ''` directly
* direct `super()` call
* do not subclass `object`

### DIFF
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -375,15 +375,14 @@ class KernelMagics(Magics):
 
 
         print (info + '\n')
-        print ("Paste the above JSON into a file, and connect with:\n"
-            "    $> jupyter <app> --existing <file>\n"
-            "or, if you are local, you can connect with just:\n"
+        print (
+            f"Paste the above JSON into a file, and connect with:\n"
+            f"    $> jupyter <app> --existing <file>\n"
+            f"or, if you are local, you can connect with just:\n"
             f"    $> jupyter <app> --existing {connection_file}\n"
-            "or even just:\n"
-            "    $> jupyter <app> --existing\n"
-            "if this is the most recent Jupyter kernel you have started.".format(
-            connection_file
-            )
+            f"or even just:\n"
+            f"    $> jupyter <app> --existing\n"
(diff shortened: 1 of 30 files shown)


ipykernel/zmqshell.py
```diff
@@ -378,7 +378,7 @@ def connect_info(self, arg_s):
         print ("Paste the above JSON into a file, and connect with:\n"
             "    $> jupyter <app> --existing <file>\n"
             "or, if you are local, you can connect with just:\n"
-            "    $> jupyter <app> --existing {0}\n"
+            "    $> jupyter <app> --existing {}\n"
```
davidbrochart, line 381: Maybe replace `.format(connection_file)` with:
```suggestion
            f"    $> jupyter <app> --existing {connection_file}\n"
```

davidbrochart (approved):
You also made the following changes:
- replace `.format()` with `f-strings`
- replace `set([1, 2])` with `{1, 2}`

ipykernel/zmqshell.py
```diff
@@ -378,7 +378,7 @@ def connect_info(self, arg_s):
         print ("Paste the above JSON into a file, and connect with:\n"
             "    $> jupyter <app> --existing <file>\n"
             "or, if you are local, you can connect with just:\n"
-            "    $> jupyter <app> --existing {0}\n"
+            f"    $> jupyter <app> --existing {connection_file}\n"
             "or even just:\n"
             "    $> jupyter <app> --existing\n"
             "if this is the most recent Jupyter kernel you have started.".format(
```
davidbrochart, line 384: You also need to remove `.format(connection_file)`.

penguinolog: it was removed, like almost all other format calls in next commit

davidbrochart: Thanks!

PR #829, opened 2021-12-29
Breakpoint in cell with leading empty lines may be ignored

fcollonval:
If a cell starts with empty lines, they are cleanup as part of the code transformation. This result in breakpoint that can be missed as the number of lines executed will be smaller than the breakpoint line number.

For example,

```python

def f(a, b):
    return a + b
f(2, 3)
```

Setting the breakpoint on the last line is not working

Xref: https://github.com/jupyterlab/jupyterlab/issues/11532

This removes the `leading_empty_lines` cleanup transform when a debug session is active.

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -8,6 +8,7 @@ from tornado.queues import Queue
 from tornado.locks import Event
 
 from IPython.core.getipython import get_ipython
+from IPython.core.inputtransformer2 import leading_empty_lines
 
 try:
     from jupyter_client.jsonutil import json_default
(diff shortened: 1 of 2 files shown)


fcollonval:
pinging @Carreau could you please have a look at this PR? I'm not very confident to turn off one of the default transformation applied by IPython.

fcollonval:
The error on Python 3.10 seems legit.

fcollonval:
Actually the debugger with Python 3.10 is much more broken than with Python 3.9. But I can't yet figure out what is going on.

fcollonval:
I propose to skip the new test for Python 3.10 for now. And tackle the issue related to Python 3.10 in a separate PR.

Carreau:
I have no objection to have this as a workaround, we can make the transformer have an enabled/disabled flag in IPython itself to make it less brittle to index positions.

fcollonval:
Thanks @Carreau

This is ready for review.

blink1073 (approved):
Thanks!

PR #838, opened 2022-01-07
Normalize debugger temp file paths on Windows

### LINKED ISSUE #784: Temp file paths can be inconsistent with debugpy
The user directory on my Windows machine is `C:\Users\kycutler.NORTHAMERICA`. However the temporary file location returned by `tempfile.gettempdir()` is shortened to `C:\Users\KYCUTL~1.NOR\AppData\Local\Temp`. Thus the shortened version is used in `tmpFilePrefix` and results from `dumpCell`:

image

image

However, `debugpy` [converts file paths to their long form](https://github.com/microsoft/debugpy/blob/869babbd9370cd5b5a4f4683f6d5460af33906c8/src/debugpy/_vendored/pydevd/pydevd_file_utils.py#L271), so the temp file paths are different:

image

This causes issues in the frontend since we can not simply check for equality between the paths returned from debugpy and those returned from `dumpCell` or constructed using `tmpFilePrefix`.

kycutler:
Fixes #784

### DIFF
--- a/ipykernel/compiler.py
+++ b/ipykernel/compiler.py
@@ -1,6 +1,7 @@
 from IPython.core.compilerop import CachingCompiler
 import tempfile
 import os
+import sys
 
 
 def murmur2_x86(data, seed):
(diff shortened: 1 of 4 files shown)


blink1073 (commented):
Rather than adding a dependency, we could use a `ctypes` version of this function call: https://mail.python.org/pipermail/python-win32/2008-January/006642.html

kycutler:
@blink1073 thanks for the review! Updated to use ctypes and remove the direct dependency on pywin32. The ctypes code is loosely based on the similar handling in debugpy [here](https://github.com/microsoft/debugpy/blob/869babbd9370cd5b5a4f4683f6d5460af33906c8/src/debugpy/_vendored/pydevd/pydevd_file_utils.py#L139).

ipykernel/compiler.py
```diff
@@ -36,11 +37,35 @@ def murmur2_x86(data, seed):
...
+
+        def _convert_to_long_pathname(filename):
+            buf = ctypes.create_unicode_buffer(MAX_PATH)
+            rv = _GetLongPathName(filename, buf, MAX_PATH)
+            if rv != 0 and rv <= MAX_PATH:
+                filename = buf.value
+            return filename
+
+        # test that it works
+        _convert_to_long_pathname(__file__)
```
blink1073, line 59: Can you please add this as a test in `test_debugger.py` instead of inline here?

kycutler: I can add it there to test as well. The intent with adding it here is to catch any errors with e.g. DLL references early so we can fail just once up front and fallback to not normalizing, rather than failing every time we get the path below.

blink1073 (approved):
Looks good, thank you!

PR #812, opened 2021-11-30
Set debugOptions for breakpoints in python standard library source

echarles:
Users need to be able to add breakpoints in python standard library source.

For this, the needed flag `DebugStdLib` needs to be added in the `debugOptions` when attaching the debug session.

PR #839, opened 2022-01-12
Removed DebugStdLib from arguments of attach

### LINKED ISSUE #832: v6.6.0 broke debugging "step over" instruction at top level
Since ipykernel>=6.6.0, using the "step over" instruction when debugging top-level code will step into IPython runtime code (see GIF below). This is confusing behavior and definitely not what most users will want or expect.

stepoverbug

I believe this regression is due to #812. This also complicates the stack trace by including frames from IPython -- at best this clutters the interface, and at worst this could confuse users as to what code is running or where their code is.

JohanMabille:
Fixes #832.

This reverts #812 , which is responsible for the bug of #832 and another issue where the code suddenly breaks in ipykernel sources. Resuming the execution leads to many errors in the kernel.

debugstdlib

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -477,8 +477,11 @@ class Debugger:
             'port': port
         }
         message['arguments']['logToFile'] = True
+        # Reverts that option for now since it leads to spurious break of the code
+        # in ipykernel source and resuming the execution leads to several errors
+        # in the kernel.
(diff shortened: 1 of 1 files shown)


JohanMabille:
Actually this does not fix the second bug, it just changes the point where the code breaks.

blink1073 (approved):
Thank you!

PR #844, opened 2022-01-14
Handled AllThreadsContinued and workaround for wrong threadId in cont…

JohanMabille:
…inued event after a next request

This PR adds correct handling of `allThreadsContinued` attribute of `continued` event. Besides, it workarounds a bug of debugpy/pydev where the `threadId` attribute of the `continued` event is not correct after a `next` request. Notice the workaround is working only if a single thread is executed on the shell.

This bug results in the weird behavior noticed in [Lab](https://github.com/jupyterlab/jupyterlab/pull/11566). An issue has been opened on the [debugpy repo](https://github.com/microsoft/debugpy/issues/819).

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -1,5 +1,6 @@
 import os
 import re
+import threading
 
 import zmq
 from zmq.utils import jsonapi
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!

PR #848, opened 2022-02-01
Fix the debug modules model

echarles:
Fixes https://github.com/jupyterlab/jupyterlab/issues/11966

We need to respect the debug modules https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Modules

This PR updates the layout of the returned JSON (reply on the modules request) to respect the spec.

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -603,8 +603,9 @@ class Debugger:
             filename = getattr(getattr(module, '__spec__', None), 'origin', None)
             if filename and filename.endswith('.py'):
                 mods.append({
-                    'id': module.__name__,
-                    'name': filename
+                    'id': i,
+                    'name': module.__name__,
+                    'path': filename
                 })
 
         reply = {'body': {'modules': mods, 'totalModules': len(modules)}}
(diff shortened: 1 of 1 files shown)


ipykernel/debugger.py
```diff
@@ -600,11 +600,14 @@ async def modules(self, message):
...
-            filename = getattr(getattr(module, "__spec__", None), "origin", None)
-            if filename and filename.endswith(".py"):
-                mods.append({module.__name__: filename})
-
-        reply = {"body": {"modules": mods, "totalModules": len(modules)}}
+            filename = getattr(getattr(module, '__spec__', None), 'origin', None)
+            if filename and filename.endswith('.py'):
+                mods.append({
+                    'id': module.__name__,
+                    'name': filename
```
fcollonval, line 607: From the screenshot [there](https://github.com/jupyterlab/jupyterlab/issues/11966#issuecomment-1026031708), it seems more appropriate to use:

```python
{
  'id': <incremental integer>,
  'name': module.__name__,
  'path': filename
}
```

echarles: Done. Thx @fcollonval

blink1073:
Kicking CI

blink1073 (approved):
Thanks!

blink1073:
I'll hold off on a release for a bit in case we can get #849 included.

PR #853, opened 2022-02-03
use message queue for abort_queues

minrk:
ensures queued messages are aborted, rather than racing with coroutine processing which would complete too soon during asynchronous executions.

Discovered in https://github.com/ipython/ipyparallel/pull/663 where making `do_execute` return a Future would cause abort tests to fail.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -671,7 +671,7 @@ class Kernel(SingletonConfigurable):
         self.log.debug("%s", reply_msg)
 
         if not silent and reply_msg['content']['status'] == 'error' and stop_on_error:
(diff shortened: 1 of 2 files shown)


blink1073:
This test has failed twice in a row, seems relevant:

```
FAILED work/ipykernel/ipykernel/test_venv/lib/python3.9/site-packages/jupyter_client/tests/test_kernelmanager.py::TestParallel::test_start_parallel_process_kernels[tcp]
```

minrk:
Re-ran the tests and they passed. I don't really see how it could have affected that test.

blink1073 (approved):
Thank you!

PR #855, opened 2022-02-04
Fixed event forwarding

JohanMabille:
I know a lot of calls to `event_callback` could be factorized out of the `if/else` logic, but I find it more readable this way.


fcollonval:
Sadly [structural pattern matching](https://www.python.org/dev/peps/pep-0636/) is a new thing only in Python 3.10

SylvainCorlay:
I think we should make a patch release including this fix as soon as possible.

blink1073:
I can make a release on Monday, I try not to release before a weekend.

PR #861, opened 2022-02-11
process_one only accepts coroutines for dispatch

### LINKED ISSUE #860: IPykernel 6.9.0 generates error when Python errors are returned after execution
This is very simple to reproduce in QtConsole. After executing something like this:

imagen

you'll see the following traceback in the terminal where it was started

```python-traceback
  File "/home/carlos/miniconda3/envs/py310/lib/python3.10/site-packages/ipykernel/kernelbase.py", line 461, in dispatch_queue
    await self.process_one()
  File "/home/carlos/miniconda3/envs/py310/lib/python3.10/site-packages/ipykernel/kernelbase.py", line 450, in process_one
    await dispatch(*args)
TypeError: object NoneType can't be used in 'await' expression
```

I can't reproduce this in 6.8.0 or older versions and it's not related to the Python version I'm using (it was also reproduced in 3.8, as reported [here](https://github.com/spyder-ide/spyder/issues/17316)).

I'd say this was introduced in PR #853, but I don't know how to fix it.

minrk:
closes #860

alternately: make the `await dispatch` check `inspect.isawaitable` first.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -985,7 +985,8 @@ class Kernel(SingletonConfigurable):
         self.shell_stream.flush()
 
         # Callback to signal that we are done aborting
-        def stop_aborting():
+        # dispatch functions _must_ be async
+        async def stop_aborting():
             self.log.info("Finishing abort")
             self._aborting = False
 


blink1073:
I'm not sure how `asyncio` handles an `async` function that is used as a callback to `call_later`, it doesn't say in the [docs](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_later).  I think the alternate approach you mentioned is safer.

minrk:
`call_later` isn't getting an async function to call. `call_later` is being passed the synchronous `schedule_dispatch` function to call, with the async function passed as an _argument_, eventually to be called and awaited by us [here](https://github.com/ipython/ipykernel/blob/2ad426e383b7a52c36aa5de524d068617ad53baf/ipykernel/kernelbase.py#L451).

blink1073:
Ah, understood, thanks!

Issue #862, opened 2022-02-14
Add a configuration to debug in the standard library


echarles:
We need a way to enable the standard library removed in https://github.com/ipython/ipykernel/commit/aa0d637bb8a7094f4c29259c078569788ad9bc12

A configuration to enable that feature is a good option. This feature will be disabled by default.

PR #863, opened 2022-02-14
Enable standard library debugging via config

echarles:
Fix https://github.com/ipython/ipykernel/issues/862

This adds a configuration `debug_just_my_code` being True by default.

The name has been chosen based on the existing VS Code configuraiton https://docs.microsoft.com/en-us/visualstudio/debugger/just-my-code?view=vs-2022

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -270,13 +270,14 @@ class Debugger:
         'richInspectVariables', 'modules'
     ]
 
-    def __init__(self, log, debugpy_stream, event_callback, shell_socket, session):
+    def __init__(self, log, debugpy_stream, event_callback, shell_socket, session, just_my_code):
         self.log = log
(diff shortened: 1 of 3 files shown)


ipykernel/debugger.py
```diff
@@ -270,13 +270,14 @@ class Debugger:
         'richInspectVariables', 'modules'
     ]
 
-    def __init__(self, log, debugpy_stream, event_callback, shell_socket, session):
```
blink1073, line 273: Can we make this an optional arg so it doesn't change the API?

echarles: @blink1073 Thx for the review. I have just pushed a change for that.

blink1073 (approved):
LGTM, thanks!

echarles:
Thx @blink1073

Issue #864, opened 2022-02-14
More information on the host in the usage reply


echarles:
We need to add more information regarding the hostname in the usage reply.

This will be used by https://github.com/Quansight/jupyterlab-kernel-usage/issues/4

PR #865, opened 2022-02-14
Add hostname to the usage reply

echarles:
Fix https://github.com/ipython/ipykernel/issues/864

The hostname is added to the usage reply


blink1073 (approved):
Thanks!

echarles:
Thx @blink1073 👍

PR #869, opened 2022-02-17
BUG: Kill subprocesses on shutdown.

### LINKED ISSUE #193: Import deadlock with streams
Write this text to the file `minimal.py`:

```
import sys

sys.__stdout__.write('About to do\n')
sys.__stdout__.flush()
sys.stdout.write('Doing\n')
sys.stdout.flush()
sys.__stdout__.write('Done\n')
sys.__stdout__.flush()
```

Start up an IPython (Python 2; the issue may not exist in Python 3) notebook. Execute `import minimal`. The kernel will become unresponsive.

The problem is that `sys.stdout.flush()` will call the [`OutputStream.flush()` method](https://github.com/ipython/ipykernel/blob/master/ipykernel/iostream.py#L259-L271) which adds a callback to the event loop that sends off a `zmq` message to the notebook. That callback is executed in another thread. To create the message to send, it will create a UUID using `uuid.uuid4()` which has a local import of `os` in it. Python 2 has a global import lock that was acquired by the main thread which is executing our `import minimal`. That import does not complete because `OutputStream.flush()` is synchronous and is waiting for the event loop to be processed. Deadlock.

As far as I can tell, this is the only place that something is imported at runtime in the message-sending thread, so reimplementing `uuid.uuid4()` to not locally import would be a minimal fix that avoids the issue.

Carreau:
Fixes jupyter/jupyter_client#104.

This should make sure we properly cull all subprocesses at shutdown,
it does change one of the private method from sync to async in order to
no user time.sleep or thread so this may affect subclasses, though I
doubt it.

It's also not completely clear to me whether this works on windows as
SIGINT I belove is not a thing.

Regardless as this affects things like dask, and others that are mostly
on unix, it should be an improvement.

It does the following, stopping as soon as it does not find any more
children to current process.

 - Send sigint to everything
 - Immediately send sigterm in look with an exponential backoff from
   0.01 to 1 second roughtly multiplying the delay until next send by 3
     each time.
 - Switch to sending sigkill with same backoff.

There is no delay after sigint, as this is just a courtesy.
The delays backoff are not configurable. I can imagine that on slow
systems it may make sens

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -1136,8 +1136,7 @@ class Kernel(SingletonConfigurable):
             raise EOFError
         return value
 
-
-    async def _killpg(self, *, signal):
+    def _killpg(self, signal):
         """
         similar to killpg but use psutil if it can on windows
         or if pgid is none
@@ -1147,8 +1146,8 @@ class Kernel(SingletonConfigurable):
         if pgid and hasattr(os, "killpg"):
             try:
                 os.killpg(pgid, signal)
-            except OSError:
-                self.log.warning("OSError running killpg, not killing children")
+            except (OSError) as e:
+                self.log.exception(f"OSError running killpg, not killing children.")
             return
         elif psutil is not None:
             children = parent.children(recursive=True)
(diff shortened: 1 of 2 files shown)


ipykernel/kernelbase.py
```diff
@@ -5,44 +5,49 @@
...
 import socket
+import sys
 import time
 import uuid
 import warnings
+from datetime import datetime
+from functools import partial
+from signal import SIGINT, SIGTERM, Signals, default_int_handler, signal
+
+if sys.platform == "win32":
```
blink1073, line 21: Should be `!=`?

Carreau: Yep, I'm dumb, and there are other issues on windows. I think I'm going to just not try to do that on windows, it might not even be a problem. I don't know how processes on windows work and maybe they get reaped anyway ?

ipykernel/kernelbase.py
```diff
@@ -1131,9 +1136,62 @@ def _input_request(self, prompt, ident, parent, password=False):
             raise EOFError
         return value
 
-    def _at_shutdown(self):
+    async def _progressively_terminate_all_children(self):
+        if sys.platform != "win32":
```
blink1073, line 1140:
```suggestion
        if sys.platform == "win32":
```

ipykernel/kernelbase.py
```diff
@@ -1131,9 +1136,62 @@ def _input_request(self, prompt, ident, parent, password=False):
             raise EOFError
         return value
 
-    def _at_shutdown(self):
+    async def _progressively_terminate_all_children(self):
+        if sys.platform == "win32":
```
blink1073, line 1140: I think this method should work on both Windows and Posix: https://stackoverflow.com/a/25134985

Carreau: I'll give it a try. I know that the "loop and kill children" was switched back to `killpg` at some point in jupyter_client as the first one had issues.

Carreau: (that was in jupyter/jupyter_client#743, i'll try to get the same logic from there)

blink1073:
*Weird*.  I have no idea why the `ipyparallel` downstream test is getting skipped in this PR.  It didn't get skipped in my PR or in `main`.

blink1073 (approved):
Thanks!

Carreau:
> Thanks!

Thanks to you for the review and everything.

PR #874, opened 2022-02-21
Only kill children in process group at shutdown

### LINKED ISSUE #872: ipyparallel downstream test broken
It looks like #869 broke our downstream [`ipyparallel` tests](https://github.com/ipython/ipykernel/runs/5253781618?check_suite_focus=true).  I hadn't been able to diagnose it when reviewing the PR because the job wasn't showing any output at all.

minrk:
- not the whole process group (`killpg`), which includes kernel itself and possibly parents (closes #872)
- not children which have started their own process groups, which is the standard indicator for "leave me alone when my parent shuts down"

closes #873

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -19,10 +19,14 @@ except ImportError:
 
 from .compiler import (get_file_name, get_tmp_directory, get_tmp_hash_seed)
 
-# This import is required to have the next ones working...
-from debugpy.server import api  # noqa
-from _pydevd_bundle import pydevd_frame_utils
-from _pydevd_bundle.pydevd_suspended_frames import SuspendedFramesManager, _FramesTracker
+try:
+    # This import is required to have the next ones working...
+    from debugpy.server import api  # noqa
+    from _pydevd_bundle import pydevd_frame_utils
+    from _pydevd_bundle.pydevd_suspended_frames import SuspendedFramesManager, _FramesTracker
+    _is_debugpy_available = True
+except ImportError:
+    _is_debugpy_available = False
 
 # Required for backwards compatiblity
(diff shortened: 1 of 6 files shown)


minrk:
One question here: this cleanup logic depends on psutil, but I haven't made psutil a hard dependency. So we have a choice:

1. make this cleanup an optional enhancement when psutil is available, or
2. bump psutil to a hard dependency so this always runs

Carreau:
Previous PR made is a dependency on windows, so I think we can make it a hard dep.

Carreau:
pushed 5c16fde1ce3dd5dd4489bf1969688f78c8a6ec3d to not make psutil necessary to build the wheel and changing what is imported where.

@echarles Is that ok with you ?

echarles:
> @echarles Is that ok with you ?

Moving `_is_debugpy_available` to `debugger.py` makes sense (related to `debugpy` dependency). Looking more to the `psutil` dependency, main branch only installs `psutil` for windows.

https://github.com/ipython/ipykernel/blob/78c83ad9ea4ab152698d7f5cbe7eb5c0069afae8/setup.py#L71

which is needed on all platform to honor the `usage_request`

https://github.com/ipython/ipykernel/blob/78c83ad9ea4ab152698d7f5cbe7eb5c0069afae8/ipykernel/kernelbase.py#L896-L901

Any particular reason to restrict `psutil` to only windows?

Carreau:
> Any particular reason to restrict `psutil` to only windows?

Well originally it was because killpg is not avail on windows, I guess it was not added as a dep when `usage_request` was added, so it's an argument to make it a hard dependency.

Carreau:
Ok, It seem we are all in agreement, merging.

Issue #870, opened 2022-02-17
Can debugger forward "source" requests to debugpy?


vidartf:
The current handler for "source" requests to the debugger is as follows:

https://github.com/ipython/ipykernel/blob/221dca63d7b2e2b8fea32bf8a101d07645fc9d3c/ipykernel/debugger.py#L427-L445

Intuitively, I would expect that the source request would be forwarded to debugpy, so that the [more complete implementation of pydevd](https://github.com/microsoft/debugpy/blob/cf0d684566edc339545b161da7c3dfc48af7c7d5/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py#L1108-L1140) could handle the request. Some differences of that implementation:
- Both implementations tries direct file access first.
- pydevd then tries what it calls the "linecache" with the following comment:
  >   File might not exist at all, or we might not have a permission to read it, but it might also be inside a zipfile, or an IPython cell. In this case, linecache might still be able to retrieve the source.
- Finally, pydevd falls back to using the sourceReference supplied (which would previously be given on e.g. "module" events I think).

One of the advantages of the approach of pydevd is that it also is able to handle modules that are loaded from custom module loaders (zip file loader, DB loaders, web loaders, etc.). It would therefore be good if someone could write out why we're not simply forwarding the request to debugpy, and if there is a solid reason for it, maybe we could improve upon it so that we can still support the sourceReference field?

vidartf:
(CC @JohanMabille )

JohanMabille:
Hi @vidartf ,

Sorry for the late reply, but I have now the whole story ;)

When sending a `Source` [request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Source), you need to provide a `sourceReference` argument because of old adapters. The issue is that the sources of notebook cells are never loaded, the are created by the kernel which dumps the content in temporary files. Therefore, there is no sourceReference for them. Besides, specifying `sourceReference: 0` in the sourceRequest does not work; debugpy does not find the source if you do so, even if you provide a `Source` argument with the right path.

Therefore it is required that the kernel handles the request when `sourceReference` is 0.

PR #875, opened 2022-02-22
Forward source requests to debugpy as fallback

vidartf:
~If the source cannot be looked up by the `path` argument, forward it to debugpy as it might be able to resolve it using the sourceReference field.~

EDIT: If the `sourceReference` is non-zero, forward it to debugpy, as the specification requires the reference to be used in those cases.

Proposed fix to #870.

TODO: Write some tests.

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -442,9 +442,7 @@ class Debugger:
                     'content': f.read()
                 }
         else:
-            reply['success'] = False
-            reply['message'] = 'source unavailable'
-            reply['body'] = {}
+            reply = await self._forward_message(message)
 
         return reply
 
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -434,15 +434,23 @@ class Debugger:
             'request_seq': message['seq'],
             'command': message['command']
         }
-        source_path = message["arguments"]["source"]["path"]
-        if os.path.isfile(source_path):
-            with open(source_path, encoding='utf-8') as f:
-                reply['success'] = True
-                reply['body'] = {
-                    'content': f.read()
-                }
-        else:
+        source_ref = message["arguments"].get("sourceReference", 0)
+        if source_ref == 0:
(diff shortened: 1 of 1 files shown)


ipykernel/debugger.py
```diff
@@ -442,9 +442,7 @@ async def source(self, message):
                     'content': f.read()
                 }
         else:
-            reply['success'] = False
-            reply['message'] = 'source unavailable'
-            reply['body'] = {}
+            reply = await self._forward_message(message)
```
JohanMabille, line 445: See my [comment](https://github.com/ipython/ipykernel/issues/870#issuecomment-1049008052); I think the right implementation should be to handle the request in the kernel if and only if the attribute `sourceReference` is 0. Notice that the `source` attribute is optional, accoriding to the [DAP](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Source), therefore this implementation is not totally compliant.

vidartf: I've had another go at this, trying to follow the spec more closely this time.

ipykernel/debugger.py
```diff
@@ -434,17 +434,23 @@ async def source(self, message):
...
-        source_path = message["arguments"]["source"]["path"]
-        if os.path.isfile(source_path):
-            with open(source_path, encoding='utf-8') as f:
-                reply['success'] = True
-                reply['body'] = {
-                    'content': f.read()
-                }
+        source_ref = message["arguments"].get("sourceReference", 0)
+        if source_ref == 0:
+            message["arguments"].get("source", {}).get("sourceReference", 0)
```
r3m0t, line 439:
```suggestion
            source_ref = message["arguments"].get("source", {}).get("sourceReference", 0)
```

vidartf: @r3m0t Thanks for testing. This change is meant to solve this for non-vscode usages as well 👍

Carreau:
it's been 4+ years; closing

PR #876, opened 2022-02-23
Drop Tornado

### LINKED ISSUE #656: Dropping the dependency to tornado
We are not using the tornado coroutines anymore in the codebase (since #632) which also fixed some issue with respect to context variables. I think that it would make sense to drop tornado completely at some point.

I am opening this issue to track the work required for this, as it may require significant work in the codebase:

 - switching the asyncio API of pyzmq.
 - asyncio enforces stronger constraints about instanciating queues and loops in the thread where they are run.

davidbrochart:
Closes #656

### DIFF
--- a/ipykernel/control.py
+++ b/ipykernel/control.py
@@ -1,19 +1,19 @@
+import asyncio
 from threading import Thread
-
-from tornado.ioloop import IOLoop
 
 
 class ControlThread(Thread):
     def __init__(self, **kwargs):
         Thread.__init__(self, name="Control", **kwargs)
-        self.io_loop = IOLoop(make_current=False)
         self.pydev_do_not_trace = True
         self.is_pydev_daemon_thread = True
+        self.io_loop = asyncio.new_event_loop()
 
     def run(self):
         self.name = "Control"
+        asyncio.set_event_loop(self.io_loop)
         try:
-            self.io_loop.start()
+            self.io_loop.run_forever()
         finally:
             self.io_loop.close()
 
@@ -22,4 +22,4 @@ class ControlThread(Thread):
 
         This method is threadsafe.
         """
-        self.io_loop.add_callback(self.io_loop.stop)
+        self.io_loop.call_soon_threadsafe(self.io_loop.stop)
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -1,3 +1,4 @@
+import asyncio
 import os
 import re
 import sys
@@ -6,8 +7,6 @@ import typing as t
 import zmq
 from IPython.core.getipython import get_ipython
 from IPython.core.inputtransformer2 import leading_empty_lines
-from tornado.locks import Event
-from tornado.queues import Queue
 from zmq.utils import jsonapi
 
 try:
@@ -86,11 +85,13 @@ class DebugpyMessageQueue:
     SEPARATOR = "\r\n\r\n"
     SEPARATOR_LENGTH = 4
 
+    message_queue: asyncio.Queue[t.Dict[str, t.Any]]
+
     def __init__(self, event_callback, log):
         self.tcp_buffer = ""
         self._reset_tcp_pos()
         self.event_callback = event_callback
-        self.message_queue: Queue[t.Any] = Queue()
+        self.message_queue = asyncio.Queue()
         self.log = log
 
     def _reset_tcp_pos(self):
@@ -171,7 +172,7 @@ class DebugpyClient:
         self.debugpy_port = -1
         self.routing_id = None
         self.wait_for_attach = True
-        self.init_event = Event()
+        self.init_event = asyncio.Event()
         self.init_event_seq = -1
 
     def _get_endpoint(self):
@@ -245,7 +246,7 @@ class DebugpyClient:
     def disconnect_tcp_socket(self):
         self.debugpy_stream.socket.disconnect(self._get_endpoint())
         self.routing_id = None
-        self.init_event = Event()
+        self.init_event = asyncio.Event()
         self.init_event_seq = -1
         self.wait_for_attach = True
 
@@ -278,6 +279,8 @@ class Debugger:
         "configurationDone",
     ]
 
+    stopped_queue: asyncio.Queue[t.Dict[str, t.Any]]
+
     # Requests that can be handled even if the debugger is not running
     static_debug_msg_types = ["debugInfo", "inspectVariables", "richInspectVariables", "modules"]
 
@@ -291,7 +294,7 @@ class Debugger:
         self.is_started = False
         self.event_callback = event_callback
         self.just_my_code = just_my_code
-        self.stopped_queue: Queue[t.Any] = Queue()
+        self.stopped_queue = asyncio.Queue()
 
         self.started_debug_handlers = {}
         for msg_type in Debugger.started_debug_msg_types:
@@ -367,7 +370,7 @@ class Debugger:
     def tcp_client(self):
         return self.debugpy_client
 
-    def start(self):
+    async def start(self):
         if not self.debugpy_initialized:
             tmp_dir = get_tmp_directory()
             if not os.path.exists(tmp_dir):
@@ -384,7 +387,7 @@ class Debugger:
                 (self.shell_socket.getsockopt(ROUTING_ID)),
             )
 
-            ident, msg = self.session.recv(self.shell_socket, mode=0)
+            ident, msg = await self.session.async_recv(self.shell_socket, mode=0)
             self.debugpy_initialized = msg["content"]["status"] == "ok"
 
         # Don't remove leading empty lines when debugging so the breakpoints are correctly positioned
@@ -631,7 +634,7 @@ class Debugger:
             if self.is_started:
                 self.log.info("The debugger has already started")
             else:
-                self.is_started = self.start()
+                self.is_started = await self.start()
                 if self.is_started:
                     self.log.info("The debugger has started")
                 else:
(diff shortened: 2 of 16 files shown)


ipykernel/ipkernel.py
```diff
@@ -282,8 +282,8 @@ def set_sigint_result():
                 if sigint_future.cancelled() or sigint_future.done():
                     return
                 sigint_future.set_result(1)
-            # use add_callback for thread safety
-            self.io_loop.add_callback(set_sigint_result)
+            # use call_soon_threadsage for thread safety
```
blink1073, line 285:
```suggestion
            # use call_soon_threadsafe for thread safety
```

davidbrochart: Thanks!

ipykernel/control.py
```diff
@@ -1,25 +1,24 @@
...
 class ControlThread(Thread):
 
     def __init__(self, **kwargs):
         Thread.__init__(self, name="Control", **kwargs)
-        self.io_loop = IOLoop(make_current=False)
         self.pydev_do_not_trace = True
         self.is_pydev_daemon_thread = True
+        self.io_loop = None
+        self.loop_ready = Event()
+        self.start()
```
fcollonval, line 13: It is not a good idea to start a thread in the init (people that want to extend this class will be very annoyed by that). We should respect the behavior of built-in Thread and we should avoid changing the behavior for backward compatible reason.

ipykernel/iostream.py
```diff
@@ -63,24 +59,33 @@ def __init__(self, socket, pipe=False):
...
         self._pipe_flag = pipe
-        self.io_loop = IOLoop(make_current=False)
+        self.io_loop = None
+        self.loop_ready = threading.Event()
+        self.thread = threading.Thread(target=self._thread_main, name="IOPub")
+        self.thread.daemon = True
+        self.thread.pydev_do_not_trace = True
+        self.thread.is_pydev_daemon_thread = True
+        self.thread.name = "IOPub"
+        self.thread.start()
```
fcollonval, line 69: It is not a good idea to start a thread in the init (people that want to extend this class will be very annoyed by that). We should respect the behavior of built-in Thread and we should avoid changing the behavior for backward compatible reason.

ipykernel/iostream.py
```diff
@@ -63,24 +59,33 @@ def __init__(self, socket, pipe=False):
...
+        # make sure we don't prevent process exit
+        # I'm not sure why setting daemon=True above isn't enough, but it doesn't appear to be.
+        atexit.register(self.stop)
 
     def _thread_main(self):
         """The inner loop that's actually run in a thread"""
-        self.io_loop.make_current()
-        self.io_loop.start()
-        self.io_loop.close(all_fds=True)
+        self.io_loop = asyncio.new_event_loop()
```
fcollonval, line 82: Can't you create the loop in the init (as done before) and only setting it here? This will avoid using the Event, isn't it?

fcollonval (changes requested):
Thanks @davidbrochart for this. I got a question and a comment.

ipykernel/iostream.py
davidbrochart: I thought that the event loop had to be created inside the thread, but it seems to work when it is created outside. Indeed the Event is not necessary then. Thanks, I'll make the change.

ipykernel/control.py
davidbrochart: At some point I had to put `start` in the `__init__` but it seems that it now works as before, with a separate `start` method. I'll push this change, thanks!

davidbrochart: At some point I had to put `start` in the `__init__` but it seems that it now works as before, with a separate `start` method. I'll push this change, thanks!

davidbrochart:
> Thanks @davidbrochart for this. I got a question and a comment.

Thanks for the review @fcollonval.
This is still a WIP so I'll make this PR a draft. I can see that the kernel is functional e.g. in JupyterLab, but there are still some issues, at least with shutdown.

ipykernel/kernelbase.py
```diff
@@ -837,11 +835,12 @@ async def shutdown_request(self, stream, ident, parent):
...
         self.log.debug('Stopping control ioloop')
         control_io_loop = self.control_stream.io_loop
-        control_io_loop.add_callback(control_io_loop.stop)
+        control_io_loop.call_soon_threadsafe(control_io_loop.stop)
 
         self.log.debug('Stopping shell ioloop')
         shell_io_loop = self.shell_stream.io_loop
-        shell_io_loop.add_callback(shell_io_loop.stop)
+        # FIXME: shouldn't need to be threadsafe
+        shell_io_loop.call_soon_threadsafe(shell_io_loop.stop)
```
davidbrochart, line 843: I thought that we shouldn't need to `call_soon_threadsafe` because the shell event loop runs in the main thread, but actually we do because we are processing the shutdown request from the control channel, which runs in another thread.
Correct me if I'm wrong :smile:

ipykernel/debugger.py
```diff
@@ -382,7 +382,7 @@ def start(self):
             self.session.send(self.shell_socket, 'execute_request', content,
                               None, (self.shell_socket.getsockopt(ROUTING_ID)))
 
-            ident, msg = self.session.recv(self.shell_socket, mode=0)
+            ident, msg = await self.session.async_recv(self.shell_socket, mode=0)
```
davidbrochart, line 385: Will need https://github.com/jupyter/jupyter_client/pull/749.

ipykernel/iostream.py
```diff
@@ -445,9 +445,7 @@ def _schedule_flush(self):
 
         # add_timeout has to be handed to the io thread via event pipe
         def _schedule_in_thread():
-            # FIXME: original code was:
-            # self._io_loop.call_later(self.flush_interval, self._flush)
-            self._io_loop.call_soon_threadsafe(self._flush)
+            self._io_loop.call_soon_threadsafe(self._io_loop.call_later, self.flush_interval, self._flush)
```
davidbrochart, line 448: Not sure if `call_soon_threadsafe` is needed. If not, we could keep the original code.

ipykernel/kernelapp.py
```diff
@@ -367,7 +367,7 @@ def close(self):
             if socket and not socket.closed:
                 socket.close()
         self.log.debug("Terminating zmq context")
-        # FIXME
+        # FIXME: this line breaks shutdown
         # self.context.term()
```
davidbrochart, line 371: Not sure why terminating the ZMQ context creates issues here.

blink1073: Maybe related: https://github.com/jupyter-server/jupyter_server/issues/677#issuecomment-1033597413

davidbrochart: Thanks @blink1073, I can see that currently, [we don't close the heartbeat socket before terminating the ZMQ context](https://github.com/ipython/ipykernel/blob/d6744f9e423dacc6b317b1d31805304e89cbec5d/ipykernel/heartbeat.py#L112), but the other way around: we wait for the context to be terminated before closing the socket.
Maybe this can be the root cause, @minrk ?

ipykernel/zmqstream.py
```diff
@@ -0,0 +1,31 @@
...
+
+    async def recv_task(self, callback, copy):
+        while True:
+            msg_list = await self.socket.recv_multipart(copy=copy)
+            callback(msg_list)
+
+    def send_multipart(self, msg_list, flags=0, copy=True, track=False, **kwargs):
+        return self.socket.send_multipart(msg_list, copy=copy)
+
+    def flush(self, flag=None, limit = None):
```
minrk, line 30: Flush has historically been very important behavior for ipykernel to behave appropriately. Making it a no-op doesn't seem great to me (It may not be as relevant today with everything being async, but this is a big change).

ipykernel/kernelbase.py
minrk: I think you're right

ipykernel/kernelapp.py
minrk: Context.term blocks until all sockets are closed and all unsent messages are delivered or discarded. A socket may close after the context if it is done from another thread. But it can't be done from a coroutine, for instance.

The things to check when a context hangs are always:

1. are there sockets that haven't been closed, and
2. was LINGER set on those sockets so they discard messages eventually

You can debug leftover sockets with `self.context._sockets` which is a set of weakrefs to sockets (this is private and subject to change, but works for debugging).

The heartbeat socket [doesn't share this context](https://github.com/ipython/ipykernel/blob/d6744f9e423dacc6b317b1d31805304e89cbec5d/ipykernel/kernelapp.py#L337), so I don't think that's relevant.

ipykernel/zmqstream.py
davidbrochart: But flush seems to be supported only for Tornado's IOLoop and ZMQStream, right? I haven't found any reference of it outside of this context.

minrk: That's true, because `send` is async. We don't need this anymore as long as we `await` all our sends.

davidbrochart:
Although ipykernel seems functional with these changes, I don't really understand the test failures.
I think the next step is getting rid of the `ZMQStream` with their `on_recv` callbacks (which were really specific to Tornado), and just have tasks that receive messages and handle them. I would also like to get rid of the jupyter_client dependency. It seems weird for a "server" to depend on a "client".

minrk:
> I would also like to get rid of the jupyter_client dependency

I wouldn't do that. `jupyter-protocol` would have been a better name for that repo, since that's really what it defines. It is meant to be a dependency for any Python use of the Jupyter protocol, and that includes ipykernel.

davidbrochart:
Closing in favor of #1079.

Issue #334, opened 2018-08-15
Memory Error: IOStream.Flush timed out


acerock6:
Hi,

I was running a long session on Jupyter notebook. I mistakenly ran a function with no variables and the kernel got stuck. After a while, I saw the error: Memory Error., IOStream.Flush timed out
When I checked the OS task manager, there was sufficient memory present. What's weird is the kernel got stuck at a simple function call with no variables (something like do_something() )
Anyone has any idea how this might have occurred and what's the best way to avoid the same in the future (I lost 10-12 hours of work due to this error)
TIA

RubensZimbres:
It happened to me also, while trying to see the output of a function that used multiprocessing.

MustafaWaheed91:
I had the same issue when running a program in jupyter kernel that also was using multiprocessing module.

Abhishek-31:
I am also having the same issue.

kribu:
me too :/

yuhang-lin:
Same here. Jupyter was up for 4 days and I had to kill it by control-C.

zinglax:
Getting this issue when trying to import a module installed for local development ( pip install -e . )

omerb01:
after upgrading my Mac from Mojave to Catalina, I experience the same issue with a phrase of code that worked before.

Schmidtbit:
I am having this problem as well. Anyone find a reason for this and how to address it? I was running a tsfresh job

```
from tsfresh import select_features

rel = select_features(extracted_features, y)
```

lymweb:
me too :/

Rafaila98:
Hi, 
I had the same problem and it was solved by restarting jupyter. Exit and then `jupyter notebook` command again.

waterflyer:
Same problem here, instead of exit and relaunch the jupyter notebook, is there any way to fix it inside the program? thanks.

jkim2mozjay:
Just ran into this problem too

Pranav-Goel:
running into this problem myself :/

wfxu:
me to

StephanieAuer-Moodys:
I'm also having this problem applying a function to a dask dataframe.

jaded0:
This problem is still happening, l
`lyrics_list_clean = list(allpossible['letras'].apply(clean_lyrics, convert_dtype=str))`   
freezes at only about 200, again at about 900, then no progress at all

daniel-fry:
[...]

PR #884, opened 2022-03-22
Check if the current thread is the io thread

jamadeo:
I noticed that I'll somewhat unpredictably run into this issue where the kernel prints "IOStream.flush timed out". I found that OutStream.flush() could be called in the I/O thread itself, leading to a wait() that will block for the full 10 second timeout.

I can't reliably reproduce an instance of flush() getting called in the I/O thread, but in my case it looked like a `logger.warning` call inside the asyncio event loop handling was responsible for it.

I think this could be the issue in #334 ?

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -455,7 +455,12 @@ class OutStream(TextIOBase):
 
         send will happen in the background thread
         """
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thank you!

PR #886, opened 2022-03-22
Try to pass cell id to executing kernel.

Carreau:
Is seem that few kernel want to make use of cell id for various
reasons.

One way to get the cell_id in the executing context is to make use of
init_metadata but it is marked for removal.

for example in [there](https://github.com/robots-from-jupyter/robotkernel/blob/19b28267406d1f8555ce73b96e7efb8af8417266/src/robotkernel/kernel.py#L196-L205)

Another is to start passing cell_id down.

This is technically a change of API as our consumer may not support more
parameters so we take a peak at the signature, and pass cell_id only if
it is :
- an explicit parameter,
- the function/method takes varkwargs.

We could also start to refactor to pass a metadata object with multiple
fields if this is the preferred route.

One questions is whether and how we want to deprecated not receiving
cell_id as a parameter.

Related to https://github.com/ipython/ipython/issues/13579 and
subsequent PR on IPython side to support cell_id in progress.

### DIFF
--- a/ipykernel/inprocess/tests/test_kernel.py
+++ b/ipykernel/inprocess/tests/test_kernel.py
@@ -3,20 +3,18 @@
 
 import sys
 import unittest
+from contextlib import contextmanager
 from io import StringIO
 
 import pytest
 import tornado
 from IPython.utils.io import capture_output
+from jupyter_client.session import Session
 
 from ipykernel.inprocess.blocking import BlockingInProcessKernelClient
 from ipykernel.inprocess.ipkernel import InProcessKernel
 from ipykernel.inprocess.manager import InProcessKernelManager
 from ipykernel.tests.utils import assemble_output
-from IPython.utils.io import capture_output
-
-from jupyter_client.session import Session
-from contextlib import contextmanager
 
 orig_msg = Session.msg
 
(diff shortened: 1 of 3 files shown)


blink1073:
I'm addressing the CI failures in #887

SylvainCorlay:
I am not sure about this. Wouldn't it be some kind of abstraction leak?

Carreau:
IMHO that ship has sailed, this info is already sent as part of the execute
request from front end to here, and is already used by other kernels. The
goal is also to help having a mapping from execution/limiting to original
cell.
I agree that kernel should be able to work without, but I also don’t see
any reason to prevent it.

SylvainCorlay:
OK, LGTM.

blink1073:
I assume we can't add a test for this until `ipython` supports it?

Carreau:
> I assume we can't add a test for this until `ipython` supports it?

I should probably add some test that subclasses/methods that both do and do not support gettign cell_id behave properly. I'll try to do that at least.

blink1073:
@Carreau I think you need to rebase and run black to avoid the merge conflicts.  I don't yet quite trust the meeseeks black runner, it didn't seem to work when I tried [yesterday](https://github.com/jupyterlab/jupyterlab/pull/12280#issuecomment-1080899967)

Carreau:
The remaining failure has been fixed downstream in ipyparallel.
I think this is good to merge.

blink1073 (approved):
Thanks!

Carreau:
Thanks !

gitrubim (commented):
Vírus vunerabilidade

PR #895, opened 2022-03-30
Include method signatures in experimental completion results

MrBago:
Method signatures are trivially available for autocomplete requests, I'd like to make them available via the jupyter complete_request.


blink1073 (approved):
LGTM, thanks!

PR #900, opened 2022-04-01
use packaging instead of pkg_resources to parse versions

minrk:
Pinning setuptools as a runtime dependency is often a problem, and setting a minimum of _extremely_ recent setuptools 60 is causing conflicts with e.g. pandas which pins `setuptools<60` (this is making an unsolvable conda environment for me right now).

[migration advice](https://peps.python.org/pep-0632/#migration-advice) suggests using the lighter `packaging` package, and I don't see a reason to use a version pin.

The pkg_resources function is also available for at least 30 versions (setuptools 30), so the `>=60` doesn't seem to have been the right number to use.

Recommend a patch release with the fix, since the dependency in 6.11 is causing installation problems.

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -9,7 +9,7 @@ import sys
 from functools import partial
 
 import zmq
-from pkg_resources import parse_version as V
+from packaging.version import Version as V
 from traitlets.config.application import Application
 
 
--- a/ipykernel/tests/test_message_spec.py
+++ b/ipykernel/tests/test_message_spec.py
@@ -9,7 +9,7 @@ from queue import Empty
 
 import jupyter_client
 import pytest
-from pkg_resources import parse_version as V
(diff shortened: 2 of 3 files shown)


bollwyvl:
Ick:

```
         File "/tmp/pip-download-q4sqesd9/ipyparallel_89b290df93ac4e98934f08ef9634b493/setupbase.py", line 37, in <module>
          from setuptools.config import StaticModule
      ImportError: cannot import name 'StaticModule' from 'setuptools.config' (/tmp/pip-build-env-tf4je9_5/overlay/lib/python3.9/site-packages/setuptools/config/__init__.py)
      [end of output]
```

bollwyvl:
ugh, it's the homebrew `get_version` in `jupyter_packaging`... i don't think that's a blocker here

blink1073 (approved):
Thanks!

PR #909, opened 2022-04-05
Update setup.py

tlinhart:
Add project URLs to the `setup.py`. Especially adding the source URL is beneficial since it allows tools like Renovate (https://github.com/renovatebot/renovate) handle the dependencies with features like grouping the packages based on source URLs or URL prefixes (see https://docs.renovatebot.com/configuration-options/#matchsourceurlprefixes).


blink1073 (approved):
Nice, thanks!

PR #913, opened 2022-04-10
Add basic mypy support

blink1073:
- Add mypy checking using a manual pre-commit hook
- Add py.typed file for downstream typing support


echarles (approved):
Code change LGTM. CI is green. Tested locally and kernel works fine with jupyterlab. Merging. Thx @blink1073

PR #915, opened 2022-04-11
Fixes #914

### LINKED ISSUE #914: No module named 'IPython.qt'
WHen trying to open a qtconsole in jupyter notebook

in this line 

https://github.com/ipython/ipykernel/blob/51a613d501a86073ea1cdbd8023a168646644c6a/ipykernel/connect.py#L109

the command 

    cmd = ";".join(["from IPython.qt.console import qtconsoleapp", "qtconsoleapp.main()"])

is spawned. However on my machine this results in the error in the header. For me to make it work I have to change that to 

   cmd = ";".join(["from qtconsole import qtconsoleapp", "qtconsoleapp.main()"])

then everything is working the way it should. It seems this is the more uptodate way to spawn an jupyter-console

andia89:
Newer way to spawn qtconsole from notebook

### DIFF
--- a/ipykernel/connect.py
+++ b/ipykernel/connect.py
@@ -106,7 +106,7 @@ def connect_qtconsole(connection_file=None, argv=None):
 
     cf = _find_connection_file(connection_file)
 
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thank you!

PR #924, opened 2022-04-29
Update Pytest configuration (deprecated implicit asyncio mode)

Carreau:
This make downstream test fail.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ testpaths = [
     "ipykernel/tests",
     "ipykernel/inprocess/tests"
 ]
+asyncio_mode = "strict"
 timeout = 300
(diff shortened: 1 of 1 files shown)


Carreau:
Ok, I'm confused why downstream it asks for the option and here it says the option is unknown

blink1073:
If you're referring to `IPython`'s downstream checks, I think the check is using the `pytest` config from `IPython` itself.  In the `downstream-test` action in `jupyterlab/maintainer-tools` we made sure to put the downstream folder in a separate file hierarchy to avoid this.

Carreau:
Thanks for the pointers I'll try to have a look.

Carreau:
Thanks for the pointers I'll try to have a look.

PR #927, opened 2022-05-03
Switch to flit build backend

blink1073:
Uses an [in-tree build backend](https://peps.python.org/pep-0517/#in-tree-build-backends) to handle the creation of the `ipykernel` kernelspec.


blink1073:
Ah, this was a fun experiment but won't work because flit only installs a single package, and we need a separate `ipykernel_launcher.py` that can be run as `python -m ipykernel_launcher`.

Issue #935, opened 2022-05-23
Changing memory calculation methodology from RSS to PSS


nishikantparmariam:
### Problem

[usage_request](https://github.com/ipython/ipykernel/blob/5817d296ad33bca03a43e7fa7dd8bdc0de0696bc/ipykernel/kernelbase.py#L979) should probably use PSS instead of RSS. As RSS overestimates due to copy-on-write. See the below reproducer -

```python3
import multiprocessing
import time
import psutil

def foo():
    time.sleep(10000)

parents_original_memory = bytearray(int(1e9)) # 1GB

for i in range(10):
    multiprocessing.Process(target=foo).start()

def get_memory_info(type):
    process_metric_value = lambda process: getattr(process.memory_full_info(), type)
    current_process = psutil.Process()
    all_processes = [current_process] + current_process.children(recursive=True)
    return (
        f"{sum([process_metric_value(process) for process in all_processes]) / 1e9} GB"
    )

print("RSS: ", get_memory_info("rss"))
print("PSS: ", get_memory_info("pss"))
```

Output is -

```
RSS: 11.590012928 GB 
PSS: 1.082778624 GB
```

PSS seems to be more accurate here.

### Additional context

This is very similar to - https://github.com/jupyter-server/jupyter-resource-usage/issues/130

echarles:
Thx for the report. I have opened https://github.com/ipython/ipykernel/pull/936 to request memory usage with pss.

nishikantparmariam:
One note here - PSS is only [supported in Linux OS](https://psutil.readthedocs.io/en/latest/#:~:text=pss%20(Linux)%3A%20aka%20%E2%80%9CProportional%20Set%20Size%E2%80%9D) via [memory_full_info()](https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_full_info) method. Also, discussed [here](https://github.com/jupyter-server/jupyter-resource-usage/issues/16#issuecomment-615549097).

nishikantparmariam:
We will have to use [memory_full_info](https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_full_info) to get PSS and since PSS is a linux only feature, we might want to check if memory_full_info is an attribute of psutil.Process() ([reference](https://psutil.readthedocs.io/en/latest/#psutil.Process.as_dict)), and fallback to RSS if not supported by the OS.

PR #936, opened 2022-05-23
Request the memory usage with pss mode

echarles:
Fixes https://github.com/ipython/ipykernel/issues/935


echarles:
The current fix does not work. I need to look at this in more details.

echarles:
I am developing on Mac OS, so this may the reason it is failing. I will work out the fallback solution, and would need to create a Docker image to confirm it is working fine on Linux.

echarles:
Closing this PR as https://github.com/ipython/ipykernel/pull/948 is now opened instead.

Issue #912, opened 2022-04-08
Thread to monitor the kernel usage (CPU, Memory...)


echarles:
While implementing a jupyterlab extension to monitor the kernel usage (https://github.com/Quansight/jupyterlab-kernel-usage), a few PRs have been merged to use [psutil](https://psutil.readthedocs.io/en/latest) to return Memory, CPU usage of the machine and the kernel processes.

- https://github.com/ipython/ipykernel/pull/805
- https://github.com/ipython/ipykernel/pull/865
- https://github.com/ipython/ipykernel/pull/908

This has worked great for most of the metrics, but we hit now an issue for the kernel CPU reported on https://github.com/Quansight/jupyterlab-kernel-usage/issues/11. psutil returns the CPU based on the last call, so it looks like everytime we ask the CPU via the kernel websocket, it return zero as psutil considers it is the first time the request is made, see info on:

- https://psutil.readthedocs.io/en/latest/#psutil.cpu_percent
- https://stackoverflow.com/questions/24367424/cpu-percentinterval-none-always-returns-0-regardless-of-interval-value-python/27229967#27229967

Please note that similar issue is reported on the jupyter-resource-usage extension (which monitors the server usage, not the kernel usage) https://github.com/jupyter-server/jupyter-resource-usage/issues/107

A solution would be to maintain a separated thread that would be responsible to return that CPU in a correct way based on psutil. 

What would be the impact to create such as separated thread? What are the already existing threads? Does it sound like a reasonable approach? Any other idea?

blink1073:
> What are the already existing threads? 

It looks like the existing threads are heartbeat, iopub, and control.

Jupyter Client has threads for iopub, shell, stdin, control, and heartbeat.

I think it is worth trying to add a another thread, but maybe this is a good opportunity to add a performance regression check.  SciPy uses [airspeed velocity](https://github.com/scipy/scipy/blob/8a64c938ddf1ae4c02a08d2c5e38daeb8d061d38/doc/source/dev/contributor/benchmarking.rst#benchmarking-scipy-with-airspeed-velocity).

echarles:
> It looks like the existing threads are heartbeat, iopub, and control.

Thx @blink1073 Great to know ipykernel is already kind with kernel hosting. I guess an additional one `monitor`, or maybe `telemetry` (thinking about the discussion in jupyter-server https://github.com/jupyter-server/jupyter_server/issues/780 - maybe offtopic here, not sure?) could be added.

[...]

PR #937, opened 2022-05-23
Ensure psutil for the  process is accurate

echarles:
Fixes https://github.com/ipython/ipykernel/issues/912

This add to the control_thread a map of processes to allow the requesting the cpu_usage in a way psutil excepts (the cpu is only correct if is requested on the same process object).

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -957,13 +957,7 @@ class Kernel(SingletonConfigurable):
 
     def get_process_metric_value(self, process, name, attribute=None):
         try:
-            # psutil.Process methods will either return...
-            pid = process.pid
-            p = self.processes.get(pid, None)
-            if not p:
-                self.processes[pid] = process
-                p = self.processes.get(pid)
-            metric_value = getattr(p, name)()
+            metric_value = getattr(process, name)()
             if attribute is not None:  # ... a named tuple
                 return getattr(metric_value, attribute)
             else:  # ... or a number
@@ -979,15 +973,14 @@ class Kernel(SingletonConfigurable):
         all_processes = [current_process] + current_process.children(recursive=True)
         # Ensure 1) self.processes is updated to only current subprocesses
         # and 2) we reuse processes when possible (needed for accurate CPU)
-        self.processes = {process.pid: self.processes.get(process.pid, process) for process in all_processes}
         self.processes = {
             process.pid: self.processes.get(process.pid, process) for process in all_processes
         }
         reply_content["kernel_cpu"] = sum(
-            [self.get_process_metric_value(process, "cpu_percent", None) for process in all_processes]
+            [self.get_process_metric_value(process, "cpu_percent", None) for process in self.processes.values()]
         )
         reply_content["kernel_memory"] = sum(
-            [self.get_process_metric_value(process, "memory_info", "rss") for process in all_processes]
+            [self.get_process_metric_value(process, "memory_info", "rss") for process in self.processes.values()]
         )
         cpu_percent = psutil.cpu_percent()
         # https://psutil.readthedocs.io/en/latest/index.html?highlight=cpu#psutil.cpu_percent
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -977,6 +977,12 @@ class Kernel(SingletonConfigurable):
         reply_content = {"hostname": socket.gethostname(), "pid": os.getpid()}
         current_process = psutil.Process()
         all_processes = [current_process] + current_process.children(recursive=True)
+        pruned_processes = {}
+        for process in all_processes:
+            p = self.processes.get(process.pid)
+            if p is not None:
+                pruned_processes[process.pid] = p
+        self.processes = pruned_processes
         reply_content["kernel_cpu"] = sum(
             [self.get_process_metric_value(process, "cpu_percent", None) for process in all_processes]
         )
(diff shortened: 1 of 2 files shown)


ipykernel/kernelbase.py
```diff
@@ -76,6 +76,8 @@ class Kernel(SingletonConfigurable):
     # attribute to override with a GUI
     eventloop = Any(None)
 
+    processes = {}
```
mlucool, line 79: Why is this map needed?

mlucool: Maybe not a big deal, but this seems a bit "append" only (i.e. a process object is never removed)

echarles: > Maybe not a big deal, but this seems a bit "append" only (i.e. a process object is never removed)

https://github.com/ipython/ipykernel/pull/937/commits/17358fb26e2a362691c3f59e4aa0615d6abaf50c addresses this.

echarles:
> self.processes = {process.pid: self.processes.get(process.pid, process) for process in all_processes}

I have used that block code and the CPU does not return the correct value. 

...or should I also remove https://github.com/ipython/ipykernel/pull/937/files#diff-874f94d86eaa767fb34865311dd8fd8c0d9a2fa6c49bc21f28690058f3dfa83cR961 as you said ? (I am not sure to get the correct link, it points to just one line).

mlucool:
This works for me:
```
import subprocess
from time import sleep

import psutil


class Foo:
    processes = {}

    def update(self):
        current_process = psutil.Process()
        all_processes = [current_process] + current_process.children(recursive=True)
        self.processes = {
            process.pid: self.processes.get(process.pid, process)
            for process in all_processes
        }

    def print_cpu(self):
        print("-----")
        self.update()
        for p, v in self.processes.items():
            print(f"{p}: {v.cpu_percent()}")


def work():
    y = 0
    for x in range(10):
        y += x

    return y


f = Foo()
f.print_cpu()
work()
f.print_cpu()
# Let's use CPU
p = subprocess.Popen(
    ["dd", "if=/dev/zero", "of=/dev/null"],
    stdout=subprocess.PIPE,
    stderr=subprocess.STDOUT,
)
f.print_cpu()
sleep(0.1)
work()
f.print_cpu()
p.kill()
p.wait()
work()
sleep(2)
f.print_cpu()  # Less work
```

Output:
```
-----
24567: 0.0
-----
24567: 87.9
-----
24567: 65.0
127672: 0.0
-----
24567: 23.8
127672: 95.1
-----
24567: 1.5
```

Basically, if we are going to prune on every run, I think we should prune + keep the list up to date to limit the number of times we need to iterate these datastructures.

echarles:
@mlucool I have retested and now it work as expected on my env. Thx for the suggestion and better implementation. I have committed those changes.

Anything else before asking a review?

ipykernel/kernelbase.py
```diff
@@ -953,11 +955,15 @@ async def debug_request(self, stream, ident, parent):
...
-    # Taken from https://github.com/jupyter-server/jupyter-resource-usage/blob/e6ec53fa69fdb6de8e878974bcff006310658408/jupyter_resource_usage/metrics.py#L16
     def get_process_metric_value(self, process, name, attribute=None):
         try:
             # psutil.Process methods will either return...
-            metric_value = getattr(process, name)()
+            pid = process.pid
+            p = self.processes.get(pid, None)
+            if not p:
+                self.processes[pid] = process
+                p = self.processes.get(pid)
```
mlucool, line 965: These lines are not need as we ensure the a single, reused process below

mlucool (changes requested):

ipykernel/kernelbase.py
```diff
@@ -971,18 +977,23 @@ async def usage_request(self, stream, ident, parent):
         reply_content = {"hostname": socket.gethostname(), "pid": os.getpid()}
         current_process = psutil.Process()
         all_processes = [current_process] + current_process.children(recursive=True)
-        process_metric_value = self.get_process_metric_value
+        # Ensure 1) self.processes is updated to only current subprocesses
+        # and 2) we reuse processes when possible (needed for accurate CPU)
+        self.processes = {process.pid: self.processes.get(process.pid, process) for process in all_processes}
```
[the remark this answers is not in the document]
echarles: Removed. Thx.

ipykernel/kernelbase.py
echarles: Fixed. Thx.

ipykernel/kernelbase.py
```diff
@@ -971,18 +977,23 @@ async def usage_request(self, stream, ident, parent):
...
+        self.processes = {
+            process.pid: self.processes.get(process.pid, process) for process in all_processes
+        }
         reply_content["kernel_cpu"] = sum(
-            [process_metric_value(process, "cpu_percent", None) for process in all_processes]
+            [self.get_process_metric_value(process, "cpu_percent", None) for process in all_processes]
```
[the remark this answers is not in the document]
echarles: Fixed. Thx.

mlucool (approved):

blink1073:
Docs failure is unrelated, but the pre-commit ones are.

echarles:
I have fixed 2 format issues and added a type to the new `processes` class field. pre-commit is now happy, but the typed class field generates other errors, eg on https://github.com/ipython/ipykernel/runs/6805383473?check_suite_focus=true

```
    processes: dict[str, psutil.Process] = {}
E   TypeError: 'type' object is not subscriptable
```

blink1073:
Yeah you need to use `typings.Dict` before Python 3.8.

echarles:
@blink1073 CI is now green.

blink1073 (approved):
Thanks!

blink1073:
I'll cut a new release early next week.

echarles:
@blink1073 👍 

> I'll cut a new release early next week.

Super!

PR #943, opened 2022-05-30
Fix richInspectVariables

davidbrochart:
It seems that we can always use the interpreter to get the rich representation of the variable, whether we hit a breakpoint or not. The problem with the current situation where we hit a breakpoint and display an HTML representation of a variable is that the HTML is striped and we get something like `'text/html': '<div>\n<style scoped>...le>\n</div>'` instead of the full HTML. I can't find where it is striped, but if we don't need this branch anyway, can we just get rid of it?
@JohanMabille what do you think?
Also, @fcollonval might have some feedback since he [refactored that part](https://github.com/ipython/ipykernel/commit/73efd890b64145b0d28ea011df58e0745270458b#diff-a30b59f7d3dd6506cbed49059d190fec6a329aed08722be170c5c0dbacfae56dR530-R552).

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -26,10 +26,7 @@ try:
         SuspendedFramesManager,
         _FramesTracker,
     )
-    from _pydevd_bundle.pydevd_safe_repr import SafeRepr  # isort: skip
 
-    SafeRepr.maxstring_inner = 2**16
-    SafeRepr.maxother_inner = 2**16
     _is_debugpy_available = True
 except ImportError:
     _is_debugpy_available = False
@@ -598,7 +595,7 @@ class Debugger:
                     "type": "request",
                     "command": "evaluate",
                     "seq": seq + 1,
-                    "arguments": {"expression": code, "frameId": frame_id},
+                    "arguments": {"expression": code, "frameId": frame_id, "context": "clipboard"},
                 }
             )
             if reply["success"]:
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -595,7 +595,7 @@ class Debugger:
                     "type": "request",
                     "command": "evaluate",
                     "seq": seq + 1,
-                    "arguments": {"expression": code, "frameId": frame_id},
+                    "arguments": {"expression": code, "frameId": frame_id, "context": "clipboard"},
                 }
             )
             if reply["success"]:
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -26,7 +26,10 @@ try:
         SuspendedFramesManager,
         _FramesTracker,
     )
+    from _pydevd_bundle.pydevd_safe_repr import SafeRepr  # isort: skip
 
+    SafeRepr.maxstring_inner = 2**16
+    SafeRepr.maxother_inner = 2**16
     _is_debugpy_available = True
 except ImportError:
     _is_debugpy_available = False
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -595,7 +595,7 @@ class Debugger:
                     "type": "request",
                     "command": "evaluate",
                     "seq": seq + 1,
-                    "arguments": {"expression": code, "frameId": frame_id, "context": "clipboard"},
+                    "arguments": {"expression": code, "frameId": frame_id},
                 }
             )
             if reply["success"]:


fcollonval:
Does this mean the strip is related to the evaluate debug request?

davidbrochart:
Maybe, I can't find where it is striped. It is received already striped from the `ZMQStream`, but the `evaluate` debug request makes `ipython` generate a valid HTML. So it's somewhere after being sent on the ZMQ channel.

davidbrochart:
I created [a test](https://github.com/ipython/ipykernel/pull/944) which shows that the `text/plain` repr is striped.

davidbrochart:
To be more precise, we set [maxstring_inner](https://github.com/microsoft/debugpy/blob/5e80ce1f774d84878bbd97a879ec4e381e948526/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py#L23) to `2 ** 16` (like [maxstring_outer](https://github.com/microsoft/debugpy/blob/5e80ce1f774d84878bbd97a879ec4e381e948526/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py#L22])) and [maxother_inner](https://github.com/microsoft/debugpy/blob/5e80ce1f774d84878bbd97a879ec4e381e948526/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py#L63) to `2 ** 16` (like [maxother_outer](https://github.com/microsoft/debugpy/blob/5e80ce1f774d84878bbd97a879ec4e381e948526/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py#L62)). These values we set to `30` previously, which means that the maximum length of a representation was limited to 30 characters, having a head of `30 * 2 / 3 = 20` and a tail of `30 / 3 = 10`, and `...` in between to show that the repr was truncated. But in the debugger we don't want the representation to be truncated, especially if it's an HTML repr, since it would lead to invalid HTML.
Since these values are class attributes, it works for us but it could also impact every other user of `debugpy`, if it were used in the kernel, although it's not very likely. A cleaner solution would be to provide a way to configure this `SafeRepr` in `debugpy`/`pydevd`, but I don't know if it's (easily) feasible, any thought @fabioz?

fcollonval:
Thanks @davidbrochart for digging that up. I think a cleaner solution is to set `SafeRepr.raw_value = True` as other limiters exists too (like on collections):

https://github.com/microsoft/debugpy/blob/5e80ce1f774d84878bbd97a879ec4e381e948526/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_safe_repr.py#L268

To avoid side-effect for other use cases, could we apply the change in a transient fashion during the evaluation request?

davidbrochart:
Thanks for helping @fcollonval!
Actually setting `SafeRepr.raw_value = True` doesn't work because it doesn't escape `\`, and so it breaks [the evaluation](https://github.com/ipython/ipykernel/blob/71b926551f6bc725524a1160dad3411ec020ae39/ipykernel/debugger.py#L602).
But I found a [clean way to set the string length limits to a high value](https://github.com/microsoft/debugpy/blob/eaadb6c681d5375dae26093b93d239c019f0d2bc/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_comm.py#L1210-L1216) just for our request.

davidbrochart:
Thanks @fabioz, we want to send an `evaluate` request with some code that computes the representation of a variable in IPython. This representation can be e.g. a big HTML string, and it should not be truncated because we show it in JupyterLab.
I'm wondering if it's better to send `"context": "clipboard"` or `"format": {"rawString": True}`. @fcollonval mentioned that with the former, other limits might apply on e.g. collections, but with the latter we still need to escape `\` in the reply.

davidbrochart:
I'm not sure, I think we can get back bytes if the type is e.g. `image/png`, so I would have thought that we could use `rawString` and do the repr ourselves, since we `eval` after that:
```py
eval(repr(b"123"))
# b'123'

eval(repr("\n"))
# '\n'
```
But if bytes are decoded using `latin1` that's not going to work, right?

davidbrochart:
Anyways, I've tried it in JupyterLab with an image and it works fine with `"context": "clipboard"`, so let's merge this.
Thanks @flabioz!

blink1073:
I tried kicking the two failing builds, but they still failed

davidbrochart:
The last commit reverted the changes, we have the same failures so I guess it's not related. I'm going to re-apply them.

blink1073:
I just kicked https://github.com/ipython/ipykernel/actions/runs/2412265322 and it passed

davidbrochart:
Yeah, it seems to be a bit random, but https://github.com/ipython/ipykernel/pull/943/commits/448e16bbb8903540352ef1b03a4fd054271b84e6 is identical to `main` and it failed.

davidbrochart:
I re-launched the only failing build and it passed.

blink1073:
Great, thanks!

blink1073:
I can cut a release on Monday

PR #944, opened 2022-05-30
Add richInspectVariables test with HTML repr

davidbrochart:
@fcollonval In this test the `text/plain` repr is also striped.


Carreau:

PR #956, opened 2022-06-14
Fix compatibility with tornado 6.2 beta

minrk:
I _suspect_  this will reveal failures with the latest tornado 6.2 beta, which is mainly an update to try to deal with asyncio getting rid of the notion of a 'current' event loop that's not running.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
 
       - name: Install the Python dependencies
         run: |
-          pip install .[test] codecov
+          pip install --pre .[test] codecov
 
       - name: Install matplotlib
         if: ${{ !startsWith(matrix.os, 'macos') && !startsWith(matrix.python-version, 'pypy') }}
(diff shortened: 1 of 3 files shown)


blink1073:
We have an explicit job that tests [prereleases](https://github.com/ipython/ipykernel/blob/ba85b1c557be37d052256333c07abde764bcd60d/.github/workflows/ci.yml#L178), and it has been failing since the new tornado beta came out.  The intent was to contain the failure to one job.

minrk:
Good plan! I'll rebase to just the fix tomorrow, unless you want to do it first.

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
       - name: Install the Python dependencies
         run: |
-          pip install .[test] codecov
+          pip install --pre .[test] codecov
```
blink1073, line 40:
```suggestion
          pip install .[test] codecov
```

blink1073 (approved):
LGTM, thank you!

PR #957, opened 2022-06-15
Explicitly require pyzmq >= 17

minrk:
pyzmq is pulled transitively from jupyter-client, so this will have little to no effect in practice, but it is a direct dependency here.

pyzmq 17 is the version that introduces support for running directly with tornado

it's possible an even more recent version is required


blink1073 (approved):
Thanks!

PR #973, opened 2022-08-01
Extract the Comm Python package

martinRenou:
~~Currently working on a Python comm package on https://github.com/martinRenou/comm for now, but it should probably be moved in the IPython org~~
This PR relies on the new comm package https://github.com/ipython/comm

This has the benefit of being able to:
- remove the ipykernel dependency in ipywidgets https://github.com/jupyter-widgets/ipywidgets/pull/3533
- remove the ipykernel monkey patch in xeus-python https://github.com/jupyter-xeus/xeus-python/pull/548
- remove the ipykernel package mock in jupyterlite's pyolite 
- remove the ipykernel package mock in emscripten-forge


pyproject.toml
```diff
@@ -27,6 +27,7 @@ requires-python = ">=3.8"
 dependencies = [
     "debugpy>=1.0",
     "ipython>=7.23.1",
+    "comm",
```
blink1073, line 30: We should add a `>=0.1` here so it gets picked up by our minimum versions check.

martinRenou: Thanks for the review! I will update this

martinRenou:
I marked the PR as ready to review! 

I might be wrong but the ipyparallel issue seems unrelated?

```
  =========================== short test summary info ============================
  ERROR  - DeprecationWarning: Jupyter is migrating its paths to use standard p...
  !!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
  =============================== 1 error in 0.90s ===============================
```

blink1073:
> I might be wrong but the ipyparallel issue seems unrelated?

Correct

blink1073 (approved):
Thanks!

blink1073:
I'll cut a minor release with this change early next week.

blink1073:
Yes, the equivalent classes are exported from the `comm` module.

maartenbreddels:
This broke compatibility in Solara, since I was doing 
 * https://github.com/widgetti/solara/blob/c41b87089164a4761ffc925a015a79072054ec0a/solara/server/kernel.py#L163

Giving me:
```
>       self.comm_manager = CommManager(parent=self, kernel=self)
E       TypeError: __init__() got an unexpected keyword argument 'parent'
```

Not sure if this should be considered a public API (the constructor), if so, happy to open an issue.

maartenbreddels:
@martinRenou really nice to see this out this soon btw :) congrats

blink1073:
I think the cause is the same as #1026

PR #984, opened 2022-09-01
Add python logo in svg format

steff456:
Hi!

This PR adds the python logo in svg format in the resources folder and it is related to https://github.com/jupyterlab/jupyterlab/issues/12997 where we want to fix the blurry icon present when a lot of zoom is made in jupyterlab. 

The asset was downloaded from the official assets of the Python Software Foundation available [here.](https://www.python.org/community/logos/)


blink1073 (approved):
Thank you!

PR #990, opened 2022-09-13
PR: Close memory leak

impact27:
Every time a message is recieved on ipykernel while qt5 is activated, a `QTimer` and a `QSocketNotifier` leak. This closes the memory leak.

Fixes https://github.com/spyder-ide/spyder/issues/19026


blink1073 (approved):
Thank you!

PR #1012, opened 2022-10-30
Enable webagg in %matplotlib

zhizheng1:
webagg is enabled in %matplotlib magic in the latest IPython 8.6.0 via this PR:
https://github.com/ipython/ipython/pull/13778

However this trivial change is needed for it to work in an ipykernel environment, otherwise an exception is thrown.


blink1073 (approved):
Thank you!

PR #1019, opened 2022-11-09
Ignore the new Jupyter_core deprecation warning in CI

jasongrout:
See https://github.com/jupyter/jupyter_core/issues/310


blink1073 (approved):
Thanks!

PR #1037, opened 2022-11-28
Configurables needs to be configurable

Carreau:
IPython rely on configurables being configurables to ... configure things.

Comm seem to insert or return a non-configurable in IPython, that make ipykernel crash in some instances. For example but not limited to:

    %config Completer.use_jedi = False

Or trying to use hvplot for example.

This is just a "this is wrong" PR, there should be a proper fix, and I will likely add a hard check in IPython that what is added in configurables is actually a configurables.

This will need a fix in a Comm pacakge.

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -55,13 +55,13 @@ def _create_comm(*args, **kwargs):
     return BaseComm(*args, **kwargs)
 
 
-def _create_manager(*args, **kwargs):
+def _get_comm_manager(*args, **kwargs):
     """Create a new CommManager."""
     return CommManager(*args, **kwargs)
 
 
 comm.create_comm = _create_comm
-comm.create_manager = _create_manager
+comm.get_comm_manager = _get_comm_manager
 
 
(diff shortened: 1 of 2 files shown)


Carreau:
cc @fperez

blink1073:
Thanks for the report!  I'll make a new release with this tomorrow.

fperez:
Thx a lot @blink1073!! 

cc @yuvipanda @tsnow03 @weiji14 - we might be able to release the ipykernel pin on our cryo env soon.

blink1073:
I already yanked the offending version, you should be able to remove the pin now

fperez:
Ah, got it, thx!

weiji14:
Looks like [`ipykernel=6.18.1` on conda-forge](https://anaconda.org/conda-forge/ipykernel/files?version=6.18.1) isn't marked as broken yet, so maybe keep the pin around for a little longer :slightly_smiling_face:

blink1073:
https://github.com/ipython/ipykernel/releases/tag/v6.18.2

PR #1047, opened 2022-12-07
Use pydevd ipython hook

### LINKED ISSUE #1004: When importing `debugpy`, `PYDEVD_IPYTHON_COMPATIBLE_DEBUGGING=1` should be set
A new flag is available in `pydevd` which allows a user to step into library code without getting inside of IPython internals while doing so.

To make use of this, `PYDEVD_IPYTHON_COMPATIBLE_DEBUGGING` must be set to `1` in the `os.environ` prior to importing `debugpy`.

See: https://github.com/microsoft/debugpy/issues/869 for more information.

vidartf:
This enables pydevd/debugpy to use IPython specific frame filtering for stack requests etc.

Closes #1004.

## TODO:
- [ ] Verify behavior
- [ ] Needs some unittests

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -18,6 +18,8 @@ except ImportError:
 from .compiler import get_file_name, get_tmp_directory, get_tmp_hash_seed
 
 try:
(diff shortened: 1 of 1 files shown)


Carreau:

PR #1051, opened 2022-12-08
fix: too many arguments dropped when passing to base comm constructor

### LINKED ISSUE #1050: Latest version break IPyWidgets
IPyKernel Version: 6.19.0
Install Jupyter and ipywidgets 
Run a notebook with the following code

```python
import ipywidgets as widgets
widgets.IntSlider()
```

Slider is not renderered
Here is the error message
```
A Jupyter widget could not be displayed because the widget state could not be found. 
This could happen if the kernel storing the widget is no longer available, or if the widget state was not saved in the notebook. 
You may be able to create the widget by running the appropriate cells.
```

maartenbreddels:
Bug introduced in fee98e5, covered by test now.

Closes #1050

### DIFF
--- a/ipykernel/comm/comm.py
+++ b/ipykernel/comm/comm.py
@@ -72,9 +72,7 @@ class Comm(traitlets.config.LoggingConfigurable, BaseComm):
         # Comm takes positional arguments, LoggingConfigurable does not, so we explicitly forward arguments
         traitlets.config.LoggingConfigurable.__init__(self, **kwargs)
         # drop arguments not in BaseComm
-        for name in ["kernel"]:
-            if name in kwargs:
-                kwargs.pop(name)
+        kwargs.pop("kernel", None)
         BaseComm.__init__(self, *args, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


ipykernel/comm/comm.py
```diff
@@ -71,7 +71,8 @@ def _default_comm_id(self):
     def __init__(self, *args, **kwargs):
         # Comm takes positional arguments, LoggingConfigurable does not, so we explicitly forward arguments
         traitlets.config.LoggingConfigurable.__init__(self, **kwargs)
-        for name in self.trait_names():
+        # drop arguments not in BaseComm
+        for name in ["kernel"]:
             if name in kwargs:
                 kwargs.pop(name)
```
martinRenou, line 77:
```suggestion
        # drop arguments not in BaseComm
        kwargs.pop("kernel", None)
```

martinRenou (commented):
Just a minor comment

blink1073 (approved):
Thanks!  I don't think you need to eat your laptop, this was my mistake @maartenbreddels.  😅.  I yanked 6.19.0.

PR #1053, opened 2022-12-08
Fix error in `%edit` magic

ccordoba12:
- The problem is the `_find_edit_target` method is a `staticmethod` in IPython.
- We have a test in Spyder for this magic, which detected that it's broken with `6.19.1`.
- This regression was introduced in #1034, according to `git blame`.

### DIFF
--- a/ipykernel/tests/test_zmq_shell.py
+++ b/ipykernel/tests/test_zmq_shell.py
@@ -217,13 +217,9 @@ def test_magics(tmp_path):
     shell.user_ns["hi"] = 1
     magics = KernelMagics(shell)
 
-    def find_edit_target(*args):
-        return str(tmp_path), 0, 1
-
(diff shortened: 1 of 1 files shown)


blink1073:
@ccordoba12, you can remove the edit magic test for now.  The proper fix would be to use `mock.patch`, I thought what I had done was a proper work-around.

ccordoba12:
I made a simple fix for now, just to check that it's working. Should I remove that instead?

blink1073:
> I made a simple fix for now, just to check that it's working. Should I remove that instead?

That sounds good, let's keep it in.  Thank you!

blink1073:
https://github.com/ipython/ipykernel/releases/tag/v6.19.2

ccordoba12:
Thanks @blink1073 for your prompt response!

PR #782, opened 2021-10-07
ENH: add support for qt6

tacaswell:
Addresses #775

Issue #775, opened 2021-09-21
support for %gui qt6 


daliagachc:
Ipython already provides support for qt6. for example running this in ipython works: 
```
%gui qt6
```

however when running the same using the ipykernel i get
```
%gui qt6

ERROR:root:Invalid GUI request 'qt6', valid ones are:dict_keys(['inline', 'nbagg', 'notebook', 'ipympl', 'widget', None, 'qt4', 'qt', 'qt5', 'wx', 'tk', 'gtk', 'gtk3', 'osx', 'asyncio'])
```

blink1073:
Hi @daliagachc, if you'd like to try a fix and verify it in your environment, it should be a matter of adding a `loop_qt6` function and exit handler similar to what is done for qt5 [here](https://github.com/ipython/ipykernel/blob/cf0ef994cfd1847323add274ab28aaa959e7d41f/ipykernel/eventloops.py#L123).

Carreau:
+1, there was https://github.com/ipython/ipython/pull/13085 on the IPython side. 

If Qt6 is installed it looks like matplotlib might be automatically trying to set qt6 up. We have test errors on napari with install --pre and matplotlib:

```
/tmp/.tox/py38-linux-pyside/lib/python3.8/site-packages/matplotlib/pyplot.py:2467: in <module>
      install_repl_displayhook()
  /tmp/.tox/py38-linux-pyside/lib/python3.8/site-packages/matplotlib/pyplot.py:138: in install_repl_displayhook
      ip.enable_gui(ipython_gui_name)
  /tmp/.tox/py38-linux-pyside/lib/python3.8/site-packages/ipykernel/inprocess/ipkernel.py:181: in enable_gui
      enable_gui(gui, kernel=self.kernel)
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  gui = 'qt6'
  kernel = <ipykernel.inprocess.ipkernel.InProcessKernel object at 0x7fcb20138070>
  
      def enable_gui(gui, kernel=None):
          """Enable integration with a given GUI"""
          if gui not in loop_map:
              e = "Invalid GUI request %r, valid ones are:%s" % (gui, loop_map.keys())
  >           raise ValueError(e)
[...]

PR #1054, opened 2022-12-08
ENH: add `%gui` support for Qt6

shaperilio:
Addresses #775 

1. Support for Qt 4, 5, and 6, via `PyQt` or `PySide`. Replicated logic from `IPython` (e.g. prefer `PyQt` over `PySide`)
2. Put Qt importing at the level of `enable_gui` so that any import errors show up in the client as opposed to in the kernel's console.
3. Added checks for more helpful error messages / warnings when you can't / aren't going to get the version you're requesting.
4. Included relevant parts of @tacaswell's work in #782

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -90,7 +90,6 @@ def _notify_stream_qt(kernel):
         # if there were any, wake it up
         if kernel.shell_stream.flush(limit=1):
             kernel._qt_notifier.setEnabled(False)
-            print('ZMQ events to process; quitting Qt event loop.')
             kernel.app.qt_event_loop.quit()
 
     if not hasattr(kernel, "_qt_notifier"):
@@ -118,21 +117,18 @@ def _notify_stream_qt(kernel):
 @register_integration("qt", "qt4", "qt5", "qt6")
 def loop_qt(kernel):
     """Event loop for all versions of Qt."""
-    print(f'Starting Qt loop with {os.environ.get("QT_API", None)=}')
     _notify_stream_qt(kernel) # install hook to stop event loop.
     # Start the event loop.
     kernel.app._in_event_loop = True
     # `exec` blocks until there's ZMQ activity.
     el = kernel.app.qt_event_loop  # for brevity
     el.exec() if hasattr(el, 'exec') else el.exec_()
-    print('Qt loop exited.')
     kernel.app._in_event_loop = False
 
 
 # exit and watch are the same for qt 4 and 5
 @loop_qt.exit
 def loop_qt_exit(kernel):
-    print('Request Qt app exit (will close all open windows).')
     kernel.app.exit()
 
 
@@ -480,13 +476,11 @@ def set_qt_api(gui, kernel):
                     os.environ["QT_API"] = "pyqt5"
         elif gui == 'qt6':
             try:
-                print('Trying to import PyQt6...')
                 import PyQt6  # noqa
 
                 os.environ["QT_API"] = "pyqt6"
             except ImportError:
                 try:
-                    print('Trying to import PySide6...')
                     import PySide6  # noqa
 
                     os.environ["QT_API"] = "pyside6"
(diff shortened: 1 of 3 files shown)


shaperilio:
In case it helps, I've published a [notebook I used to test functionality](https://gist.github.com/shaperilio/f10272fc2c205c6b25a0611baeeafedb). It would be nice to turn these into unit tests, but I'm afraid setting those up is over my head. I'd be happy to take a stab at it if someone can point me in the right direction.

shaperilio:
I'm playing with this in Linux now and have become aware that `IPython` is not quite there. I guess I got mixed up yesterday while working on `ipykernel`. See [this issue](https://github.com/ipython/ipython/issues/13859).

Effectively, the logic here is not in `IPython`, i.e. if you request `%gui qt6` in straight `IPython`, it will  never try to use `PySide6`. Since `ipykernel` defers to `IPython` to import Qt, I'm thinking the logic should be all in one place (and probably `IPython`). Any thoughts?

tacaswell:
https://github.com/matplotlib/matplotlib/blob/1d2abe7ab909ebca32908694fc3e3828c37cf11d/lib/matplotlib/tests/test_backends_interactive.py#L24-L68 this and `matplotlib.testing.subprocess_run_helper` may be a useful pattern to copy.

shaperilio:
I wrote a bit of a unit test, but it won't run unless we make the CI install a couple of versions of either PyQt or PySide. How do I do that / do we want that?

blink1073:
We can create a coverage matrix:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 9904b54..05be2c2 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ cov = [
   "curio",
   "trio",
 ]
+pyqt5 = ["pyqt5"]

 [tool.hatch.version]
 path = "ipykernel/_version.py"
@@ -92,6 +93,15 @@ features = ["test", "cov"]
 test = "python -m pytest -vv --cov ipykernel --cov-branch --cov-report term-missing:skip-covered {args}"
 nowarn = "test -W default {args}"

+[[tool.hatch.envs.cov.matrix]]
+qt = ["qt5", "qt6"]
+
+[tool.hatch.envs.cov.overrides]
+matrix.qt.features = [
+  { value = "pyqt5", if = ["qt5"] },
+  { value = "pyqt6", if = ["qt6"] },
+]
+
 [tool.hatch.envs.typing]
 features = ["test"]
 dependencies = ["mypy>=0.990"]
```

So when you call `hatch run cov:test`, it would run with each variant.

shaperilio:
OK, I think this is ready! I've put code in [this pull request to IPython](https://github.com/ipython/ipython/pull/13864/) which replicates the behavior here.

The only difference is that, in IPython, 'qt4' is an alias to 'qt', so you can never explicitly get a Qt4 / PySide event loop there. That's not the case here. I left the behavior in IPython because I don't know the history of that alias.

shaperilio:
@blink1073 sorry about that - I had removed the windows skips because those tests passed fine on my windows machine.

blink1073:
> I had removed the windows skips because those tests passed fine on my windows machine.

No worries!  Not sure why they hang in CI then, but things like that happen for other platforms too.  :smile:

blink1073 (approved):
This is great, thanks!  I'll leave it open for a few days in case anyone else gets a chance to review.

ccordoba12:
@blink1073, this PR broke the `%matplotlib qt ` magic in Spyder (detected by one of our Spyder tests). So, could you yank it on PyPI? I'm still investigating how to fix it (along with other problems I found with this approach).

blink1073:
Yanked

ccordoba12:
Thanks @blink1073! And sorry for not reviewing it on time, I'm finishing a Spyder release right now.

blink1073:
I'd love to add spyder as a downstream test, how feasible do you think that is?

ccordoba12:
That'd be great! Thanks for the offer. I think adding our Spyder-kernels test suite shouldn't be that hard.

As for the Spyder tests (which also exercise our kernel under different scenarios), we could only run a file called `test_ipythonconsole.py` (I could give you a hand with that one because it's harder to setup).

ccordoba12:
BTW, I found a simple to solve the regressions introduced by this PR. I hope to submit a fix in the next couple of days, after testing that change against our two test suites.

blink1073:
Great, thanks!

shaperilio:
Thanks y'all; @ccordoba12, let me know if I can be of any help.

PR #1067, opened 2022-12-29
Add api docs

blink1073:


blink1073:
Thanks for the prompt!  They are available at https://ipykernel.readthedocs.io/en/latest/?badge=latest.  I'll update the badge and the link on the main page.

PR #1071, opened 2023-01-03
Don't raise error when trying to create another Qt app for Qt eventloop

ccordoba12:
- This is a follow-up PR that fixes the `%matplotlib qt` magic in Spyder.
- Avoid raising `RuntimeError('Kernel already running a Qt event loop.')` in `make_qt_app_for_kernel`. This was the cause of the mentioned failure and it also generated an unnecessary traceback when trying to set another backend:

    imagen

    Instead, replace that raise with a return.

- Remove the Qt4 eventloop because it's ancient and not supported anymore.
- Restore `loop_qt5` function (and make it equal to `loop_qt`) because it's been public API for years and we're using it in Spyder-kernels.
- Avoid raising errors when calling `set_qt_api_env_from_gui` because they are not displayed to clients and instead use prints. Now you get this, for instance:

    imagen

- Do some cleanup after PR #1054.

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -117,10 +117,12 @@ def _notify_stream_qt(kernel):
 
 @register_integration("qt", "qt5", "qt6")
 def loop_qt(kernel):
-    """Event loop for all versions of Qt."""
+    """Event loop for all supported versions of Qt."""
     _notify_stream_qt(kernel)  # install hook to stop event loop.
+
     # Start the event loop.
     kernel.app._in_event_loop = True
+
     # `exec` blocks until there's ZMQ activity.
     el = kernel.app.qt_event_loop  # for brevity
     el.exec() if hasattr(el, 'exec') else el.exec_()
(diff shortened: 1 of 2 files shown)


ccordoba12:
@shaperilio,, this should be ready for review.

@blink1073, I think the failure in the ipyparallel test suite is unrelated to this. Also, please wait until I can run Spyder tests with this PR before merging it (I'll give you a ping when I have the chance to do that).

blink1073:
I agree the `ipyparallel` failure is unrelated.  I'll fix the `check release` failure separately, and will wait to hear from you before merging this one.  Cheers!

blink1073:
@ccordoba12, fyi I rebased to fix merge conflicts picked up from #1073

ccordoba12:
Thanks but your merge reverted some of the changes I made that fix the Spyder bug and improve other things. So, I need to drop it and do a merge that preserves what I did before.

blink1073:
Ugh, sorry about that.

ccordoba12:
No worries, I understand you wanted to check if everything was green here too after PR #1073.

ccordoba12:
@blink1073, I still don't understand the failure in IPyparallel, but the rest seems ok now.

blink1073:
I got the same `ipyparallel` error in #1073, I'm not sure what is going on there.

shaperilio:
Shoudl we also remove Qt4 from [ipython](https://github.com/ipython/ipython/pull/13864)?

ccordoba12:
Yeah, I think that's a good idea. PyQt4 is really outdated and it doesn't have wheels for the most recent Python versions.

Issue #881, opened 2022-03-14
Cannot interrupt top-level coroutine


davidbrochart:
It looks like interrupting a cell execution which consists of awaiting a top-level coroutine, e.g.:
```python
import asyncio

async def main():
    await asyncio.sleep(5)

await main()
```
doesn't take effect immediately, but once the coroutine completes execution.

See #880 for a test that reveals this bug.

davidbrochart:
This also happens with:

```py
import asyncio

await asyncio.sleep(5)
```

Which means that [this test](https://github.com/ipython/ipykernel/blob/4dc303364ca73b5185685807de41c7fe4b4a4dc7/ipykernel/tests/test_async.py#L32) is incomplete.

PR #1079, opened 2023-01-25
Replace Tornado with AnyIO

### LINKED ISSUE #656: Dropping the dependency to tornado

davidbrochart:
## References

Supersedes #876.
Closes #656.

IPykernel should not depend on a web server. It uses [Tornado](https://www.tornadoweb.org) as an async framework, but modern Python has better solutions, including standard `asyncio`. Tornado's use in ipykernel involves using a lot of queues and creates complexity that makes maintenance or improvements to the code base harder.
Structured concurrency as introduced by [Trio](https://trio.readthedocs.io) is gaining a lot of traction. [AnyIO](https://anyio.readthedocs.io) brings structured concurrency on top of `asyncio` and acts as an abstraction layer which allows to choose a concurrency backend, such as `asyncio` or `trio`. Major projects have switched to AnyIO, including [FastAPI](https://fastapi.tiangolo.com).

This PR is still a work in progress. I have disabled some tests, like the ones for in-process kernels. Also, I think we should re-think event-loop integration (maybe using [Trio's approach](https://trio.readthedocs.io/en/stable/reference-lowlevel.html#using-guest-mode-to-run-trio-on-top-of-other-event-loops)). But I would like to get feedback from the community and discuss how we can move forward with these changes.

cc @SylvainCorlay @JohanMabille @minrk @ccordoba12 

## Code changes

The dependency to Tornado is dropped, in favor of AnyIO. This means that instead of using [ZMQStream](https://pyzmq.readthedocs.io/en/latest/api/zmq.eventloop.zmqstream.html) with `on_recv` callbacks, we use async/await everywhere.
Using AnyIO in the main thread event loop gives us [Trio](https://trio.readthedocs.io) support for free in the user code. It is currently disabled until we get [AnyIO support in pyzmq](https://github.com/zeromq/pyzmq/issues/1827), or decide to process shell messages in a separate thread, which [might be needed for sub-shells anyway](https://github.com/jupyter/enhancement-proposals/pull/91).

## User-facing changes

The goal is to have no user-facing change. However, I have identified the following ones:
- handling of standard streams (`stdout`, `stderr`) is improved as there is no delay anymore.
- interrupting async user code now works (see #881).
- traceback when interrupting blocking user code leaks one frame into the kernel source code (might be fixable).

## Backwards-incompatible changes

- The stream traits (e.g. [shell_stream](https://github.com/ipython/ipykernel/blob/4f0e252d745c44aa3a3109c8274e5014a512f462/ipykernel/kernelbase.py#L91)) are replaced with their socket equivalent (e.g. `shell_socket`).
- The API is now mostly `async`, including the `Kernel.start` method.
- There is a new blocking `Kernel.stop` (thread-safe) method.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -5,11 +5,13 @@
 from __future__ import annotations
 
 import asyncio
+import concurrent.futures
 import inspect
 import itertools
 import logging
 import os
 import queue
+import socket
 import sys
 import threading
 import time
@@ -17,6 +19,7 @@ import typing as t
 import uuid
 import warnings
 from datetime import datetime
+from queue import Queue
 from signal import SIGINT, SIGTERM, Signals
 
 from .control import CONTROL_THREAD_NAME
@@ -52,6 +55,7 @@ from traitlets.traitlets import (
     Set,
     Unicode,
     default,
+    observe,
 )
 
 from ipykernel.jsonutil import json_clean
@@ -326,43 +330,14 @@ class Kernel(SingletonConfigurable):
         control_loop.add_callback(_flush)
         return awaitable_future
 
-    async def process_control(self, msg):
-        """dispatch control requests"""
-        if not self.session:
-            return
-        idents, msg = self.session.feed_identities(msg, copy=False)
+    async def process_control(self):
         try:
-            msg = self.session.deserialize(msg, content=True, copy=False)
-        except Exception:
-            self.log.error("Invalid Control Message", exc_info=True)  # noqa: G201
-            return
-
-        self.log.debug("Control received: %s", msg)
-
-        # Set the parent message for side effects.
-        self.set_parent(idents, msg, channel="control")
-        self._publish_status("busy", "control")
-
-        header = msg["header"]
-        msg_type = header["msg_type"]
-
-        handler = self.control_handlers.get(msg_type, None)
-        if handler is None:
-            self.log.error("UNKNOWN CONTROL MESSAGE TYPE: %r", msg_type)
-        else:
-            try:
-                result = handler(self.control_stream, idents, msg)
-                if inspect.isawaitable(result):
-                    await result
-            except Exception:
-                self.log.error("Exception in control handler:", exc_info=True)  # noqa: G201
-
-        sys.stdout.flush()
-        sys.stderr.flush()
-        self._publish_status("idle", "control")
-        # flush to ensure reply is sent
-        if self.control_stream:
-            self.control_stream.flush(zmq.POLLOUT)
+            while True:
+                await self.process_control_message()
+        except BaseException as e:
+            if self.control_stop.is_set():
+                return
+            raise e
 
     async def should_handle(self, stream, msg, idents):
         """Check whether a shell-channel message should be handled
@@ -455,7 +430,7 @@ class Kernel(SingletonConfigurable):
         try:
             msg = self.session.deserialize(msg, content=True, copy=True)
         except BaseException:
-            self.log.error("Invalid Message", exc_info=True)
+            self.log.error("Invalid Message", exc_info=True)  # noqa: G201
             return
 
         # Set the parent message for side effects.
@@ -521,30 +496,6 @@ class Kernel(SingletonConfigurable):
             await to_thread.run_sync(self.control_stop.wait)
             tg.cancel_scope.cancel()
 
-    async def process_control(self):
-        try:
-            while True:
-                await self.process_control_message()
-        except BaseException:
-            if self.control_stop.is_set():
-                return
-            self.log.debug("Advancing eventloop %s", eventloop)
-            try:
-                eventloop(self)
-            except KeyboardInterrupt:
-                # Ctrl-C shouldn't crash the kernel
-                self.log.error("KeyboardInterrupt caught in kernel")
-            if self.eventloop is eventloop:
-                # schedule advance again
-                schedule_next()
-
-        def schedule_next():
-            """Schedule the next advance of the eventloop"""
-            # flush the eventloop every so often,
-            # giving us a chance to handle messages in the meantime
-            self.log.debug("Scheduling eventloop advance")
-            self.io_loop.call_later(0.001, advance_eventloop)
-
     async def process_control_message(self, msg=None):
         assert self.control_socket is not None
         assert self.session is not None
@@ -554,7 +505,7 @@ class Kernel(SingletonConfigurable):
         try:
             msg = self.session.deserialize(msg, content=True, copy=True)
         except Exception:
-            self.log.error("Invalid Control Message", exc_info=True)
+            self.log.error("Invalid Control Message", exc_info=True)  # noqa: G201
             return
 
         self.log.debug("Control received: %s", msg)
@@ -571,10 +522,15 @@ class Kernel(SingletonConfigurable):
             self.log.error("UNKNOWN CONTROL MESSAGE TYPE: %r", msg_type)
         else:
             try:
-                t, dispatch, args = self.msg_queue.get_nowait()
-            except (asyncio.QueueEmpty, QueueEmpty):
-                return
-        await dispatch(*args)
+                result = handler(self.control_socket, idents, msg)
+                if inspect.isawaitable(result):
+                    await result
+            except Exception:
+                self.log.error("Exception in control handler:", exc_info=True)  # noqa: G201
+
+        sys.stdout.flush()
+        sys.stderr.flush()
+        self._publish_status("idle", "control")
 
     async def dispatch_queue(self):
         """Coroutine to preserve order of message handling
@@ -587,7 +543,7 @@ class Kernel(SingletonConfigurable):
             try:
                 await self.process_one()
             except Exception:
-                self.log.error("Exception in control handler:", exc_info=True)
+                self.log.error("Exception in control handler:", exc_info=True)  # noqa: G201
 
         sys.stdout.flush()
         sys.stderr.flush()
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -5,7 +5,6 @@
 from __future__ import annotations
 
 import asyncio
-import concurrent.futures
 import inspect
 import itertools
 import logging
@@ -138,7 +137,7 @@ class Kernel(SingletonConfigurable):
             self.shell_stream = change.new[0]
 
     control_socket = Instance(zmq.asyncio.Socket, allow_none=True)
-    control_tasks = List()
+    control_tasks: t.Any = List()
 
     debug_shell_socket = Any()
 
@@ -297,48 +296,6 @@ class Kernel(SingletonConfigurable):
     def dispatch_control(self, msg):
         self.control_queue.put_nowait(msg)
 
-    async def poll_control_queue(self):
-        while True:
-            msg = await self.control_queue.get()
-            # handle tracers from _flush_control_queue
-            if isinstance(msg, (concurrent.futures.Future, asyncio.Future)):
-                msg.set_result(None)
-                continue
-            await self.process_control(msg)
-
-    async def _flush_control_queue(self):
-        """Flush the control queue, wait for processing of any pending messages"""
-        tracer_future: concurrent.futures.Future[object] | asyncio.Future[object]
-        if self.control_thread:
-            control_loop = self.control_thread.io_loop
-            # concurrent.futures.Futures are threadsafe
-            # and can be used to await across threads
-            tracer_future = concurrent.futures.Future()
-            awaitable_future = asyncio.wrap_future(tracer_future)
-        else:
-            control_loop = self.io_loop
-            tracer_future = awaitable_future = asyncio.Future()
-
-        def _flush():
-            # control_stream.flush puts messages on the queue
-            if self.control_stream:
-                self.control_stream.flush()
-            # put Future on the queue after all of those,
-            # so we can wait for all queued messages to be processed
-            self.control_queue.put(tracer_future)
-
-        control_loop.add_callback(_flush)
-        return awaitable_future
-
-    async def process_control(self):
-        try:
-            while True:
-                await self.process_control_message()
-        except BaseException as e:
-            if self.control_stop.is_set():
-                return
-            raise e
-
     async def should_handle(self, stream, msg, idents):
         """Check whether a shell-channel message should be handled
 
@@ -496,6 +453,15 @@ class Kernel(SingletonConfigurable):
             await to_thread.run_sync(self.control_stop.wait)
             tg.cancel_scope.cancel()
 
+    async def process_control(self):
+        try:
+            while True:
+                await self.process_control_message()
+        except BaseException as e:
+            if self.control_stop.is_set():
+                return
+            raise e
+
     async def process_control_message(self, msg=None):
         assert self.control_socket is not None
         assert self.session is not None
@@ -525,23 +491,6 @@ class Kernel(SingletonConfigurable):
                 result = handler(self.control_socket, idents, msg)
                 if inspect.isawaitable(result):
                     await result
-            except Exception:
-                self.log.error("Exception in control handler:", exc_info=True)  # noqa: G201
-
-        sys.stdout.flush()
-        sys.stderr.flush()
-        self._publish_status("idle", "control")
-
-    async def dispatch_queue(self):
-        """Coroutine to preserve order of message handling
-
-        Ensures that only one message is processing at a time,
-        even when the handler is async
-        """
-
-        while True:
-            try:
-                await self.process_one()
             except Exception:
                 self.log.error("Exception in control handler:", exc_info=True)  # noqa: G201
 
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -17,8 +17,8 @@ import typing as t
 import uuid
 import warnings
 from datetime import datetime
-from queue import Queue
-from signal import SIGINT, SIGTERM, Signals
+from functools import partial
+from signal import SIGINT, SIGTERM, Signals, signal
 
 from .control import CONTROL_THREAD_NAME
 
@@ -333,7 +333,7 @@ class Kernel(SingletonConfigurable):
         if self.control_stream:
             self.control_stream.flush(zmq.POLLOUT)
 
-    def should_handle(self, stream, msg, idents):
+    async def should_handle(self, stream, msg, idents):
         """Check whether a shell-channel message should be handled
 
         Allows subclasses to prevent handling of certain messages (e.g. aborted requests).
@@ -349,9 +349,6 @@ class Kernel(SingletonConfigurable):
     def pre_handler_hook(self):
         """Hook to execute before calling message handler"""
         # ensure default_int_handler during handler call
-
-    def post_handler_hook(self):
-        """Hook to execute after calling message handler"""
 
     async def enter_eventloop(self):
         """enter eventloop"""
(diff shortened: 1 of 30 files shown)


maartenbreddels:
+1 on AnyIO, they seem to have all the missing primitives needed for doing asyncio+threaded work, seems like a solid project.

dhirschfeld:
As a `trio` user I'm +:100: for using AnyIO :heart:

agronholm:
I must point out that using blocking portals in this manner will block the original event loop, preventing any other tasks from running until the blocking call is complete. As a former Java programmer, I recall that Java used a similar approach to solve this problem, but there the nested event loop was able to process events from the parent loop. What should also be taken into account is any Protocols and tasks launched from the nested loop, which will also stop functioning once the blocking call completes.

davidbrochart:
Thanks for the feedback @agronholm.
Actually this PR doesn't use [blocking portals](https://anyio.readthedocs.io/en/stable/threads.html#calling-asynchronous-code-from-an-external-thread). Instead, I'm running a new event loop in a separate thread (for the [control thread](https://github.com/ipython/ipykernel/blob/83f7d6cea4791a04c17e5197aa8ece3749f77180/ipykernel/control.py)). So there is no nested loops, just two separate (parallel) loops, each in their own thread. There shouldn't be any issue with that, right? Except if there is a global state in AnyIO's event loop, which I doubt.

agronholm:
> I'm running a new event loop in a separate thread (for the [control thread](https://github.com/ipython/ipykernel/blob/83f7d6cea4791a04c17e5197aa8ece3749f77180/ipykernel/control.py)). So there is no nested loops, just two separate (parallel) loops, each in their own thread. There shouldn't be any issue with that, right?

Yeah, sorry, it's a rather large and complex PR so I haven't read through it, so I just made assumptions based on your other posts. So are you creating one-off event loops for running async callables from sync functions? Or do you have a lingering, shared event loop thread running in the background? Either way, thread safety is the biggest problem. One of the main selling points of async is that you can reliably determine where you can mutate shared objects safely (without explicit locks) since all other code is suspended until the next yield point. Those guarantees go out the window when two simultaneously running event loops share memory.

> Except if there is a global state in AnyIO's event loop, which I doubt.

Technically there is no such thing as an AnyIO event loop :smile:

davidbrochart:
> Or do you have a lingering, shared event loop thread running in the background?

Yes, the control thread runs for the whole duration of the kernel process, so we're not creating lots of event loops.
The control channel running in its own thread was already there before this PR, and we're taking care of thread safety. We need parallelism as the control channel should work even when executing user blocking code.
At least this PR should not break things with regards to that.

ccordoba12:
Just a quick note to say that this seems significant enough to mark it for IPykernel 7.

davidbrochart:
I'd like to merge this PR and update downstream projects in other PRs.
Any feedback?

davidbrochart:
Let's move forward with this. I'll merge soon if there are no objections.

blink1073:
I think at least the local tests should all be passing before merging.

SylvainCorlay:
> I think at least the local tests should all be passing before merging.

There is a bit of a chicken and egg situation though with respect to the downstream projects that are tested in the CI.

blink1073:
> There is a bit of a chicken and egg situation though with respect to the downstream projects that are tested in the CI.

No I mean explicitly the unit tests in ipykernel

davidbrochart:
Thanks so much for working on this PR Steve!

blink1073:
Okay, unit tests are passing.  I think we should wait for #1210 and #1212 to land and be released, before considering `main` to be `ipykernel 7`.   In the mean time, I'll write some migration docs here and try to address this portion:

"traceback when interrupting blocking user code leaks one frame into the kernel source code".

blink1073:
The fact that async interrupt isn't working on Windows, and is failing on interrupt-related tests on the `jupyter_client` downstream test, leads me to think there is something not quite working about the new interrupt logic (the section in `do_execute` with `execution.interrupt = await to_thread.run_sync(self.shell_interrupt.get)`.  I don't know enough about AnyIO yet to debug properly.

davidbrochart:
Yes, the [AnyIO documentation](https://anyio.readthedocs.io/en/stable/signals.html) says that "Windows does not natively support signals so do not rely on this in a cross platform application". This is about signal handlers in general, not even related to AnyIO, so I don't think we have async cell interrupt support on Windows. But in current ipykernel this is not even possible, on any platform, right? So this is still an improvement.

blink1073:
No, it did work previously, this test was not skipped on Windows: https://github.com/ipython/ipykernel/blob/eddd3e666a82ebec287168b0da7cfa03639a3772/tests/test_async.py#L32.  Additionally, the Jupyter Client test needs to pass before we merge, since it is treating the kernel as a process, and nothing we do here should break that contract.

davidbrochart:
I haven't looked at this test, but with a fresh install of JupyterLab I cannot interrupt a top-level await on Linux:

https://github.com/ipython/ipykernel/assets/4711805/9331be71-2a0d-4f2c-8d4c-5509e03446e1

The `CancelledError` arrives after the cell has finished executing.

davidbrochart:
Indeed, the test only tests that there is a `CancelledError` in the reply, but it doesn't test that it arrives before the cell has finished executing.

blink1073:
Great, we can skip that test on Windows, and concentrate on getting client to pass.

blink1073:
Aha!  The issue is that we need to update [`SignalTestKernel`](https://github.com/jupyter/jupyter_client/blob/4ffbfcf5809ad8913724dd1359089ac3d42031e1/tests/signalkernel.py#L14C7-L14C23) in `jupyter_client`, which is directly subclassing `BaseKernel`.  I'd like to be able to update that class to prove it it still working, but I can't yet figure out why the interrupt is not working the same way.

blink1073:
More progress, I see now that we have to handle `self.kernel.shell_interrupt.get` now in `SignalTestKernel.do_execute`.

ipykernel/kernelbase.py
```diff
@@ -5,21 +5,20 @@
 from __future__ import annotations
 
 import asyncio
-import concurrent.futures
 import inspect
 import itertools
 import logging
 import os
-import socket
+import queue
```
ianthomas23, line 12: There are multiple imports of `queue`, this line and line 20. Just keep the latter?

davidbrochart: > Just keep the latter?

Actually Ruff should have removed line 20 since it's unused?
But I agree we should only keep line 20.

ipykernel/kernelbase.py
```diff
@@ -90,16 +90,9 @@ class Kernel(SingletonConfigurable):
 
     processes: dict[str, psutil.Process] = {}
 
-    @observe("eventloop")
-    def _update_eventloop(self, change):
-        """schedule call to eventloop from IOLoop"""
-        loop = ioloop.IOLoop.current()
-        if change.new is not None:
-            loop.add_callback(self.enter_eventloop)
-
```
jdranczewski, line 99: Hi, this `observe` was the way the kernel entered the eventloop when the `%gui` magic is used, and I don't see a replacement for it - this functionality is thus broken in my testing. `enable_gui` only sets this variable and relies on the observer to start the loop: https://github.com/davidbrochart/ipykernel/blob/92f979eda8280a8898d5235c35df62ac3ee15c79/ipykernel/eventloops.py#L397

It's maybe a little concerning that this didn't cause any of the eventloop tests to fail?

ipykernel/kernelbase.py
```diff
@@ -379,21 +304,88 @@ def should_handle(self, stream, msg, idents):
...
+                # schedule advance again
+                await schedule_next()
+
+        async def schedule_next():
+            """Schedule the next advance of the eventloop"""
+            # flush the eventloop every so often,
+            # giving us a chance to handle messages in the meantime
+            self.log.debug("Scheduling eventloop advance")
+            await sleep(0.001)
+            await advance_eventloop()
```
jdranczewski, line 348: Since I think most of the eventloops rely on returning from the `eventloop()` call to hand control back to the kernel, it may be the case that we enter and exit from the eventloop pretty often, nesting the calls of `schedule_next` and `advance_eventloop` pretty deeply. Would that kind of potentially deep recursion be ok? It seems that the maximum depth allowed by Python is very high, so this is maybe more of an architecture question...

davidbrochart: Yes, we could schedule `advance_eventloop` to be called in a task instead of `await advance_eventloop()`.

ipykernel/kernelbase.py
jdranczewski: Yes, at https://github.com/ipython/ipykernel/blob/main/tests/test_eventloop.py - they seem to mostly not use the magic (entering the loop explicitly), or not check whether the loop is running (just if the related variables are set up)

ipykernel/kernelbase.py
blink1073: Removed the dupe

ipykernel/kernelbase.py
davidbrochart: I'm wondering if we should take advantage of the switch to AnyIO to do event loop integration differently. Trio has a [guest mode](https://trio.readthedocs.io/en/stable/reference-lowlevel.html#what-is-guest-mode) just for that, so that would be the way to go if ipykernel is configured to run with the Trio backend, but then users wouldn't be able to use asyncio. AnyIO doesn't support a guest mode for the asyncio backend, but [it could](https://github.com/agronholm/anyio/issues/251).

jdranczewski: I'm actually not sure of the top of my head how to test for this better - "is the GUI eventloop running" is a slightly nebulous question... Perhaps firing timers within the loop could be an approach - it's what the Tk test does, though it calls the eventloop method directly rather than through the kernel currently

ipykernel/kernelbase.py
```diff
@@ -290,10 +290,50 @@ def __init__(self, **kwargs):
             self.do_execute, ["cell_meta", "cell_id"]
         )
 
-    def dispatch_control(self, msg):
-        self.control_queue.put_nowait(msg)
+    async def dispatch_control(self, msg):
+        # Ensure only one control message is processed at a time
+        async with asyncio.Lock():
```
davidbrochart: But we shouldn't use `asyncio` anymore, right?

blink1073: I messed a few things up in the merge conflict effort, I'll have to come back to this.

ipykernel/kernelbase.py
jdranczewski: Do i understand it right that trio's guest mode needs trio to run as the guest within some external event loop? Currently we 'host' the GUI loop within the kernel's loop, would trio's approach require us to start the GUI loop and then add a smaller trio loop within that as as a guest?

davidbrochart: Yes I think you're right.

dhirschfeld: In case it helps, there's also [`aioguest`](https://github.com/oremanj/aioguest)

minrk:
FWIW, I don't think we can use anyio on anything other than asyncio [until they support polling on FDs](https://github.com/agronholm/anyio/issues/369), and pyzmq will continue to require tornado to run on asyncio on Windows Proactor for the foreseeable future, until/unless someone figures out how to poll on edge-triggered FDs for these event loops that have decided to _remove_ basic functionality on Windows.

agronholm:
> FWIW, I don't think we can use anyio on anything other than asyncio [until they support polling on FDs](https://github.com/agronholm/anyio/issues/369), 

Are you sure you linked to the correct issue? I'm having some trouble understanding the connection. What is lacking FD polling?

minrk:
> Are you sure you linked to the correct issue

ha, not at all, sorry. Fixed link: https://github.com/agronholm/anyio/issues/386

agronholm:
That issue has a simple workaround: calling `socket.fromfd()`. What do you need that can't use that?

minrk:
> That issue has a simple workaround: calling socket.fromfd(). What do you need that can't use that?

Unfortunately, that doesn't help here. The socket in question cannot be read from, it can only be polled for readability. It is an edge-triggered FD, used for signaling events to process on zmq sockets. trio's `wait_readable` works for this using [IOCTL_AFD_POLL](https://github.com/python-trio/trio/blob/48f8293b7d6467a79dff86e7f4b2767fb883d220/src/trio/_core/_io_windows.py#L752). You can also see https://github.com/python-trio/trio/issues/52 for loads of background.

blink1073:
I'm afraid I don't know how to proceed.  It seems as if the changes from #1210 are incompatible with the changes in this PR.  I am getting deadlocked on reading the control loop.

blink1073:
What's really confusing to me is why `nest_asyncio` would show up in the stack trace:

```python
 async with asyncio.Lock():
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/_pytest/runner.py", line 179, in pytest_runtest_teardown
    item.session._setupstate.teardown_exact(nextitem)
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/_pytest/runner.py", line 522, in teardown_exact
    fin()
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/_pytest/fixtures.py", line 677, in <lambda>
    subrequest.node.addfinalizer(lambda: fixturedef.finish(request=subrequest))
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/_pytest/fixtures.py", line 1022, in finish
    func()
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/_pytest/fixtures.py", line 909, in _teardown_yield_fixture
    next(it)
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/anyio/pytest_plugin.py", line 75, in wrapper
    yield from runner.run_asyncgen_fixture(func, kwargs)
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 1940, in run_asyncgen_fixture
    self.get_loop().run_until_complete(
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/nest_asyncio.py", line 84, in run_until_complete
    self._run_once()
  File "/Users/silvester/workspace/.venvs/ipykernel/lib/python3.10/site-packages/nest_asyncio.py", line 107, in _run_once
    event_list = self._selector.select(timeout)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/selectors.py", line 562, in select
    kev_list = self._selector.control(None, max_ev, timeout)

+++++++++++++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++++++++++++
```

blink1073:
If no one else has any ideas or is willing to pick this up, I recommend we close this PR and allow sub-shells work to continue on top of #1210.

blink1073:
Okay, I figured part of it out.  All the tests pass for me locally except:

```
tests/test_kernel.py::test_print_to_correct_cell_from_thread FAILED                          [  7%]
tests/test_kernel.py::test_print_to_correct_cell_from_child_thread FAILED                    [ 11%]
tests/test_kernel.py::test_print_to_correct_cell_from_asyncio FAILED                         [ 14%]
```

And then it times out on:

```
tests/test_kernel.py::test_raw_input
```

davidbrochart:
I can have a look at it next week.

blink1073:
> I can have a look at it next week.

Great, thank you!

davidbrochart:
> Okay, I figured part of it out. All the tests pass for me locally except:
> 
> ```
> tests/test_kernel.py::test_print_to_correct_cell_from_thread FAILED                          [  7%]
> tests/test_kernel.py::test_print_to_correct_cell_from_child_thread FAILED                    [ 11%]
> tests/test_kernel.py::test_print_to_correct_cell_from_asyncio FAILED                         [ 14%]
> ```
> 
> And then it times out on:
> 
> ```
> tests/test_kernel.py::test_raw_input
> ```

I'm looking into it, but at first sight this code was working in my [first commit](https://github.com/ipython/ipykernel/pull/1079/commits/8ff47dc9fb16e7e8aaafc66e3f9908b88bf07891):
```py
from threading import Thread                                                                                                                        
from time import sleep                                                                                                                              

def thread_target():
    for i in range(5):
        print(i, end='', flush=True)
        sleep(0.25)

Thread(target=thread_target).start()
```
It successfully prints `01234` in JupyterLab, while it doesn't work with the HEAD of this PR. Test `test_print_to_correct_cell_from_thread` fails because of that.

davidbrochart:
> Is `test_print_to_correct_cell_from_thread` using the `anyio.from_thread` functions?

No, but yes we might replace all this complicated way of communicating between threads using ZMQ pipes with AnyIO's functions.
Let's see if https://github.com/ipython/ipykernel/pull/1079/commits/c1dce807d3172942ab2bb11a48c5b07ccf7542ba improves things first.

davidbrochart:
> Let's see if [c1dce80](https://github.com/ipython/ipykernel/commit/c1dce807d3172942ab2bb11a48c5b07ccf7542ba) improves things first.

Doesn't look good :smile:

davidbrochart:
@blink1073 For me even without https://github.com/ipython/ipykernel/pull/1079/commits/c1dce807d3172942ab2bb11a48c5b07ccf7542ba a lot of tests hang locally, for instance tests related to cell interrupt and in-process kernels. It seemed to be the case [in the CI](https://github.com/ipython/ipykernel/actions/runs/8223298522/job/22485833531) too.

davidbrochart:
https://github.com/ipython/ipykernel/pull/1079/commits/56fdee2744a5c2a9c9a11f46c201b7811abab2fb fixes `tests/test_kernel.py::test_raw_input` and probably other tests (it was in my first commit https://github.com/ipython/ipykernel/pull/1079/commits/8ff47dc9fb16e7e8aaafc66e3f9908b88bf07891 but removed later).

davidbrochart:
So currently the remaining failing tests are in `tests/inprocess`.

blink1073:
> So currently the remaining failing tests are in tests/inprocess.

I'd say let's skip those for now, and then merge and iterate.

davidbrochart:
One remaining failing test on Windows: `tests/test_async.py::test_async_interrupt[asyncio]`.
But should we ignore it on Windows, since it's unlikely we can have it working and in current ipykernel we cannot interrupt an async cell on any platform? See [this comment](https://github.com/ipython/ipykernel/pull/1079#issuecomment-1958966597).

blink1073:
Sounds right to me

blink1073:
Thanks @davidbrochart!

davidbrochart:
Thank you @blink1073 !

rayosborn:
Is a fix for the in-process kernel test failure planned. before the release of v7?

PR #1084, opened 2023-01-31
Remove test_enter_eventloop

davidbrochart:
[test_enter_eventloop](https://github.com/ipython/ipykernel/blob/fe8b7b0f3ca0e7a3cb4b7cb1660e949c121905b7/ipykernel/tests/test_kernel_direct.py#L138) does the following:
```py
kernel.eventloop = asyncio.get_running_loop()
```
The only place where `kernelbase.eventloop` is used is [here](https://github.com/ipython/ipykernel/blob/4f8438855187c90bc233567fba2d703d8719164c/ipykernel/kernelbase.py#L459), and it seems to call the event loop. But an asyncio event loop is not callable, or am I missing something?
I inserted an error just before the call to the event loop, and the test still passes, so what does this test do?
Should we remove it?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,7 +131,7 @@ jobs:
 
       - name: Run the tests
         timeout-minutes: 10
-        run: pytest -W default -vv || pytest --vv -W default --lf
+        run: pytest -k test_enter_eventloop -W default -vv || pytest -k test_enter_eventloop --vv -W default --lf
 
   test_miniumum_versions:
     name: Test Minimum Versions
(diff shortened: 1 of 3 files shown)


davidbrochart:
This last commit removes [test_enter_eventloop](https://github.com/ipython/ipykernel/blob/fe8b7b0f3ca0e7a3cb4b7cb1660e949c121905b7/ipykernel/tests/test_kernel_direct.py#L135-L153).
If someone is against it, please speak up. Otherwise, I will merge.

blink1073:
That test was there to add more coverage in `enter_eventloop`, I'm fine to remove it.

PR #1091, opened 2023-02-08
Return Comm rather than BaseComm from create_comm

### LINKED ISSUE #1090: ipywidgets expects Comm, not BaseComm
Hi!  I've been trying to track down an issue with my usage of `ipywidgets`, described [on discourse](https://discourse.jupyter.org/t/trouble-creating-a-widget-in-kernel-from-frontend/17887).  The basic gist is:

 - I'm attempting to create a widget from the client by using a `model.widget_manager`'s method `new_widget`.  In previous versions, this worked, but now, it doesn't.
 - The issue seems to stem from the kernel attempting to create a new comm instance for a widget, but supplying to that widget a `BaseComm` rather than a `Comm`, which throws a validation error.

After digging through this, it looks to me like `ipywidgets` declares `comm = Instance('ipykernel.comm.Comm', allow_none=True)`, but `ipykernel.ipkernel.create_comm` explicitly creates a `BaseComm`, so the traitlet doesn't validate, and the comm fails to properly initialize.  This results in the widgets not being synced back to the kernel from the frontend.

I confess that while I've spent time going through the code, I'm not entirely sure if this could potentially be a version mismatch, or if this is an oversight for a use case that may not come up terribly frequently.

My versions:

 - `ipywidgets` - 8.0.4
 - `ipykernel` - 6.21.1
 - `comm` - 0.1.2

I was able to address the problem I was having by changing the import and usage of `BaseComm` to `Comm` in `ipkernel.py`, and so if this is an acceptable solution I'd be happy to issue a pull request.  (My hesitance is that there may be unexpected downstream effects of a change to something so deep in the kernel code, and so I wanted to open the issue as possible first step.)

matthewturk:
This replaces `BaseComm` in `create_comm` with `Comm`, which should match what is expected by `ipywidgets` for validation of the `comm` trait in subclasses of `ipywidgets.Widget`.

Fixes #1090 .

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -29,7 +29,7 @@ from .zmqshell import ZMQInteractiveShell
 try:
     from IPython.core.interactiveshell import _asyncio_runner  # type:ignore[attr-defined]
 except ImportError:
-    _asyncio_runner = None
+    _asyncio_runner = None  # type:ignore
 
 try:
     from IPython.core.completer import provisionalcompleter as _provisionalcompleter
@@ -626,7 +626,7 @@ class IPythonKernel(KernelBase):
         try:
             from ipyparallel.serialize import serialize_object, unpack_apply_message
         except ImportError:
-            from .serialize import serialize_object, unpack_apply_message  # type:ignore
+            from .serialize import serialize_object, unpack_apply_message
 
         shell = self.shell
         try:
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -17,7 +17,7 @@ from IPython.utils.tokenutil import line_at_cursor, token_at_cursor
 from traitlets import Any, Bool, HasTraits, Instance, List, Type, observe, observe_compat
 from zmq.eventloop.zmqstream import ZMQStream
 
-from .comm.comm import BaseComm
+from .comm.comm import Comm
 from .comm.manager import CommManager
 from .compiler import XCachingCompiler
 from .debugger import Debugger, _is_debugpy_available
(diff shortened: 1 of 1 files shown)


matthewturk:
It looks like the failure is in `ipyparallel` with:

```
______________________ AsyncResultTest.test_wait_for_send ______________________
  
  self = <ipyparallel.tests.test_asyncresult.AsyncResultTest testMethod=test_wait_for_send>
  
      def test_wait_for_send(self):
          view = self.client[-1]
          view.track = True
  >       with self.assertRaises(TimeoutError):
  E       AssertionError: TimeoutError not raised
  
  ipyparallel/tests/test_asyncresult.py:377: AssertionError
```

My intuition doesn't have much to say about if this is related.

blink1073:
It seems to be an intermittent failure.  I saw it in #1089, but then it passed on the merge commit.  I just kicked it.

ipykernel/ipkernel.py
```diff
@@ -45,7 +45,7 @@
 
 def _create_comm(*args, **kwargs):
     """Create a new Comm."""
-    return BaseComm(*args, **kwargs)
```
blink1073, line 48: Ah, I misunderstood what you intended to change.  We wanted to explicitly return `BaseComm` here because it is more efficient.   I think the right answer is for `ipywidgets` to relax its expected trait.

blink1073:
cc @maartenbreddels @martinRenou

ipykernel/ipkernel.py
maartenbreddels: yeah, BaseComm is 'the new thing', Comm is for backwards compatibility with ipywidgets.

matthewturk: I see, but since `ipywidgets` doesn't accept `BaseComm` should that be the submitted change?  I guess I'm confused because @blink1073's suggestion seemed straightforward, but now I don't understand how to interpret @maartenbreddels's reply.

Carreau:

PR #1095, opened 2023-02-13
Un-expose __file__ and expose __session__ instead.

Carreau:
See https://github.com/jupyter-server/jupyter_server/issues/1198,

setting __file__ will break multiprocessing.

I'm not setting `__notebook__` as things that set this value may note be notebooks.


blink1073 (approved):
Thank you!

Issue #1100, opened 2023-02-24
InterruptReply missing status and error properties


garlandz-db:
According to the jupyter [messaging docs](https://jupyter-client.readthedocs.io/en/stable/messaging.html#kernel-interrupt), we should be passing a `status` field in the content of the `InterruptReply` message. However in the[ implementation](https://github.com/ipython/ipykernel/blob/0f09aa690d1881ab6a2b92119215d8561d344e67/ipykernel/kernelbase.py#L908), we are just forwarding the request content (empty) back into the reply. As such, the reply has no information on `status` and the error properties in case the interrupt is not successful.

My suggestion on the fix should be a try catch around the `send_interrupt_children` and return a `{ 'status' : 'ok' }` if no catch occurs and otherwise `{ 'status': 'error', ... }`.

blink1073:
Good catch!  Would you like to submit a PR?

jasongrout-db:
It would be good to add a check in the tests at 

https://github.com/ipython/ipykernel/blob/ea33a31ca48fa13d73b44837b6791aded1fe4858/ipykernel/tests/test_kernel_direct.py#L71-L73 

https://github.com/ipython/ipykernel/blob/ea33a31ca48fa13d73b44837b6791aded1fe4858/ipykernel/tests/test_ipkernel_direct.py#L92-L94

 (and to change those tests to send the interrupt request on the control channel instead of the shell channel).

garlandz-db:
i can get a PR in next week @blink1073.

blink1073:
Great, thanks!

garlandz-db:
any idea where i can get the traceback?

blink1073:
I would set a debug point in the tests and inspect the `reply`.

PR #1101, opened 2023-02-28
Fix interrupt reply

garlandz-db:
issue: https://github.com/ipython/ipykernel/issues/1100#issuecomment-1448133194 

Changes in PR
* Added reply content properties for interrupt_request
* `send_interupt_children` => `send_interrupt_children` renaming
* Added test coverage to verify statuses are received

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -898,13 +898,13 @@ class Kernel(SingletonConfigurable):
 
     async def interrupt_request(self, stream, ident, parent):
         """Handle an interrupt request."""
+        content: t.Dict[str, t.Any] = {"status": "ok"}
         try:
             self._send_interrupt_children()
-            content = {"status": "ok"}
         except OSError as err:
             import traceback
 
-            content: t.Dict[str, t.Any] = {
+            content = {
                 "status": "error",
                 "traceback": traceback.format_stack(),
                 "ename": str(type(err).__name__),
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -883,7 +883,7 @@ class Kernel(SingletonConfigurable):
         msg = self.session.send(stream, "comm_info_reply", reply_content, parent, ident)
         self.log.debug("%s", msg)
 
-    def _send_interupt_children(self):
+    def _send_interrupt_children(self):
         if os.name == "nt":
             self.log.error("Interrupt message not supported on Windows")
         else:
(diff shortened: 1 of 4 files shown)


ipykernel/kernelbase.py
```diff
@@ -892,20 +892,25 @@ def _send_interupt_children(self):
...
-                    os.killpg(pgid, SIGINT)
-                    return
-                except OSError:
-                    pass
-            try:
+                os.killpg(pgid, SIGINT)
+            else:
                 os.kill(pid, SIGINT)
-            except OSError:
-                pass
```
garlandz-db, line 903: move out OSError try except

ipykernel/tests/test_ipkernel_direct.py
```diff
@@ -90,8 +90,21 @@ async def test_comm_info_request(ipkernel):
 
 
 async def test_direct_interrupt_request(ipkernel):
-    reply = await ipkernel.test_shell_message("interrupt_request", {})
+    reply = await ipkernel.test_control_message("interrupt_request", {})
```
garlandz-db, line 93: shell => control

garlandz-db:
@blink1073 how do i add the label? i dont see an option

ipykernel/kernelbase.py
blink1073: This is a change in behavior.  I think we want to do something like:

```python
try:
    os.killpg(pgid, SIGINT)
except OSError:
    os.kill(pid, SIGINT)
    raise
os.kill(pid, SIGINT)
````

garlandz-db: oops youre right

blink1073:
@ccordoba12 this fix breaks one of the `spyder-kernels` tests, but I believe it is the correct behavior.

ccordoba12:
@impact27, could you take a look at this one, please?

impact27:
In spyder kernel we want to be able to raise an interrupt from a thread on the main thread. This is done by 
`self.kernel._send_interupt_children()`

I guess we just need to rename it `self.kernel._send_interrupt_children()` but that means we will have to change the minimum ipykernel version so this works

blink1073:
We could make this backward compatible by having `_send_interupt_children` warn and call `_send_interrupt_children`.

ccordoba12:
Well, the call to `_send_interupt_children()` is only present in our master branch, which hasn't been released yet. So, I think it's simpler if we skip the failing test so that this PR passes. Then we switch to the new method name and increase our minimal IPykernel supported, as @impact27 suggested.

ccordoba12:
@blink1073, our test suite should pass now thanks to PR https://github.com/spyder-ide/spyder-kernels/pull/446. I went for a simpler solution: detect if the kernel has `_send_interupt_children` and if not use `_send_interrupt_children`.

blink1073:
Confirmed, thanks all!  I'll make a release early next week.

PR #1106, opened 2023-03-25
InProcessKernelClient: Do not use run_sync if already in a running loop

frmdstryr:
I have an application that uses an integrated Qt and asyncio event loop that embeds an ipython console using [enaml's ipython console](https://github.com/nucleic/enaml/blob/main/enaml/qt/qt_ipython_console.py). 

When the `run_sync` call finishes here it attempts to close the main loop causing the whole app to hang. 

This just adds a check so if it's already in a running loop it just uses that instead.

### DIFF
--- a/ipykernel/inprocess/client.py
+++ b/ipykernel/inprocess/client.py
@@ -27,6 +27,12 @@ from .channels import InProcessChannel, InProcessHBChannel
 # Main kernel Client class
 # -----------------------------------------------------------------------------
 
+def in_running_loop() -> bool:
+    try:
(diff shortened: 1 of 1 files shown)


frmdstryr:
After reading through the run_sync code it looks like it might be something with async qt integration not working in a nested thread.

blink1073:
Hi @frmdstryr, this might be fixed by #1137, I'll be releasing a bugfix early next week.

frmdstryr:
Indeed. Thanks!

PR #1110, opened 2023-04-14
Add outstream hook similar to display publisher

maartenbreddels:
Similar to #115 but for stdout/stderr

This can be useful for a thread to redirect stdout/stderr to a file, while other threads can still write to stdout/stderr.


blink1073 (approved):
LGTM, thanks!

PR #1114, opened 2023-05-04
Support control<>iopub messages to e.g. unblock comm_msg from command execution 

tkrabel-db:
**Disclaimer:** I work for Databricks, where we patched the control channel to handle jupyter messages like `comm_open` and `comm_msg`.

# What changes with this PR?

At Databricks (which uses ipykernel), we often face the situation that customers execute long running pyspark jobs which block the shell channel for minutes or even an hour. In order to provide a better user experience, we unblock some original shell messages by moving them over to the control channel. We successfully did that with `complete_request`, `comm_open` and `comm_msg`.

Part of our fix was changing the comm object so that it gets the parent header from the correct thread. My PR is about that change.

### DIFF
--- a/ipykernel/comm/comm.py
+++ b/ipykernel/comm/comm.py
@@ -3,6 +3,7 @@
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import threading
 import uuid
 from typing import Optional
 from warnings import warn
@@ -13,6 +14,7 @@ from traitlets import Bool, Bytes, Instance, Unicode, default
 
 from ipykernel.jsonutil import json_clean
 from ipykernel.kernelbase import Kernel
+from ipykernel.control import CONTROL_THREAD_NAME
 
 
 # this is the class that will be created if we do comm.create_comm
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thank you!

tkrabel-db:
@jasongrout FYI. This went faster than expected :)

maartenbreddels:
Just for the record: this does not change existing behavior? This only supports your use case of sending back to the right channel if that happens. So this is totally backward compatible, right?

tkrabel-db:
Yes. The existing behaviour is shell<>iopub channel communication, which isn't affected. This change supports adding control<>iopub channel communication without the necessity of patching internal ipykernel functions, which is what we had to do at Databricks so far.

jasongrout:
Awesome, thanks for reviewing this, @blink1073. 

I talked with @minrk about this yesterday night too, and he suggested that this if statement would be better positioned in the `get_parent` function, so that it could be used by not only comms, but by anyone getting the appropriate parent header in a thread. More concretely, he suggested the logic for picking the right parent header be implemented at https://github.com/ipython/ipykernel/blob/d63c33afb9872f2781997b2428d7e9e0c1d23d41/ipykernel/kernelbase.py#L632 and we change the signature of `get_parent` to something like `get_parent(self, channel=None)` (where the default None value means to use this logic to determine the correct parent header).

@tkrabel-db - what do you think?

PR #1116, opened 2023-05-05
close event loop when we finish

minrk:
doesn't usually make a difference, but a good practice nonetheless.


minrk:
actually breaks the reentrant case, need to figure out what we do there.

Carreau:

PR #1129, opened 2023-06-26
Bugfix: binary stdout/stderr handling

arieleiz:
`_watch_pipe_fd()` decodes the incoming data from the pipe using str::decode(). By default this passed the `strict` errors flag to the unicode engine. However, if the pipe command returns binary data this will through a `UnicodeError` exception.
We change the code to explicitly pass the `replace` errors flag, which will replace the output with output that can be displayed on screen.
We also specify the buffer size as a constant so it can be changed for edge cases where the task pipes a lot of output/error text.


blink1073 (approved):
Thank you!

PR #1127, opened 2023-06-21
Check existence of connection_file before writing

fecet:
I find the connection file `kernel-<uuid>.json` would be created twice when starting and connecting kernel. I thought this is not as expected since the first time kernel created:

https://github.com/jupyter/jupyter_client/blob/57b11c36b1eceae138cfc6044a2274e2e1caa135/jupyter_client/connect.py#L488

do check this. But the second call, i.e., come from the kernelapp doesn't. Thus that json file would overwrite the previous one. I don't know the general order though, this order is I observed when open a new jupyter notebook.

And this time `kernelname` was not passed. So I never find `kernelname` in connection file, I believe this was a bug.

PR #1133, opened 2023-07-13
Merge connection info into existing connection file if it already exists

jasongrout:
This PR lets a connection file specify a port as 0 and then be updated when ipykernel picks a port, as a follow-up to https://github.com/ipython/ipykernel/pull/1127.

In https://github.com/ipython/ipykernel/pull/1127, ipykernel was changed to not overwrite a kernel connection file when it already exists. This broke a pattern we used at Databricks (thanks to @mrbago) that addresses the same handshaking issues as https://github.com/jupyter/jupyter_client/issues/487 and https://github.com/jupyter/enhancement-proposals/pull/66 (i.e., [JEP 66](https://github.com/jupyter/enhancement-proposals/blob/master/66-jupyter-handshaking/jupyter-handshaking.md)). Our pattern was this:

1. The kernel client writes a connection file `connect.json` like the following, where all the port numbers are zero:
   ```json
    {
      "shell_port": 0,
      "iopub_port": 0,
      "stdin_port": 0,
      "control_port": 0,
      "hb_port": 0,
      "ip": "*",
      "key": "abc123",
      "transport": "tcp",
      "signature_scheme": "hmac-sha256",
      "kernel_name": "My Kernel"
    }
   ```
2. ipykernel is started, pointing at that connection file.
   ```
   python -m ipykernel -f ./connect.json
   ```
4. ipykernel/zmq picks random ports for each of the channels
5. ipykernel (before https://github.com/ipython/ipykernel/pull/1127) overwrites the connection file with the correct port numbers and `ip` interface address it actually connected to
6. We read the connection file to see what ports the client needs to use to communicate with ipykernel.
    ```
    $ cat connect.json 
    {
      "shell_port": 62757,
      "iopub_port": 62761,
      "stdin_port": 62758,
      "control_port": 62759,
      "hb_port": 62763,
      "ip": "0.0.0.0",
      "key": "abc123",
      "transport": "tcp",
      "signature_scheme": "hmac-sha256",
      "kernel_name": ""
    }
    ```
In https://github.com/ipython/ipykernel/pull/1127, it was noticed that this overwrote the kernel_name as well. In essence, ipykernel was taking over the file that someone else had created. The solution in https://github.com/ipython/ipykernel/pull/1127 was to not touch the connection file if it exists, but that breaks the usecase of letting the kernel pick ports and interfaces.

This PR instead merges the new information from ipykernel into the connection file. If the file would not change from the updated information, then we don't write to the connection file. In short, with this PR, we get the following in the last step above:
```
$ cat connect.json 
{
  "shell_port": 62914,
  "iopub_port": 62918,
  "stdin_port": 62915,
  "control_port": 62916,
  "hb_port": 62920,
  "ip": "0.0.0.0",
  "key": "abc123",
  "transport": "tcp",
  "signature_scheme": "hmac-sha256",
  "kernel_name": "My Kernel"
}
```

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -24,7 +24,6 @@ from IPython.core.application import (  # type:ignore[attr-defined]
 )
 from IPython.core.profiledir import ProfileDir
 from IPython.core.shellapp import InteractiveShellApp, shell_aliases, shell_flags
-from jupyter_client import write_connection_file
 from jupyter_client.connect import ConnectionFileMixin
 from jupyter_client.session import Session, session_aliases, session_flags
 from jupyter_core.paths import jupyter_runtime_dir
@@ -44,10 +43,11 @@ from traitlets.utils import filefind
 from traitlets.utils.importstring import import_item
 from zmq.eventloop.zmqstream import ZMQStream
 
-from .control import ControlThread
-from .heartbeat import Heartbeat
+from .connect import get_connection_info, write_connection_file
 
 # local imports
+from .control import ControlThread
+from .heartbeat import Heartbeat
 from .iostream import IOPubThread
 from .ipkernel import IPythonKernel
 from .parentpoller import ParentPollerUnix, ParentPollerWindows
@@ -260,12 +260,7 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp, ConnectionFileMix
     def write_connection_file(self):
         """write connection info to JSON file"""
         cf = self.abs_connection_file
-        if os.path.exists(cf):
-            self.log.debug("Connection file %s already exists", cf)
-            return
-        self.log.debug("Writing connection file: %s", cf)
-        write_connection_file(
-            cf,
+        connection_info = dict(
             ip=self.ip,
             key=self.session.key,
             transport=self.transport,
@@ -275,6 +270,19 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp, ConnectionFileMix
             iopub_port=self.iopub_port,
             control_port=self.control_port,
         )
+        if os.path.exists(cf):
+            # If the file exists, merge our info into it. For example, if the
+            # original file had port number 0, we update with the actual port
+            # used.
+            existing_connection_info = get_connection_info(cf, unpack=True)
+            connection_info = dict(existing_connection_info, **connection_info)
+            if connection_info == existing_connection_info:
+                self.log.debug("Connection file %s with current information already exists", cf)
+                return
+
+        self.log.debug("Writing connection file: %s", cf)
+
+        write_connection_file(cf, **connection_info)
 
     def cleanup_connection_file(self):
         """Clean up our connection file."""
--- a/ipykernel/tests/test_ipkernel_direct.py
+++ b/ipykernel/tests/test_ipkernel_direct.py
@@ -20,7 +20,7 @@ class user_mod:
     __dict__ = {}
 
 
-async def test_properities(ipkernel: IPythonKernel) -> None:
+async def test_properties(ipkernel: IPythonKernel) -> None:
     ipkernel.user_module = user_mod()
     ipkernel.user_ns = {}
 
--- a/ipykernel/tests/test_kernelapp.py
+++ b/ipykernel/tests/test_kernelapp.py
@@ -50,7 +50,7 @@ def test_start_app():
     app.kernel.destroy()
     app.close()
 
-
+@pytest.mark.skipif(os.name == "nt", reason="permission errors on windows")
 def test_merge_connection_file():
     cfg = Config()
     with TemporaryWorkingDirectory() as d:
--- a/ipykernel/tests/test_kernelapp.py
+++ b/ipykernel/tests/test_kernelapp.py
@@ -50,6 +50,7 @@ def test_start_app():
     app.kernel.destroy()
     app.close()
 
+
 @pytest.mark.skipif(os.name == "nt", reason="permission errors on windows")
 def test_merge_connection_file():
     cfg = Config()
--- a/ipykernel/tests/test_kernelapp.py
+++ b/ipykernel/tests/test_kernelapp.py
@@ -1,13 +1,17 @@
+import json
 import os
 import threading
 import time
 from unittest.mock import patch
 
 import pytest
+from jupyter_core.paths import secure_write
+from traitlets.config.loader import Config
 
 from ipykernel.kernelapp import IPKernelApp
 
 from .conftest import MockKernel
+from .utils import TemporaryWorkingDirectory
 
 try:
     import trio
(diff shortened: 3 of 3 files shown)


jasongrout-db:
In https://github.com/ipython/ipykernel/pull/1127, ipykernel was changed to not overwrite a kernel connection file when it already exists. This broke a pattern we used at Databricks to address the issues from https://github.com/jupyter/jupyter_client/issues/487 and https://github.com/jupyter/enhancement-proposals/pull/66 (i.e., [JEP 66](https://github.com/jupyter/enhancement-proposals/blob/master/66-jupyter-handshaking/jupyter-handshaking.md)). Our pattern was this:

1. The kernel client writes a connection file like the following, where all the port numbers are zero:
   ```json
    {
      "shell_port": 0,
      "iopub_port": 0,
      "stdin_port": 0,
      "control_port": 0,
      "hb_port": 0,
      "ip": "127.0.0.1",
      "key": "53a0b8b4-e6272129eb8f4602affee828",
      "transport": "tcp",
      "signature_scheme": "hmac-sha256",
      "kernel_name": ""
    }
   ```
2. ipykernel is started, pointing at that connection file.
3. ipykernel/zmq picks random ports for each of the channels
4. ipykernel (before https://github.com/ipython/ipykernel/pull/1127) overwrites the connection file with the correct port numbers it actually connected to
5. We read the connection file to see what ports the client needs to use to communicate with ipykernel.

Since https://github.com/ipython/ipykernel/pull/1127, the connection file is not updated with the actual ports ipykernel is using.

jasongrout:
@blink1073 - pinging you for review, since you also reviewed #1127. I'm not sure if this usecase of setting the port number to 0 and having the kernel rewrite the connection file with the actual port was considered in the original discussion on #1127.

We discussed this briefly in the server/kernels meeting today, and we had a generally positive reaction from people there today, and no objections to the idea of updating the connection file.

jasongrout:
Ping also @fecet, the original author of #1127. This PR implements your idea at https://github.com/ipython/ipykernel/pull/1127#issuecomment-1600761391:

> ... Or can we just read the kernel file if exists, and only update the difference? ....

jasongrout:
The tests_check CI test is failing because the prerelease test timed out. The prerelease test timed out 99% of the way through, so perhaps it is just at the borderline of the timeout (i.e., it looks to me like this failure is not particular to this PR).

JohanMabille:
Now that [JEP 66](https://github.com/jupyter/enhancement-proposals/blob/master/66-jupyter-handshaking/jupyter-handshaking.md) has been approved, will you still need the ability to overwrite the connection file when ipykernel implements the handshake pattern? This latter seems more efficient to me since it does not require polling the file system.

SylvainCorlay (changes requested):
This PR changes ipykernel's behavior with respect to the mechanism for port selection. 
JEP 66 just went through, addressing this very issue. Could we work together on implementing the JEP rather than adding a special workaround in ipykernel?

fecet:
> Ping also @fecet, the original author of #1127. This PR implements your idea at https://github.com/ipython/ipykernel/pull/1127#issuecomment-1600761391:
> 
> > ... Or can we just read the kernel file if exists, and only update the difference? ....

great to hear that！

jasongrout:
> Now that [JEP 66](https://github.com/jupyter/enhancement-proposals/blob/master/66-jupyter-handshaking/jupyter-handshaking.md) has been approved, will you still need the ability to overwrite the connection file when ipykernel implements the handshake pattern? This latter seems more efficient to me since it does not require to poll the file system.

I don't know - but that is a hypothetical at this point since it isn't implemented, and the change in 6.23.3 (accidentally) broke long-standing behavior and is breaking actual systems in use right now that rely on that behavior.

> This PR changes ipykernel's behavior with respect to the mechanism for port selection.
> JEP 66 just went through, addressing this very issue. Could we work together on implementing the JEP rather than adding a special workaround in ipykernel?

The way I see it, there was an accidental regression introduced in #1127, released in a patch release, that changed long-standing behavior that people rely on right now in ipykernel. This is breaking systems that rely on ipykernel right now. We should have brought this regression to @fecet's attention in #1127, but regardless, here we are.

This PR does not change ipykernel's behavior, rather it restores long-standing behavior to what it was before the regression. Of course, another route would be to revert #1127, but I think this PR preserves the intent of #1127 without this regression, and happily aligns with one of @fecet's original suggestions for #1127.

I think once we've resolved this regression, we can have a discussion about whether this port 0 behavior should be superseded by JEP 66's ideas, but of course that succession should happen deliberately, with a period for migration, only after JEP 66 has been implemented.

SylvainCorlay:
Let's discuss how we can resolve the issue caused by the change of behavior in https://github.com/ipython/ipykernel/pull/1127.

> I think once we've resolved this regression, we can have a discussion about whether this port 0 behavior should be superseded by JEP 66's ideas, but of course that succession should happen deliberately, with a period for migration, only after JEP 66 has been implemented.

The idea of not passing ports and polling the file system to wait for the info to be populated by the kernel was addressed in the discussion about the JEP.

jasongrout:
> Let's discuss how we can resolve the issue caused by the change of behavior in #1127.

Thank you. I agree - I think *this* PR discussion should be about fixing the regression in behavior introduced in #1127. Here are, for example, two paths forward:

1. Revert #1127 (if this is going to be a long discussion blocking merging of this PR, I'd advocate for reverting to unbreak things in the short term)
2. Merge this PR, which achieves the aims of #1127 while preserving long-standing ipykernel behavior

I prefer merging this PR, as it builds on #1127 (which did solve an important issue). Which do you prefer, or do you see another path forward that fixes the behavior change in #1127 in the short term?

> The idea of not passing ports and polling the file system to wait for the info to be populated by the kernel was addressed in the discussion about the JEP.

Great! I think there are other places to discuss JEP 66 - let's not rehash a JEP 66 discussion here.

blink1073:
I'm afk until next Monday, feel free to do whatever you all think is best with this PR.

ccordoba12:
> The way I see it, there was an accidental regression introduced in https://github.com/ipython/ipykernel/pull/1127, released in a patch release, that changed long-standing behavior that people rely on right now in ipykernel. This is breaking systems that rely on ipykernel right now. We should have brought this regression to @fecet's attention in https://github.com/ipython/ipykernel/pull/1127, but hey, systems are complicated, and it's not clear anyone recognized this regression at the time, so here we are.
>
> This PR does not change ipykernel's behavior, rather it restores long-standing behavior to what it was before the regression

Given this analysis of the situation that Jason provided, I agree that we should either merge this PR or revert #1127 (preferable?) and release a new bugfix version to restore the previous behavior.

Just my 2 cents.

fecet:
If https://github.com/ipython/ipykernel/pull/1127 be revoked I hope this https://github.com/jupyter/notebook/issues/6936 can also be considered.

blink1073:
I'm going to merge this to fix the current bug, we can address JEP 66 as a follow-up.

SylvainCorlay:
I have a reservation with respect to the special casing of the port number "0" to signal that ports have to be chosen. (this behavior is not really part of the protocol. If this fixes your scenario, I guess this is fine as a temporary fix.

PR #1137, opened 2023-07-31
Avoid starting IOPub background thread after it's been stopped

minrk:
`loop.add_callback(loop.stop)` called before the thread starts can get consumed by the run_sync before loop.start, so it won't close the loop. This is causing some tests to hang (e.g. ipython downstream tests), but isn't likely to affect any users at runtime.

The run_sync was added in #1125

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -57,6 +57,7 @@ class IOPubThread:
             piped from subprocesses.
         """
         self.socket = socket
+        self._stopped = False
         self.background_socket = BackgroundSocket(self)
         self._master_pid = os.getpid()
         self._pipe_flag = pipe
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thank you!

blink1073:
Hmm it looks like this failure is related, it happened twice in a row:

```python
  FAILED ipykernel/inprocess/tests/test_kernelmanager.py::InProcessKernelManagerTestCase::test_interface - pytest.PytestUnhandledThreadExceptionWarning: Exception in thread IOPub
  
    File "/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
      self.run()
    File "/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/threading.py", line 975, in run
      self._target(*self._args, **self._kwargs)
    File "/home/runner/work/ipykernel/ipykernel/sdist/test/ipykernel/iostream.py", line 98, in _thread_main
      self.io_loop.run_sync(_cancel)
    File "/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/tornado/ioloop.py", line 526, in run_sync
      raise TimeoutError("Operation timed out after %s seconds" % timeout)
  TimeoutError: Operation timed out after None seconds
```

PR #1142, opened 2023-08-30
Don't call QApplication.setQuitOnLastWindowClosed(False).

anntzer:
That setting previously caused
`plt.plot(); print("pre"); plt.show(block=True); print("post")` to block after the `show()` call *even* after closing the Matplotlib figure window ("post" was never printed).

The new behavior (printing "post" after the window is closed) is consistent with plain IPython.

The setQuitOnLastWindowClosed call came in (with no additional explanation) in the very first commit implementing qt support (bec2d41, in 2010) and seems to have never been touched since then.

This should close #4367 (which is really an ipykernel issue per https://github.com/spyder-ide/spyder/issues/4367#issuecomment-295026348), which had previously been closed as "can't fix, workaround available".  Attn @ccordoba12 who commented on the spyder issue.

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -535,13 +535,10 @@ def make_qt_app_for_kernel(gui, kernel):
     set_qt_api_env_from_gui(gui)
 
     # This import is guaranteed to work now:
-    from IPython.external.qt_for_kernel import QtCore, QtGui
+    from IPython.external.qt_for_kernel import QtCore
     from IPython.lib.guisupport import get_app_qt4
 
     kernel.app = get_app_qt4([" "])
-    if isinstance(kernel.app, QtGui.QApplication):
-        kernel.app.setQuitOnLastWindowClosed(False)
-
     kernel.app.qt_event_loop = QtCore.QEventLoop(kernel.app)
 
 


ccordoba12:
Thanks @anntzer for taking a look at this problem! I tested your PR locally and it worked as expected for me on Linux.

Did you test it on Windows? If not, I can ask one of the Spyder devs to do it just to be extra sure.

> The setQuitOnLastWindowClosed call came in (with no additional explanation) in the very first commit implementing qt support (https://github.com/ipython/ipykernel/commit/bec2d41a5fa5c0880149f616cbbf19b63491133c, in 2010) and seems to have never been touched since then.

It was probably needed in the old Qt4 days, but it seems to not be anymore.

anntzer:
> Did you test it on Windows?

Not directly (it's a bit tricky for me to access a Windows dev setup), but I did check that on Windows, in the Spyder console, I can run QApplication.instance().setQuitOnLastWindowClosed(False) (effectively undoing the setting from ipykernel) and then plt.show(block=False) does work as desired.

ccordoba12:
Ok, then this looks good to me.

ccordoba12 (approved):
Thanks @anntzer!

blink1073 (approved):
Thank you!

PR #1167, opened 2023-11-19
Update ruff and typings

blink1073:
Update for https://github.com/scientific-python/cookie/releases/tag/2023.11.17 and mypy 1.7.


blink1073:
Failure reported upstream: https://github.com/pytest-dev/pytest-asyncio/issues/679

PR #1169, opened 2023-11-20
Extend argument handling of do_execute with cell metadata

jjvraw:
Enhances the `do_execute` method by allowing an optional `cell_meta` parameter, which represents cell metadata. This addition aligns with the Jupyter messaging protocol on [metadata](https://jupyter-client.readthedocs.io/en/latest/messaging.html#metadata), which acknowledges the potential use of metadata in various message types.

Although `metadata` may not be often used, the capability to handle metadata within `do_execute` can expand the kernel's flexibility, especially in scenarios involving complex interactions or custom extensions, as detailed in the [jupyter_client docs](https://jupyter-client.readthedocs.io/en/latest/messaging.html#metadata).

A noteable use case is in the development of [Python wrapper kernels](https://jupyter-client.readthedocs.io/en/latest/wrapperkernels.html#making-simple-python-wrapper-kernels), where additional information might be passed via metadata to influence execution behavior, provide additional context, or interact with external services.

Changes:
 - Extending the `do_execute` signature to include an optional metadata parameter
 - Implementing logic in `execute_request` to extract and pass down cell metadata to `do_execute` 
   - This involved expanding `_accepts_cell_id` to `_accepts_parameters` for broader parameter checks, which in turn seemly created nice flow in `execute_request`.

---

It seemed bold requesting to replace `cell_id` with `metadata` despite one is derived from the other. However, `_accepts_parameters` enables flexible handling of these optional parameters, thus somewhat ensuring backward compatibility. A minor concern might involve the imports of `_accepts_cell_id`, which has been removed/replaced, but can be reintegrated if necessary.

Thanks

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -23,7 +23,7 @@ from .compiler import XCachingCompiler
 from .debugger import Debugger, _is_debugpy_available
 from .eventloops import _use_appnope
 from .kernelbase import Kernel as KernelBase
-from .kernelbase import _accepts_cell_id
+from .kernelbase import _accepts_parameters
 from .zmqshell import ZMQInteractiveShell
 
 try:
@@ -347,6 +347,7 @@ class IPythonKernel(KernelBase):
         user_expressions=None,
         allow_stdin=False,
         *,
+        cell_meta=None,
         cell_id=None,
     ):
         """Handle code execution."""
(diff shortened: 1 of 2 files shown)


jjvraw:
Hi,

I've noticed that all checks have passed on the PR except for "Test Prereleases". This seems to be related to an [existing issue in pytest-asyncio](https://github.com/pytest-dev/pytest-asyncio/issues/679), which might be causing the failure.

I am happy to work further on any specific actions or suggestions.

Thank you

blink1073 (approved):
Thank you!  I agree we should keep `cell_id`, but the private method name can be changed.

PR #1178, opened 2023-12-04
Update pytest commands in README

ianthomas23:
This PR updates the two `pytest` commands in the README to be the same as those in CI and the various `hatch` commands. It replaces `pytest ipykernel` with `pytest`. An alternative would be to use the more explicit `pytest tests`, but this seems unnecessary.

Problem discovered when following the developer instructions for the first time.

This is my first contribution to `ipykernel` :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Follow the instructions from `Installation from source`.
 and then from the root directory
 
 ```bash
-pytest ipykernel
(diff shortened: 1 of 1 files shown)


davidbrochart (approved):
Thanks @ianthomas23, and welcome to ipykernel!

PR #1184, opened 2023-12-21
Enable `ProactorEventLoop` on windows for `ipykernel`

NewUserHa:
On windows `ipykernel` currently uses `SelectorEventLoop` which is slower than `ProactorEventLoop`. this issue causes many apps like `vscode` use `ipykernel` slow.

Currently, `ipykernel` requires `python` version >=3.8 and `tornado` version >= 6.1 (which can already work with `ProactorEventLoop`)

So this issue can be fixed now

related:
https://github.com/tornadoweb/tornado/issues/2608
https://www.tornadoweb.org/en/stable/releases/v6.1.0.html
https://github.com/ipython/ipykernel/commit/f40d74afe52c4f45749860d293b6f55ac7e0e7d4

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -624,43 +624,6 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp, ConnectionFileMix
         handler.setFormatter(formatter)
         logger.addHandler(handler)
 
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Oh wow, the test suite is about twice as fast, thank you!

PR #1186, opened 2023-12-21
Make outputs go to correct cell when generated in threads/asyncio

krassowski:
### References

- Fixes https://github.com/jupyterlab/jupyterlab/issues/3936
- Fixes https://github.com/jupyter-widgets/ipywidgets/issues/2358

### Code changes

- Use [`contextvars`](https://docs.python.org/3/library/contextvars.html) to retrieve the parent header of emitting asyncio task
- Use thread identity to retrieve the parent header of emitting thread
- Store iostream buffer per parent header rather than in a single buffer

### User facing changes

| Before | After |
|--|--|
| before | after |

### Backward-incompatible changes

None

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -451,7 +451,8 @@ class OutStream(TextIOBase):
             "parent_header"
         )
         self._parent_header.set({})
-        self._thread_parents = {}
+        self._thread_to_parent = {}
+        self._thread_to_parent_header = {}
         self._parent_header_global = {}
         self._master_pid = os.getpid()
         self._flush_pending = False
@@ -509,7 +510,13 @@ class OutStream(TextIOBase):
         except LookupError:
             try:
                 # thread-specific
-                return self._thread_parents[threading.current_thread().ident]
+                identity = threading.current_thread().ident
+                # retrieve the outermost (oldest ancestor,
+                # discounting the kernel thread) thread identity
+                while identity in self._thread_to_parent:
+                    identity = self._thread_to_parent[identity]
+                # use the header of the oldest ancestor
+                return self._thread_to_parent_header[identity]
             except KeyError:
                 # global (fallback)
                 return self._parent_header_global
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -5,6 +5,7 @@
 
 import asyncio
 import atexit
+import contextvars
 import io
 import os
 import sys
@@ -12,7 +13,7 @@ import threading
 import traceback
 import warnings
 from binascii import b2a_hex
-from collections import deque
+from collections import defaultdict, deque
 from io import StringIO, TextIOBase
 from threading import local
 from typing import Any, Callable, Deque, Dict, Optional
@@ -412,7 +413,7 @@ class OutStream(TextIOBase):
         name : str {'stderr', 'stdout'}
             the name of the standard stream to replace
         pipe : object
-            the pip object
+            the pipe object
         echo : bool
             whether to echo output
         watchfd : bool (default, True)
@@ -446,13 +447,18 @@ class OutStream(TextIOBase):
         self.pub_thread = pub_thread
         self.name = name
         self.topic = b"stream." + name.encode()
-        self.parent_header = {}
+        self._parent_header: contextvars.ContextVar[Dict[str, Any]] = contextvars.ContextVar(
+            "parent_header"
+        )
+        self._parent_header.set({})
+        self._thread_parents = {}
+        self._parent_header_global = {}
         self._master_pid = os.getpid()
         self._flush_pending = False
         self._subprocess_flush_pending = False
         self._io_loop = pub_thread.io_loop
         self._buffer_lock = threading.RLock()
-        self._buffer = StringIO()
+        self._buffers = defaultdict(StringIO)
         self.echo = None
         self._isatty = bool(isatty)
         self._should_watch = False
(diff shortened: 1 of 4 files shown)


davidbrochart (approved):
Impressive PR @krassowski, that's great.
Just to confirm, this wouldn't work with threads that are not created from Python, right? For instance in a C extension.

krassowski:
Yes, writing from threads which were created outside of Python will still write to the most recently executed cell, as will writing directly to OS-level file descriptor corresponding to stdout/stderr.

ipykernel/ipkernel.py
```diff
@@ -706,6 +724,58 @@ def do_clear(self):
...
+            This is needed because there is no "start" hook yet, but there
+            might be one in the future: https://bugs.python.org/issue14073
+            """
+
+            threading_start(self)
+            for stream in [stdout, stderr]:
+                if isinstance(stream, OutStream):
+                    stream._thread_parents[self.ident] = parent_header
+
+        threading.Thread.start = start_closure  # type:ignore[method-assign]
```
davidbrochart, line 751: I was wondering if there could be a way to not monkey-patch `threading.Thread.start`.
In [akernel](https://github.com/davidbrochart/akernel/blob/71b0b67bb400ff5ce229320662c89f30e02396cd/akernel/kernel.py#L464) it's the `print` function that is monkey-patched, so this PR goes further because it works for lower-level writes to `stdout`/`stderr`, but with monkey-patching there is always a risk that another library changes the same object, and we cannot know that our change will be the last applied.

krassowski: My understanding as per https://bugs.python.org/issue14073 is that there is no other way, but we could mention our use case as another situation motivating introduction of start/exit hooks/callbacks to threads (which is something Python committers have considered in the past but I presume it was not a sufficiently high priority until now).

maartenbreddels, line 751: I don't think it thread-safe. If a thread starts a new thread, that new thread will start outputting in the last executed cell.
I've taken a different approach in Solara, where I patch the ctor and run from the start:
https://github.com/widgetti/solara/blob/0ac4767a8c5f8c8b221cafe41fa9ac36270adbe8/solara/server/patch.py#L336

I think this approach might work better.

maartenbreddels (commented):
Awesome work!

ipykernel/ipkernel.py
maartenbreddels: We crossed 'comments', see my other thread on this exact line for (what I think is) a better solution.

maartenbreddels: To summarize the strategy, the ctor (`__init__`) would take a reference to the parent_header, and in run, that parent header is then set in a thread_local storage (in your case `_thread_parents`).

ipykernel/iostream.py
```diff
@@ -446,13 +447,18 @@ def __init__(
         self.pub_thread = pub_thread
         self.name = name
         self.topic = b"stream." + name.encode()
-        self.parent_header = {}
+        self._parent_header: contextvars.ContextVar[Dict[str, Any]] = contextvars.ContextVar(
+            "parent_header"
+        )
+        self._parent_header.set({})
+        self._thread_parents = {}
+        self._parent_header_global = {}
```
maartenbreddels, line 455: Curious why the ContextVar and the _thread_parents dict? Is a threading.local not more standard?

krassowski: Because contextvar works well for asyncio edge cases. There is some more explanation in the PEP 567, but the gist is:

> Thread-local variables are insufficient for asynchronous tasks that execute concurrently in the same OS thread. Any context manager that saves and restores a context value using `threading.local()` will have its context values bleed to other code unexpectedly when used in async/await code.

That said, I will take another look at using `threading.local` instead of `_thread_parents`.

ipykernel/ipkernel.py
krassowski: > If a thread starts a new thread, that new thread will start outputting in the last executed cell.

Correct (there is no such a problem with asyncio side of things).

> I think this approach might work better.

Thank you for the link, I will take a look!

ipykernel/iostream.py
maartenbreddels: I understand that we need ContextVar now, because also when we create new tasks we want to output to the right output cell, didn't think of that!

I think if we use ContextVar, we do not need the thread local storage, it's a superset of threading.local. In combination with overriding the Thread ctor, we can drop _thread_parents.

ipykernel/ipkernel.py
krassowski: Taking a closer look at solara, `current_context` takes the role of `_thread_parents`, right? This does not seem to differ that much. Your approach of patching earlier on initialization rather than startup has the advantage that it will work with:

```python
from threading import Thread
from time import sleep

def child_target():
    for i in range(iterations):
        print(i, end='', flush=True)
        sleep(interval)

def parent_target():
    thread = Thread(target=child_target)
    sleep(interval)
    thread.start()

Thread(target=parent_target).start()
```

but still not with:

```python
def parent_target():
    sleep(interval)
    Thread(target=child_target).start()

Thread(target=parent_target).start()
```

do I see this right?

krassowski:
As of now this PR also handles nested threads (started after execution of other cells) as well.

ipykernel/ipkernel.py
krassowski: In the end my implementation converged to overriding the same methods as yours after all in https://github.com/ipython/ipykernel/pull/1186/commits/e1258deac3e0fce7515efa29e444a0c4889def04 :)

krassowski:
Hi folks, just bumping this PR as there was no further review/comments in the past three weeks. Any suggestions, or is it good to merge?

blink1073:
I'll merge tomorrow if there are no further comments.

Issue #1188, opened 2024-01-11
CI failing with INTERNALERROR when skipping tests


ianthomas23:
Latest [github action run](https://github.com/ipython/ipykernel/actions/runs/7482330236) failing for `test_without_debugpy` with `INTERNALERROR> Skipped: could not import 'debugpy': No module named 'debugpy'` and on Windows with `INTERNALERROR> Skipped: skipping tests on windows`. I can reproduce this locally with `pytest-asyncio` 0.23.3 but it is OK with 0.23.2. There is a related issue pytest-dev/pytest-asyncio#729.

PR #1189, opened 2024-01-11
Pin pytest-asyncio to 0.23.2

ianthomas23:
Seeing if #1188 is fixed by pinning `pytest-asyncio`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,7 @@ test = [
     "flaky",
     "ipyparallel",
     "pre-commit",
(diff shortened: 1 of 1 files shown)


ianthomas23:
Ah, I see that @blink1073 is discussing this with the `pytest-asyncio` maintainers so there will probably be a fix for this at their end and something like this PR won't be necessary.

blink1073:
Thanks!  The prerelease test will fail until we get this sorted out which I think is fair.

PR #1200, opened 2024-01-27
Fix handling of "silent" in execute request

Haadem:
Add default value to silent parameter. Aligns with [Jupyter_core](https://jupyter-client.readthedocs.io/en/stable/messaging.html#messages-on-the-shell-router-dealer-channel).


blink1073 (approved):
Thank you!

PR #1203, opened 2024-02-05
Do git ignore of /node_modules/.cache

petervandenabeele:
Without this, running of pre-commit, adds this file:

node_modules/.cache/prettier/.prettier-caches/8b311690c76070c4437a2a5b4d0e18721f679cd9.json

PR #1204, opened 2024-02-05
fix: on exception, return a 0, so that the "sum" still computes

petervandenabeele:
This fixes a bug where after a while in a Jupyter notebook, we start to see many log lines like this:

```
[IPKernelApp] ERROR | Exception in control handler:
  File ".../lib/python3.8/site-packages/ipykernel/kernelbase.py", line 336, in process_control
    await result
  File ".../lib/python3.8/site-packages/ipykernel/kernelbase.py", line 998, in usage_request
    reply_content["kernel_memory"] = sum(
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

Since the "sum" expects a number (float or int) as elements given to the "sum", we should pass "0" (zero) and not None when we that process id does not exist anymore.

This was tested in an actually running notebook and it fixed the problem:
* no more Exceptions like the one reported above
* correct memory stats where seen with the Kernel Usage plug-ins

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -1049,7 +1049,7 @@ class Kernel(SingletonConfigurable):
         # Avoid littering logs with stack traces
         # complaining about dead processes
         except BaseException:
-            return None
+            return 0
 
     async def usage_request(self, stream, ident, parent):
         """Handle a usage request."""


petervandenabeele:
This is a bug fix and needs the "bugfix" Label. As an external contributor, I am unable to set it.

Also, I am pinging @echarles since we discussed this out-of-band and Eric is marked in git as the previous contributor that touched this code.

Many thanks for your attention (I hacked it locally here in the site-packages and it seems to work for me :-)

echarles (approved):
LGTM Thx @petervandenabeele

echarles:
@blink1073 📝 test_lint → ❌ failure [required to succeed] is failing, not sure about that one.

petervandenabeele:
There seems to be a problem with a linting ignore that is not correctly set-up ?

https://github.com/ipython/ipykernel/actions/runs/7785980351/job/21229753213?pr=1204#step:4:75

```
ipykernel/kernelapp.py:678: error: Unused "type: ignore" comment 
[unused-ignore]
                pdb.set_trace = debugger.set_trace  # type:ignore[assignme...
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Installing missing stub packages:
/home/runner/.cache/pre-commit/repo3ioeb974/py_env-python3.11/bin/python -m pip install types-Pygments types-colorama types-decorator types-psutil types-pycurl types-python-dateutil types-six


Found 1 error in 1 file (checked 43 source files)

Error: Process completed with exit code 1.
```

petervandenabeele:
FYI: the same "linting" problem also occurs in another PR I just made today (5 Feb 2024) and that only added 1 line to .gitignore:

https://github.com/ipython/ipykernel/pull/1203

blink1073:
The lint failures comes from a version update in the linter, I'll address in a separate PR.

blink1073 (approved):
Thank you!

PR #1207, opened 2024-02-07
Re-enable skipped debugger test

ianthomas23:
One of the tests in `test_debugger.py` has been skipped for python >= 3.10 since it was added (#829). The test runs and passes fine for me on Ubuntu using python 3.10.1 and 3.10.13, so here I am re-enabling it to run it on the whole CI matrix.


blink1073 (approved):
Thank you!

PR #1209, opened 2024-02-10
Unpin pytest-asyncio and update ruff config

blink1073:


ipykernel/pickleutil.py
```diff
@@ -2,25 +2,23 @@
...
-
 import copy
 import pickle
 import sys
+import typing
+import warnings
 from types import FunctionType
 
 # This registers a hook when it's imported
-from ipyparallel.serialize import codeutil
```
bollwyvl, line 20: I think this "dead code" side-effect is showing up here:

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=885003&view=logs&j=a70f640f-cc53-5cd3-6cdc-236a1aa90802&t=6119ccbe-9301-594f-7c27-f792b80a7fcc&l=1235

```
_______________________________ test_no_closure _______________________________

    def test_no_closure():
        @interactive
        def foo():
            a = 5
            return a
    
>       pfoo = dumps(foo)

tests\test_pickleutil.py:28: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = <function test_no_closure.<locals>.foo at 0x0000018BD7A07A60>

    def dumps(obj):
>       return pickle.dumps(can(obj))
E       TypeError: cannot pickle code objects


```

blink1073: Ah, I see, good catch!

PR #1210, opened 2024-02-13
Remove control queue

ianthomas23:
Currently when the control channel receives a message on the control stream, it puts it in a `tornado.queues.Queue`, and messages are popped off the queue one a time to be handled using `process_control`. I don't think the queue is necessary as the ZMQ socket/stream queues messages itself.

This PR removes the control queue, simplifying the code. The functions `_flush_control_queue` and `poll_control_queue` are removed as there is no queue to flush or poll. I assume the latter is considered public by its naming convention, but I don't know if any downstream libraries will be relying on its existence.

There are other queues used for the control channel and debugger that I have not looked at yet.

I have ensured that only one control message is handled at a time, to preserve the current behaviour. This uses a boolean flag in `dispatch_control`. This is a little strange as the use of `ZMQStream.on_recv` allows us to concurrently handle multiple control messages that we then have to suppress. There is future work to be done here to perhaps use the ZMQ socket rather than the stream, and jupyter-server/jupyter_server#1362 and jupyter/jupyter_client#997 are relevant.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -498,7 +498,7 @@ class Kernel(SingletonConfigurable):
 
         if self.control_thread is None and self.control_stream is not None:
             # If there isn't a separate control thread then this main thread handles both shell
-            # and control messages. Before processing a shell message need to flush all control
+            # and control messages. Before processing a shell message we need to flush all control
             # messages and allow them all to be processed.
             await asyncio.sleep(0)
             self.control_stream.flush()
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -269,9 +269,6 @@ class Kernel(SingletonConfigurable):
         "usage_request",
     ]
 
-    # Flag to ensure a single control request is processed at a time.
-    _block_control = False
-
     def __init__(self, **kwargs):
         """Initialize the kernel."""
         super().__init__(**kwargs)
@@ -298,17 +295,8 @@ class Kernel(SingletonConfigurable):
 
     async def dispatch_control(self, msg):
         # Ensure only one control message is processed at a time
-        while self._block_control:
-            await asyncio.sleep(0)
-
-        self._block_control = True
-
-        try:
+        async with asyncio.Lock():
             await self.process_control(msg)
-        except:
-            raise
-        finally:
-            self._block_control = False
 
     async def process_control(self, msg):
         """dispatch control requests"""
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -5,7 +5,6 @@
 from __future__ import annotations
 
 import asyncio
-import concurrent.futures
 import inspect
 import itertools
 import logging
@@ -270,6 +269,9 @@ class Kernel(SingletonConfigurable):
         "usage_request",
     ]
 
+    # Flag to ensure a single control request is processed at a time.
+    _block_control = False
+
     def __init__(self, **kwargs):
         """Initialize the kernel."""
         super().__init__(**kwargs)
@@ -289,48 +291,24 @@ class Kernel(SingletonConfigurable):
         for msg_type in self.control_msg_types:
             self.control_handlers[msg_type] = getattr(self, msg_type)
 
-        self.control_queue: Queue[t.Any] = Queue()
-
         # Storing the accepted parameters for do_execute, used in execute_request
         self._do_exec_accepted_params = _accepts_parameters(
             self.do_execute, ["cell_meta", "cell_id"]
         )
 
-    def dispatch_control(self, msg):
-        self.control_queue.put_nowait(msg)
+    async def dispatch_control(self, msg):
+        # Ensure only one control message is processed at a time
+        while self._block_control:
+            await asyncio.sleep(0)
 
-    async def poll_control_queue(self):
-        while True:
-            msg = await self.control_queue.get()
-            # handle tracers from _flush_control_queue
-            if isinstance(msg, (concurrent.futures.Future, asyncio.Future)):
-                msg.set_result(None)
-                continue
+        self._block_control = True
+
+        try:
             await self.process_control(msg)
-
-    async def _flush_control_queue(self):
-        """Flush the control queue, wait for processing of any pending messages"""
-        tracer_future: concurrent.futures.Future[object] | asyncio.Future[object]
-        if self.control_thread:
-            control_loop = self.control_thread.io_loop
-            # concurrent.futures.Futures are threadsafe
-            # and can be used to await across threads
-            tracer_future = concurrent.futures.Future()
-            awaitable_future = asyncio.wrap_future(tracer_future)
-        else:
-            control_loop = self.io_loop
-            tracer_future = awaitable_future = asyncio.Future()
-
-        def _flush():
-            # control_stream.flush puts messages on the queue
-            if self.control_stream:
-                self.control_stream.flush()
-            # put Future on the queue after all of those,
-            # so we can wait for all queued messages to be processed
-            self.control_queue.put(tracer_future)
-
-        control_loop.add_callback(_flush)
-        return awaitable_future
+        except:
+            raise
+        finally:
+            self._block_control = False
 
     async def process_control(self, msg):
         """dispatch control requests"""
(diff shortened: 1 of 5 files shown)


tests/test_kernel.py
```diff
@@ -597,6 +598,43 @@ def test_control_thread_priority():
...
+            kc.control_channel.send(msg)
+
+        # Get replies
+        replies = [get_reply(kc, msg_id, channel="control") for msg_id in msg_ids]
+
+        # Check messages are processed in order, one at a time, and of a sensible duration.
+        previous_end = None
+        for reply, sleep in zip(replies, sleeps):
+            start = datetime.fromisoformat(reply["metadata"]["started"])
+            end = reply["header"]["date"]
```
ianthomas23, line 627: This is interesting, in a bad way. The type of `end` differs depending on how the tests are run and is either a `datetime` or a `string`. Using
```
pytest -v tests/test_kernel.py -k control
```
they are `datetime`s (which is what I was expecting), and using
```
pytest -v tests -k control
```
they are strings. So there is something different happening depending on how the tests are invoked.

ipykernel/kernelbase.py
```diff
@@ -289,48 +291,24 @@ def __init__(self, **kwargs):
...
-            # and can be used to await across threads
-            tracer_future = concurrent.futures.Future()
-            awaitable_future = asyncio.wrap_future(tracer_future)
-        else:
-            control_loop = self.io_loop
-            tracer_future = awaitable_future = asyncio.Future()
+    async def dispatch_control(self, msg):
+        # Ensure only one control message is processed at a time
+        while self._block_control:
+            await asyncio.sleep(0)
```
davidbrochart, line 302: This looks like it could be implemented with an [asyncio.Lock](https://docs.python.org/3/library/asyncio-sync.html#lock).

ianthomas23: Yes, that looks like a much more elegant approach.

ianthomas23: Done.

tests/test_kernel.py
ianthomas23: I've worked out why the type of the header's date is different depending on how the tests are run. The cause is this line in `ipyparallel`:
https://github.com/ipython/ipyparallel/blob/5e13f5664d27b60338542c438b7d8d144546dfce/ipyparallel/client/client.py#L58
which monkey patches `jupyter_client.session.extract_dates` to be a no-op. If you are specific about which test to run then this line in `ipyparallel` is not run, but if you scan all the tests then it is run and datetimes are not deserialized from strings to datetimes.

blink1073: That looks like a bug in `ipyparallel`, mind reporting it there?

ianthomas23: It turns out that I cannot reproduce this problem locally with the latest `pytest-asyncio 0.23.5`, so for the moment I've just rebased this PR against `main` and will monitor it.

blink1073: Sounds good

ianthomas23:
The tests pass for this except for the downstream `ipyparallel` tests which are failing in various abort scenarios. It looks to me that `ipyparallel` doesn't use a separate control channel thread and to support that here I will need to add extra code (if the shell and control channels share a thread) to force the control messages to be processed before the shell ones. This will bring back some of the previous `_flush_control_queue` functionality, but in a different way.

ipykernel/kernelbase.py
```diff
@@ -528,6 +492,19 @@ async def process_one(self, wait=True):
                 t, dispatch, args = self.msg_queue.get_nowait()
             except (asyncio.QueueEmpty, QueueEmpty):
                 return
+
+        if self.control_thread is None and self.control_stream is not None:
+            # If there isn't a separate control thread then this main thread handles both shell
+            # and control messages. Before processing a shell message need to flush all control
```
davidbrochart, line 498:
```suggestion
            # and control messages. Before processing a shell message we need to flush all control
```

ipykernel/kernelbase.py
```diff
@@ -528,6 +492,19 @@ async def process_one(self, wait=True):
...
+            # messages and allow them all to be processed.
+            await asyncio.sleep(0)
+            self.control_stream.flush()
+
+            socket = self.control_stream.socket
+            while socket.poll(1):
```
davidbrochart, line 504: This is blocking for 1ms, right? I think it should just check if there is a message:
```suggestion
            while socket.poll(0):
```

ianthomas23: I'd like to, but if I do that the `ipyparallel` tests fail locally for me.

ianthomas23:
To get `test_debugger.py` to pass the minimum dependencies CI run I needed to increase the minimum version of `pyzmq` from 24 to 25. In turn this required a bump of `jupyter-client` and `tornado` minimum versions (jupyter/jupyter_client#915). Summary of min version changes:

| Dependency | Previous min version | Current min version | Release date of current min version |
| --- | --- | -- | --- |
| jupyter-client | 6.1.12 | 8.0.0 | 2023-01-26 |
| pyzmq | 24.0.0 | 25.0.0 | 2023-01-12 |
| tornado | 6.1.0 | 6.2.0 | 2022-07-03 |

I don't know if these changes will be considered too aggressive too soon.

blink1073:
Hmm, perhaps instead I can cut a version of 7.x that supports pyzmq 25?

blink1073:
Or, perhaps we wait and fold this in to an ipykernel 7.x release with #1079?

davidbrochart:
> Or, perhaps we wait and fold this in to an ipykernel 7.x release with #1079?

I was thinking about that, but would this PR still make sense once #1079 is in?

ianthomas23:
> I was thinking about that, but would this PR still make sense once #1079 is in?

I see that #1079 still uses a queue for the control channel but it is no longer a Tornado queue of course. So this probably isn't needed after that, but the idea of simplifying by removing the queue might still be worth looking at.

If #1079 is near merging then we should ignore this for now. But it looks like there is quite a lot of work to do for Windows and downstream projects?

davidbrochart:
Sorry Ian if your work in this PR ends up not being merged, I was aware of #1079 of course but Steve picked it up recently and I don't know if it's close to being merged, although he seems to be progressing fast.

ianthomas23:
> Sorry Ian if your work in this PR ends up not being merged, I was aware of #1079 of course but Steve picked it up recently and I don't know if it's close to being merged, although he seems to be progressing fast.

No problem. This has been my first attempt at a substantial change to `ipykernel` and even if not merged it has been a good learning experience to get used to the CI and in particular the impact on downstream projects.

blink1073:
I had a go at the 6.x branch of jupyter_client, but I don't think it is revive-able.  I think we can merge this and consider `main` to be 7.x, and focus on folding in #1079 as well.

tests/test_kernel.py
```diff
@@ -597,6 +598,47 @@ def test_control_thread_priority():
...
+                # https://github.com/python/cpython/issues/80010
+                start_str = start_str[:-1] + "+00:00"
+            start = datetime.fromisoformat(start_str)
+            end = reply["header"]["date"]  # Already a datetime
+
+            if previous_end is not None:
+                assert start > previous_end
+            previous_end = end
+
+            assert end >= start + timedelta(seconds=sleep)
```
Carreau, line 639: This fails downstream in IPython with:

TypeError: '>=' not supported between instances of 'str' and 'datetime.datetime'

ianthomas23: These are the same symptoms that I observed at https://github.com/ipython/ipykernel/pull/1210#discussion_r1487916907. `ipyparallel` unfortunately monkey-patches jupyter_clients conversion of strings to datetimes so it is a no-op, and some versions of `pytest-asyncio` don't fully isolate the tests so sometimes some of these start and end dates come through as strings.

In this PR updating `pytest-asyncio` to 0.23.5 worked, but this is evidently not sufficient as the IPython failure already uses `pytest-asyncio-0.23.5`.

I thing the sensible approach here is to assume that in this test sometimes, outside of our direct control, the dates are strings and convert them to datetimes. Otherwise this test is not testing what it is supposed to, it is testing the types of some fields returned in messages. I'll submit a PR here ASAP.

PR #1229, opened 2024-04-04
Set all min deps

blink1073:


maartenbreddels:
did you use uv to find this out?

blink1073:
Yes, https://github.com/jupyterlab/maintainer-tools/blob/8e17c2d7e4c886a75685a7a1f609b4b5201fc9c6/.github/actions/base-setup/action.yml#L144

PR #1233, opened 2024-04-22
Raise KeyboardInterrupt when handling SIGINT by default

davidbrochart:
This fixes jupyter-client's [test_signal_kernel_subprocesses](https://github.com/jupyter/jupyter_client/blob/19dbf1b69247a1cc14b5bb1e4e7523f5b52380ae/tests/test_kernelmanager.py#L197).
`shell_is_awaiting` and `shell_is_blocking` are not set in the kernel base class, but in ipykernel. By default, interrupting a kernel should raise `KeyboardInterrupt`. A kernel running an async function is really language-dependent. If a kernel supports async execution, it should set `shell_is_awaiting` and handle the `shell_interrupt` queue.


ianthomas23 (approved):
I've manually confirmed that this does indeed fix some of the downstream jupyter_client test failures.

Issue #1240, opened 2024-05-22
Incorrect version string on main


mlucool:
When I installed from main, I found it confusing that https://github.com/ipython/ipykernel/blob/main/ipykernel/_version.py#L8 still has 6.23.3 both because 6.23.4 was released, and main I believe should be considered a major breaking version (so I would have expected 7.0.0 or something like that). Happy to make a PR - but was unsure what people wanted here.

Feel free to close if you prefer not to update right now.

davidbrochart:
Yes, it can be updated to `7.0.0`. Feel free to make a PR, thanks!

PR #1241, opened 2024-05-23
Update version to 7.0.0

mlucool:
Main contains semver major changes. Updating version to reflect this.

Fixes: #1240


davidbrochart (approved):
Thanks @mlucool.

ianthomas23:
I think the incorrect version number may be causing CI failures in other PRs in the `check_release` action for example, and may also be part of the problem we experienced yesterday looking at a 6.29.5 release. So let's merge and see if it helps.

Thanks @mlucool.

PR #1242, opened 2024-05-28
Catch IPython 8.24 DeprecationWarnings

s-t-e-v-e-n-k:
test_pylab with IPython 8.24 and above raises DeprecationWarnings about backend{s,2gui} being deprecated, causing test failures. Add them into the IPython list so they are ignored.


blink1073 (approved):
Thank you!

PR #1246, opened 2024-06-07
Remove bad pipe message log

deshanxiao:
This PR is aimed to remove a log when IOPubThread received a mismatch uuid message.

These mismatch messages appear to originate from some firewall and network related components. Jupyter users don't seem to need this information, however it is displayed on the UX.

I search ["Bad pipe message"](https://github.com/search?q=%22Bad+pipe+message%22&type=code) in whole github repos and there are many ipynb files contains this error.

Some similar issue:
https://github.com/spyder-ide/spyder/issues/20591
https://github.com/ipython/ipykernel/issues/1107

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -215,7 +215,7 @@ class IOPubThread:
         if not self._pipe_flag or not self._is_master_process():
             return
         if msg[0] != self._pipe_uuid:
-            print("Bad pipe message: %s", msg, file=sys.__stderr__)
+            # print("Bad pipe message: %s", msg, file=sys.__stderr__)
             return
         self.send_multipart(msg[1:])
 


ipykernel/iostream.py
```diff
@@ -215,7 +215,7 @@ async def _handle_pipe_msg(self, msg=None):
         if not self._pipe_flag or not self._is_master_process():
             return
         if msg[0] != self._pipe_uuid:
-            print("Bad pipe message: %s", msg, file=sys.__stderr__)
+            # print("Bad pipe message: %s", msg, file=sys.__stderr__)
```
minrk, line 218: These messages do indicate a problem, and should still be logged, not entirely suppressed. The problem comes from the change  that `sys.__stderr__` is now duplicated and forwarded to users, when this message is meant to only. What we really want to write to is `sys.stderr._original_stdstream_copy`.

I see two options:

1. when we capture the low-level `__stderr__`, we could replace `sys.__stderr__` with the new captured fd, so statements like this do what we originally meant them to (only write to stderr, not captured and forwarded), or
2. keep capturing `__stderr__` and explicitly write to the "truly really actually only original stderr" FD for messages like these.

Issue #1119, opened 2023-05-23
ipykernel unconditionally calls `sys.stdout.flush()`


twizmwazin:
ipykernel unconditionally calls `sys.stdout.flush()` and `sys.stderr.flush()`. This is a problem when either `sys.stdout` or `sys.stderr` is `None`, such as in `pythonw.exe` on windows, or Pyinstaller's windowed mode. All of these checks should be garded with a not-None check.

gregory-shklover:
same issue here

gregory-shklover:
Submitted PR: https://github.com/ipython/ipykernel/pull/1247

PR #1247, opened 2024-06-10
Fixed error accessing sys.stdout/sys.stderr when those are None

gregory-shklover:
Fix for issue reported in https://github.com/ipython/ipykernel/issues/1119 (sys.stdout/sys.stderr could be None on Windows, when running in GUI mode)


gregory-shklover:
Hi. Any updates on this pull request?

PR #1249, opened 2024-06-13
Kernel subshells (JEP91) implementation

ianthomas23:
This is the implementation of the kernel subshells JEP (jupyter/enhancement-proposals#91). It follows the latest commit ([1f1ad3d](https://github.com/jupyter/enhancement-proposals/pull/91/commits/1f1ad3d43f24299efb7cc7879cefeee57277906c)) with the addition of a `%subshell` magic command that is useful for debugging. To try this out I have a [JupyterLab branch](https://github.com/ianthomas23/jupyterlab/tree/jep91_demo) that talks to this branch and is most easily tried out using https://mybinder.org/v2/gh/ianthomas23/jupyterlab/jep91_demo?urlpath=lab; once the mybinder instance has started, open the `subshell_demo_notebook.ipynb` and follow the instructions therein.

The idea is that this is mergeable as it is now, it is backward compatible in that it does not break any existing use of `ipykernel`  (subject to CI confirmation). There are some ramifications of the protocol additions (outlined below) that will need addressing eventually, but I consider these future work that can be in separate PRs. 

# Outline of changes
1. The parent subshell (i.e. the main shell) runs in the main thread.
2. Each new subshell runs in a separate thread.
3. There is a new thread that deals with all communication on the shell channel, previously this was performed in the main thread.
4. Communication between the shell channel thread and other threads is performed using ZMQ inproc pair sockets, which are essentially shared memory and avoid the use of thread synchronisation primitives.
5. Incoming shell messages are handled by the shell channel thread which extracts the `subshell_id` from the message and passes it on to the correct subshell.
6. Subshells are created and deleted via messages sent on the control channel. These are passed to the shell channel thread via inproc pair sockets so that the `SubshellManager` in the shell channel thread is responsible for subshell lifetimes.

# Example scenario

Here is an example of the communication between threads when running a long task in the parent subshell (main thread) and whilst this is running a child subshell is created, used, and deleted.

```mermaid
sequenceDiagram
    participant client as Client
    participant control as Control thread
    participant shell as Shell channel thread
    participant main as Main thread

    client->>+shell: Execute request (main shell)
    shell->>-main: Execute request (inproc)
    activate main

    client->>+control: Create subshell request
    control->>-shell: Create subshell request (inproc)
    activate shell
    create participant subshell as Subshell thread
    shell-->>subshell: Create subshell thread

    shell->>control: Create subshell reply (inproc)
    deactivate shell
    activate control
    control->>-client: Create subshell reply

    client->>+shell: Execute request (subshell)
    shell->>-subshell: Execute request (inproc)
    activate subshell

    subshell->>shell: Execute reply (inproc)
    deactivate subshell
    activate shell
    shell->>-client: Execute reply (subshell)

    client->>+control: Delete subshell request
    control->>-shell: Delete subshell request (inproc)
    activate shell
    destroy subshell
    shell-->>subshell: Delete subshell thread

    shell->>control: Delete subshell reply (inproc)
    deactivate shell
    activate control
    control->>-client: Delete subshell reply

    main->>shell: Execute reply (inproc)
    deactivate main
    activate shell
    shell->>-client: Execute reply (main shell)
```

# Future work

## ipykernel

1. Shell channel thread deserialises ~the whole~ some of the message to get the `subshell_id`. Ideally it would only deserialise the header.  May need changes in Jupyter Client.
2. Signalling a subshell to stop uses a `threading.Event` following the existing `anyio` implementation which requires an extra thread per `Event`. It would be nice if this could be changed so a subshell is a single thread not two.
3. Execution count. Should either be a separate count per subshell or a single count for a kernel. Needs a decision and changes in IPython as is currently not atomic.
4. History. Related to item 2 above.
5. `input()` on more than one subshell at the same time run but do not store correctly.
6. Debugger use needs investigating.
7. Busy/idle status needs investigating. Should there, as now, be separate status for each subshell, or the concept of kernel (i.e. any subshell) busy status? This issue is much wider than subshells as it includes status of the control channel, and how Jupyter Server should track status (jupyter-server/jupyter_server#1429).
8. Use of display hooks for e.g. Matplotlib. Should these be on the parent subshell, or child subshells too?

## JupyterLab

The [JupyterLab branch](https://github.com/ianthomas23/jupyterlab/tree/jep91_demo) I am using to demo this isn't really intended to be merged. But if it was, it needs:
1. Check `kernel_info` to see if subshells are supported.
2. Delete subshell when close a subshell's `ConsolePanel`.
3. Report subshell IDs in tree view?
4. Display of subshell busy/idle status.

(Edited for clarity)


ianthomas23:
The [kernel subshells JEP](https://github.com/jupyter/enhancement-proposals/pull/91) has been accepted and as this is the reference implementation for that I am asking for someone to review this.

ipykernel/subshell_manager.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        assert current_thread().name == "Shell channel"
+
+        while True:
+            for socket, _ in self._poller.poll(0):
+                msg = await socket.recv_multipart(copy=False)
+                self._shell_socket.send_multipart(msg)
+
+            # Yield to other tasks.
+            await sleep(0)
```
davidbrochart, line 134: This loop leads to 100% CPU usage, we need to find another way.

davidbrochart (changes requested):

ipykernel/subshell_manager.py
ianthomas23: Interesting. On my macOS dev machine this is fine, but I've confirmed that it is a problem on Linux.

ianthomas23: I've changed the implementation in 97e3e91e2a. I've removed the `sleep(0)` and `poll(0)` calls and now use an async zmq `Poller` in an `anyio` task. When a subshell is created or deleted this task is cancelled using an `anyio.Event` are rescheduled with the updated list of subshells. With this I no longer see 100% CPU load on Linux or macOS.

ianthomas23: I've changed the implementation of this piece of code again. My use of `await poller.poll()` turned out to cause problems on python < 3.10 and on Windows. Now I am avoiding use of `zmq.Poller` and instead using a separate anyio task for each subshell to listen for reply messages and send them out to the client via the shell channel. This seems to be more robust on older python and Windows. I've also replaced the use of an `anyio.Event` with a memory object stream instead (essentially an anyio async queue).

ianthomas23:
Local tests are showing occasional errors in `tests/test_message_spec.py::test_execute_stop_on_error` which I need to look at.

pyproject.toml
```diff
@@ -188,6 +188,7 @@ filterwarnings= [
   # https://github.com/python-trio/trio/issues/3053
   "ignore:The `hash` argument is deprecated in favor of `unsafe_hash` and will be removed in or after August 2025.",
 ]
+asyncio_default_fixture_loop_scope = "function"
```
davidbrochart, line 191: I think this comes from `pytest-asyncio`, but don't we use AnyIO's pytest plugin?

ianthomas23: It does come from `pytest-asyncio`, and evidently it is used via some means as there are warnings before this change and not after. But I am going to revert this change as this option is not available for the minimum dependencies run, and I can tolerate a warning.

ianthomas23:
I had missed a direct use of the `shell_socket` to send an `abort_reply` message when it should have been sent via the `SubshellManager`. With that fixed, all the local CI is now passing.

ipykernel/thread.py
```diff
@@ -0,0 +1,40 @@
+"""Base class for threads."""
+from threading import Event, Thread
+
+from anyio import create_task_group, run, to_thread
+
+
+class BaseThread(Thread):
+    """Base class for threads."""
+
```
Carreau, line 9: Misc notes – not requests for changes, It's just a long PR and I want to have notes for myself.

Should we use the opportunity to add type annotations ?

ipykernel/kernelbase.py
```diff
@@ -356,28 +361,91 @@ async def advance_eventloop():
...
+                _, msg2 = self.session.feed_identities(msg, copy=copy)
+                try:
+                    msg3 = self.session.deserialize(msg2, content=False, copy=copy)
+                    subshell_id = msg3["header"].get("subshell_id")
+
+                    # Find inproc pair socket to use to send message to correct subshell.
+                    socket = self.shell_channel_thread.manager.get_shell_channel_socket(subshell_id)
+                    assert socket is not None
+                    socket.send_multipart(msg, copy=False)
+                except Exception:
```
Carreau, line 393: Do you have any idea if we can narrow this exception, or can it be anything, or maybe too many possibilities ?

ipykernel/kernelbase.py
```diff
@@ -1274,3 +1413,6 @@ async def _at_shutdown(self):
                     ident=self._topic("shutdown"),
                 )
                 self.log.debug("%s", self._shutdown_message)
+
+    def _supports_kernel_subshells(self):
```
Carreau, line 1417:
```suggestion
    def _supports_kernel_subshells(self) -> bool:
```

Carreau: And is it worth making it a property ? (I don't really care as it's private) I guess the questions is wether shell_channel_thread may change ?

ipykernel/subshell_manager.py
```diff
@@ -0,0 +1,280 @@
...
+
+        Only used by %subshell magic so does not have to be fast/cached.
+        """
+        with self._lock_cache:
+            if thread_id == main_thread().ident:
+                return None
+            for id, subshell in self._cache.items():
+                if subshell.thread.ident == thread_id:
+                    return id
+            msg = f"Thread id '{thread_id} does not correspond to a subshell of this kernel"
```
Carreau, line 156: Unmatched ', trying to ass quotes ?  Maybe `!r` ?
```suggestion
            msg = f"Thread id {thread_id!r} does not correspond to a subshell of this kernel"
```

ipykernel/kernelbase.py
```diff
@@ -356,28 +361,91 @@ async def advance_eventloop():
     def _message_counter_default(self):
         return itertools.count()
 
-    async def shell_main(self):
+    async def shell_channel_thread_main(self):
+        """Main loop for shell channel thread.
+
+        Listen for incoming messages on kernel shell_socket.  For each message
+        received, extract the subshell_id from the message header and forward the
```
Carreau, line 368: Also vague question on subshell_id, is it arbitrary string and should we limit in lenght and to ascci, or something similar.

ipykernel/kernelbase.py
```diff
@@ -356,28 +361,91 @@ async def advance_eventloop():
...
         except BaseException as e:
             if self.shell_stop.is_set():
                 return
             raise e
 
-    async def process_shell_message(self, msg=None):
-        assert self.shell_socket is not None
+    async def process_shell_message(self, msg=None, socket=None):
+        if not socket:
+            socket = self.shell_socket
```
Carreau, line 435: I'm going to assume later will make socket mandatory ?
If so we may want to put a comment or a FutureWarning ?

ipykernel/kernelbase.py
```diff
@@ -806,14 +885,17 @@ async def connect_request(self, socket, ident, parent):
...
             "protocol_version": kernel_protocol_version,
             "implementation": self.implementation,
             "implementation_version": self.implementation_version,
             "language_info": self.language_info,
             "banner": self.banner,
             "help_links": self.help_links,
         }
+        if self._supports_kernel_subshells():
+            info["supported_features"] = ["kernel subshells"]
+        return info
```
Carreau, line 898: I need to read the enhancement proposal, but maybe always include `info["supported_features"]` even if empty list ?

ipykernel/kernelbase.py
```diff
@@ -984,6 +1066,63 @@ async def usage_request(self, socket, ident, parent):
     async def do_debug_request(self, msg):
         raise NotImplementedError
 
+    # ---------------------------------------------------------------------------
+    # Subshell control message handlers
+    # ---------------------------------------------------------------------------
+
+    async def create_subshell_request(self, socket, ident, parent):
```
Carreau, line 1073:
```suggestion
    async def create_subshell_request(self, socket, ident, parent) -> None:
```

ipykernel/kernelbase.py
```diff
@@ -984,6 +1066,63 @@ async def usage_request(self, socket, ident, parent):
...
+        reply = await other_socket.recv_json()
+
+        self.session.send(socket, "create_subshell_reply", reply, parent, ident)
+
+    async def delete_subshell_request(self, socket, ident, parent):
+        if not self.session:
+            return
+        if not self._supports_kernel_subshells():
+            self.log.error("KERNEL SUBSHELLS NOT SUPPORTED")
+            return
```
Carreau, line 1093: Should that be a hard error ?

I have the feeling that this should never happen or if it is should be handle/silence at higher level.

(Also I dont' think uppercase necessary)

ipykernel/shellchannel.py
```diff
@@ -0,0 +1,34 @@
...
+    @property
+    def manager(self):
+        # Lazy initialisation.
+        if self._manager is None:
+            self._manager = SubshellManager(self._context, self._shell_socket)
+        return self._manager
+
+    def run(self):
+        """Run the thread."""
+        self.name = SHELL_CHANNEL_THREAD_NAME
```
Carreau, line 30: See other comment, maybe `self.name =` not needed (to check), or at least change to assert ?

ipykernel/shellchannel.py
```diff
@@ -0,0 +1,34 @@
...
+        super().run()
+
+        if self._manager:
+            self._manager.close()
```
Carreau, line 34: should that be in try:finally  in case run fails ?

ipykernel/subshell.py
```diff
@@ -0,0 +1,35 @@
...
+        assert current_thread() == self
+        self._pair_socket = context.socket(zmq.PAIR)
+        self._pair_socket.connect(address)
+
+    def run(self):
+        super().run()
+
+        if self._pair_socket is not None:
+            self._pair_socket.close()
+            self._pair_socket = None
```
Carreau, line 35: try finally as well maybe ?

ipykernel/subshell_manager.py
```diff
@@ -0,0 +1,280 @@
...
+    Sending reply messages via the shell_socket is wrapped by another lock to protect
+    against multiple subshells attempting to send at the same time.
+    """
+
+    def __init__(self, context: zmq.asyncio.Context, shell_socket: zmq.asyncio.Socket):
+        assert current_thread() == main_thread()
+
+        self._context: zmq.asyncio.Context = context
+        self._shell_socket = shell_socket
+        self._cache: dict[str, Subshell] = {}
```
Carreau, line 45: Self curiosity:

check the difference between attribute typing in `__init__` and at class level.

Is mypy smart enough, or does it say `attribute may be unset` if type set in `__init__` ??

ipykernel/subshell_manager.py
```diff
@@ -0,0 +1,280 @@
...
+        self._parent_shell_channel_socket = self._create_inproc_pair_socket(None, True)
+        self._parent_other_socket = self._create_inproc_pair_socket(None, False)
+
+        # anyio memory object stream for async queue-like communication between tasks.
+        # Used by _create_subshell to tell listen_from_subshells to spawn a new task.
+        self._send_stream, self._receive_stream = create_memory_object_stream[str]()
+
+    def close(self) -> None:
+        """Stop all subshells and close all resources."""
+        assert current_thread().name == "Shell channel"
```
Carreau, line 63:
```suggestion
        assert current_thread().name == SHELL_CHANNEL_THREAD_NAME
```

Maybe ? May need an import, and may trigger circular import, so I understand if not.

ipykernel/subshell_manager.py
```diff
@@ -0,0 +1,280 @@
...
+        thread.start()
+
+        return subshell_id
+
+    def _delete_subshell(self, subshell_id: str) -> None:
+        """Delete subshell identified by subshell_id.
+
+        Raises key error if subshell_id not in cache.
+        """
+        assert current_thread().name == "Shell channel"
```
Carreau, line 199: same:
```suggestion
        assert current_thread().name == SHELL_CHANNEL_THREAD_NAME
```

??

ipykernel/subshell_manager.py
```diff
@@ -0,0 +1,280 @@
...
+
+            if type == "create":
+                reply["subshell_id"] = await self._create_subshell(subshell_task)
+            elif type == "delete":
+                subshell_id = request["subshell_id"]
+                self._delete_subshell(subshell_id)
+            elif type == "list":
+                reply["subshell_id"] = self.list_subshell()
+            else:
+                msg = f"Unrecognised message type {type}"
```
Carreau, line 261:
```suggestion
                msg = f"Unrecognised message type {type!r}"
```

Carreau (commented):
Thanks for your patience, I'm going to try reviewing this and pushing it forward. 

I'm done a partial read – not in depth, and added a few comments – which are not mandatory request for changes but thoughts to myself, for when I'll come back to it.

I have to run some errands, so I'm going to post this as is for now but will come back to it later.

ipykernel/thread.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, name, **kwargs):
+        """Initialize the thread."""
+        Thread.__init__(self, name=name, **kwargs)
```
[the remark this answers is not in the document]
davidbrochart: If using `super()`, don't pass `self`.

ipykernel/thread.py
ianthomas23: I added type annotations where it helped, such as some of the new `SubshellManager`. For `Thread` classes I followed the existing code (e.g. https://github.com/ipython/ipykernel/blob/main/ipykernel/control.py) and aimed for minimal changes, so no type annotations.

I am happy to add annotations for the code I've added and/or changed.

ipykernel/thread.py
```diff
@@ -0,0 +1,40 @@
...
+        self.pydev_do_not_trace = True
+        self.is_pydev_daemon_thread = True
+        self.__stop = Event()
+        self._tasks = []
+        self._task_args = []
```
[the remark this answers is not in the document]
ianthomas23: Yes, that is a good idea.

ipykernel/thread.py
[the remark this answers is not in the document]
ianthomas23: Absolutely. Note I was following the existing code https://github.com/ipython/ipykernel/blob/0fa543936d3f9ee69f6b9f8bd5a316692fde9aaa/ipykernel/control.py#L14

ipykernel/kernelbase.py
ianthomas23: I don't know. This is the same usage as in the existing code at
https://github.com/ipython/ipykernel/blob/0fa543936d3f9ee69f6b9f8bd5a316692fde9aaa/ipykernel/debugger.py#L442-L445
so I am assuming that a failed `deserialize` could raise any `Exception`, and I've put the forwarding of the message in the same `try` block as if it is not a valid message it should not be forwarded.

ipykernel/kernelbase.py
ianthomas23: I think making it a property is sensible. I cannot imagine any circumstances in which `shell_channel_thread` may change, but you never know...

ipykernel/subshell_manager.py
ianthomas23: This was probably just a typo on my behalf!

Carreau:
I'm +1 for this, a couple of stylistic notes – if you agree I'm happy to push some changes updates but don't want to do so without your approval as I know it can get unwieldy to have another maintainer push changes on your branch on big PRs.

I think we can merge this as "experimental" and then work on the related issues when using this features on various projects (IPython history etc...).

ipykernel/kernelbase.py
ianthomas23: The [JEP](https://github.com/jupyter/enhancement-proposals/blob/master/kernel-subshells/kernel-subshells.md) says that a `subshell_id` is a string, is unique within a kernel, and it is generated for you. I implemented this as `uuid.uuid4()`:
https://github.com/ianthomas23/ipykernel/blob/a708e0ee8c31a5e91175b484e76f8af48be93235/ipykernel/subshell_manager.py#L175
As it cannot be an arbitrary string I haven't put in any explicit validation as it is either recognised (corresponding to an actual subshell) or it isn't.

ipykernel/kernelbase.py
ianthomas23: An input argument of `socket=None` is correct if subshells are not supported in which case the behaviour of this function is unchanged compared to the `main` branch, which is (probably) important for downstream libraries. There are a few situations where this is the case, I will add extra comments to explain. Also, here I will move these lines to the `_supports_kernel_subshells` block below to assert that `socket` can only be `None` if subshells are not supported.

ipykernel/kernelbase.py
ianthomas23: The [supported features JEP](https://github.com/jupyter/enhancement-proposals/blob/master/92-jupyter-optional-features/jupyter-optional-features.md) isn't explicit about this, so I assume we can choose to add it even if it is an empty list.

ipykernel/kernelbase.py
ianthomas23: This is based on existing control message handling, if we receive a message that we don't recognise we log an error message and then carry on without considering it an error situation:
https://github.com/ipython/ipykernel/blob/0fa543936d3f9ee69f6b9f8bd5a316692fde9aaa/ipykernel/kernelbase.py#L277
and:
https://github.com/ipython/ipykernel/blob/0fa543936d3f9ee69f6b9f8bd5a316692fde9aaa/ipykernel/kernelbase.py#L291

But I suppose here the message type and contents are valid, but the kernel is not capable of doing what is requested. We could return a reply message with `{'status': 'error'}`? What to do here is not covered by the JEP.

Uppercase is certainly not necessary!

ipykernel/shellchannel.py
ianthomas23: Yes, that is a good idea.

tests/utils.py
```diff
@@ -2,6 +2,7 @@
 
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
+from __future__ import annotations
```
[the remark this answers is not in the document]
ianthomas23: In general I have tried to follow the existing code as an example of what is acceptable and this line is used in 5 other places. I think I will just remove it and the `list[str]` annotation as there are no other type annotations in this file.

ianthomas23:
@Carreau Thanks for the review and for the offer of pushing changes. I'd rather make the changes myself, and I'll try to group them sensibly to avoid to much CI churn.

ipykernel/thread.py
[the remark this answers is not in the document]
Carreau: > If using `super()`, don't pass `self`.

Yep, sorry, updated.

And indeed, I know some of the code is just moved from somewhere else, just taking the opportunity to clean it up :-)

ipykernel/kernelbase.py
Carreau: Sounds good. Let's leave `Exception` then.

ipykernel/kernelbase.py
Carreau: Ok, I think that's fine.

ipykernel/kernelbase.py
Carreau: I believe it is ok if we want to think about making `socket=` mandatory in a few years. 

I tend to not like API where things are sometime magic and sometime not.

ipykernel/kernelbase.py
Carreau: I would prefer starting with a hard failure, worse case in a minor revision we ignore. The main reason is that raising where it was not previously have more chance to break downsrtream users and is easier to track than just ignoring anything. Returning status:error woudl be ok-ish I guess.

tests/utils.py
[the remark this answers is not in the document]
Carreau: No, it's fine you can leave it.  It will eventually be correct.

ipykernel/subshell_manager.py
ianthomas23: I have moved the definitions of `CONTROL_THREAD_NAME` and `SHELL_CHANNEL_THREAD_NAME` to the `BaseThread` file as this avoids the circular imports. It makes sense anyway, thread names are ideally unique so if they are defined together you can see what the other ones are.

davidbrochart:
It seems that you use `zmq.PAIR` sockets for inter-thread communication (sub-shell thread <---> main-shell thread, sub-shell thread <--> control thread), are those real sockets? If so, sockets are a scarce resource, and AnyIO has nice facilities for inter-thread communication, that could be used instead?
I see that you use AnyIO memory object streams in the sub-shell manager, could you explain your use of them, in particular if they are used for inter-thread communication?

ipykernel/thread.py
ianthomas23: More type annotations in 5715dd09. I've held back from using `Callable` as that eventually caused me more trouble that I was prepared to deal with at this stage.

ipykernel/thread.py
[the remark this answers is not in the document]
ianthomas23: Done in 2a91bcd as `self._tasks_and_args`

ipykernel/thread.py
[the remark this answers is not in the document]
ianthomas23: Done in 4bcd59e.

ipykernel/kernelbase.py
ianthomas23: Left as it is.

ipykernel/kernelbase.py
ianthomas23: Made it a property in a1fb1de.

ipykernel/subshell_manager.py
ianthomas23: Fixed ind17c77f.

ipykernel/kernelbase.py
ianthomas23: Always return `supported_features` in 924015a.

ipykernel/kernelbase.py
ianthomas23: Done in 5715dd0.

ipykernel/shellchannel.py
ianthomas23: `try .. finally` added in b644f9e7.

ipykernel/subshell.py
ianthomas23: `try .. finally` added in b644f9e7.

ipykernel/subshell_manager.py
ianthomas23: Fixed in 9109cc8.

ipykernel/subshell_manager.py
ianthomas23: Fixed in 9109cc8.

ipykernel/subshell_manager.py
ianthomas23: Fixed in d17c77f4.

ipykernel/kernelbase.py
```diff
@@ -356,28 +361,91 @@ async def advance_eventloop():
...
+                    # Find inproc pair socket to use to send message to correct subshell.
+                    socket = self.shell_channel_thread.manager.get_shell_channel_socket(subshell_id)
+                    assert socket is not None
+                    socket.send_multipart(msg, copy=False)
+                except Exception:
+                    self.log.error("Invalid message", exc_info=True)  # noqa: G201
+        except BaseException as e:
+            if self.shell_stop.is_set():
+                return
+            raise e
```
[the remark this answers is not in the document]
ianthomas23: Fixed in d17c77f4, along with a few other uses of exactly the same.

ipykernel/kernelbase.py
```diff
@@ -984,6 +1066,63 @@ async def usage_request(self, socket, ident, parent):
...
+        if not self._supports_kernel_subshells():
+            self.log.error("KERNEL SUBSHELLS NOT SUPPORTED")
+            return
+
+        try:
+            content = parent["content"]
+            subshell_id = content["subshell_id"]
+        except Exception:
+            self.log.error("Got bad msg: ")
+            self.log.error("%s", parent)
```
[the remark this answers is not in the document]
ianthomas23: Fixed in d17c77f4, with one extra use elsewhere in the same file.

ipykernel/thread.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def add_task(self, task, *args):
+        # May only add tasks before the thread is started.
+        self._tasks.append(task)
+        self._task_args.append(args)
+
+    def run(self):
+        """Run the thread."""
+        run(self._main)
```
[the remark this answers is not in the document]
ianthomas23: Fixed in b644f9e.

ipykernel/thread.py
```diff
@@ -0,0 +1,40 @@
...
+
+    async def _main(self):
```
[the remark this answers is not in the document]
ianthomas23: Fixed in b644f9.

ipykernel/thread.py
```diff
@@ -0,0 +1,40 @@
...
+        async with create_task_group() as tg:
+            for task, args in zip(self._tasks, self._task_args):
+                tg.start_soon(task, *args)
+            await to_thread.run_sync(self.__stop.wait)
+            tg.cancel_scope.cancel()
+
+    def stop(self):
```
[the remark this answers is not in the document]
ianthomas23: Fixed in 5715dd0.

ipykernel/zmqshell.py
```diff
@@ -439,6 +439,40 @@ def autosave(self, arg_s):
...
+        app = IPKernelApp.instance()
+        kernel = app.kernel
+
+        if not (
+            hasattr(kernel, "_supports_kernel_subshells") and kernel._supports_kernel_subshells()
+        ):
+            print("Kernel does not support subshells")
+            return
+
+        import threading
```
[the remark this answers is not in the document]
ianthomas23: Fixed in d17c77f4.

ipykernel/zmqshell.py
```diff
@@ -439,6 +439,40 @@ def autosave(self, arg_s):
         else:
             print("Autosave disabled")
 
+    @line_magic
+    def subshell(self, arg_s):
```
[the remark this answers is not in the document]
ianthomas23: Fixed in d17c77f4.

tests/utils.py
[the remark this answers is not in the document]
ianthomas23: Kept as it is.

ianthomas23:
> It seems that you use `zmq.PAIR` sockets for inter-thread communication (sub-shell thread <---> main-shell thread, sub-shell thread <--> control thread), are those real sockets?

A ZMQ inproc pair is an area of memory shared between two threads and wrapped to look like ZMQ sockets.

> I see that you use AnyIO memory object streams in the sub-shell manager, could you explain your use of them, in particular if they are used for inter-thread communication?

Here is where they are created in the source code which includes some explanation:
```python
# anyio memory object stream for async queue-like communication between tasks.
# Used by _create_subshell to tell listen_from_subshells to spawn a new task.
self._send_stream, self._receive_stream = create_memory_object_stream[str]()
```
The communication is between two different async tasks running in the same thread.

ipykernel/kernelbase.py
ianthomas23: To clarify, by 'hard failure' do you mean raise an exception?

ipykernel/subshell_manager.py
```diff
@@ -0,0 +1,283 @@
...
+                self._delete_subshell(subshell_id)
+            elif type == "list":
+                reply["subshell_id"] = self.list_subshell()
+            else:
+                msg = f"Unrecognised message type {type!r}"
+                raise RuntimeError(msg)
+        except BaseException as err:
+            reply = {
+                "status": "error",
+                "evalue": str(err),
```
Carreau, line 269: not for this PR:
 - Can this contain information that should not be send to the frontend ?
 - I don't believe so as the user is anyway the one that runs the kernel, but we had case where traceback would contain sensitive values. In other apps webapp this would be a likely issue (like revealing paths and filename of a server), but I don't think it's an issue here.

Carreau (approved):
Some notes for myself and history, but all good on my side. Let's get that in and iterate if necessary.

Carreau:
I haven't merged something on this repo in quite some time, I think I would prefer a rebase-and-merge as you took time to craft the commits individually, but I think the policy here is squash-merged.

Any objections to squash-merge ?

ianthomas23:
> Any objections to squash-merge ?

A squash-merge is fine by me.

davidbrochart (approved):
Thanks @ianthomas23 !

PR #1253, opened 2024-06-28
Explicitly close memory object streams

### LINKED ISSUE #1252: Test failures with anyio 4.4.0
We are seeing test failures in CI when using anyio 4.4.0 which were OK with 4.3.0. Bisecting the anyio commits shows that the change in behaviour occurs at commit `d3f38a7c` which is PR agronholm/anyio#715. This emits a `ResourceWarning` for any `MemoryObjectReceiveStream` and `MemoryObjectSendStream` that are garbage collected before they are closed. The solution is to explicitly close the streams, I will submit a PR for this shortly.

ianthomas23:
Fixes #1252.

When using anyio `MemoryObjectReceiveStream` and `MemoryObjectSendStream` we must explicitly close them when we have finished with them. This wasn't necessary with anyio 4.3.0 but is with 4.4.0 as otherwise warnings/errors are issued which break our CI.

This is only actually an issue in CI as our only explicit use of memory object streams is in `TestSession`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -63,6 +63,12 @@ class TestSession(Session):
             send_stream, receive_stream = create_memory_object_stream(max_buffer_size=inf)
             self._streams[socket] = {"send": send_stream, "receive": receive_stream}
 
+    def close(self):
+        for streams in self._streams.values():
+            for stream in streams.values():
+                stream.close()
+        self._streams.clear()
+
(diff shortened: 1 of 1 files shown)


davidbrochart (approved):

ianthomas23:
This fixes all the local CI runs, except the docs which is some other problem. I am self-merging.

Issue #1236, opened 2024-05-07
Restore to previous state with copy-on-write


zmbc:
Apologies if this isn't the right place for this kind of feature request.

As a Jupyter user, I would find it _incredibly_ helpful if there was an "undo" for kernels. Frequently, I run something, only to realize it was the wrong thing and I want to go back one cell. If the bad cell had any destructive effect, this means I need to restart my kernel and run all the cells above _again_. If those cells took 3 hours to run, I need to wait 3 hours again. Even if I don't think the bad cell had any destructive effect, if I fix up the notebook mid-stream others (and myself) won't be 100% sure that the notebook would work with a simple "restart and run all."

I wish there were a command like "Kernel -> Save Checkpoint" and then the ability to restore back to checkpoints. In Jupyter, if this worked in tandem with the checkpointing of the notebook file itself, it could create a seamless experience: I create a checkpoint, make a mistake, restore both kernel and notebook to the checkpoint, fix the mistake and keep running, and I end up with _exactly_ the notebook I would have had if I hadn't made the mistake in the first place. (On the Jupyter side I'd also like an option that automatically checkpointed after each cell, and one that forks off into a console instead of checkpointing the notebook file.)

Technically, it seems like this could be achieved through `fork`'s copy-on-write mechanism. For many notebooks that don't have much volatile state that is rewritten multiple times, even checkpoints after every cell could have near-zero memory overhead -- they would just be forks of the kernel process at that point, using the same pages in memory as the active kernel process unless they are subsequently modified.

davidbrochart:
Thanks @zmbc, that's an interesting idea.
Would you like to experiment with this idea through a PR?

maartenbreddels:
Great idea, I worked a bit on this: https://github.com/ipython/ipykernel/pull/410 maybe it is helpful

zmbc:
@davidbrochart would you be able to provide review/feedback on my PR #1261?

PR #410, opened 2019-05-28
WIP: Feature: fork kernel

maartenbreddels:
# Issue
For dashboarding, it is useful to have a kernel + the final state ready instantly. If for instance, it takes ~5 minutes to execute a notebook, it does not lead to good user experience.

# Solution
Execute a notebook, and after it is done, consider it a template, and give each user a `fork()` of this template process.

# Implementation
The kernel gets a new message, which is simply called `fork`.  Asyncio/tornado-ioloop with fork is a no-go, it is basically not supported. To avoid this, we stop the current ioloop, and use the ioloop object to communicate the fork request up the callstack. The kernelapp now checks if a fork was requested and it  will fork:
 * The parent resumes the ioloop
 * The child should clean up the ioloop and start listening on new ports:

# Usage (although it does not work yet)

It also needs this PR: https://github.com/jupyter/jupyter_client/pull/441

Shell1:
```bash
$ python -m ipykernel_launcher -f conn.json --debug
```
Shell2:
```bash
$ python fork_kernel.py ./conn.json # this will print the result of the fork
DEBUG:traitlets:Loading connection file ./conn.json
DEBUG:traitlets:connecting shell channel to tcp://127.0.0.1:56573
DEBUG:traitlets:Connecting to: tcp://127.0.0.1:56573
DEBUG:traitlets:connecting iopub channel to tcp://127.0.0.1:56574
DEBUG:traitlets:Connecting to: tcp://127.0.0.1:56574
DEBUG:traitlets:connecting stdin channel to tcp://127.0.0.1:56575
DEBUG:traitlets:Connecting to: tcp://127.0.0.1:56575
DEBUG:traitlets:connecting heartbeat channel to tcp://127.0.0.1:56577
[...]

PR #1261, opened 2024-08-04
Feature: Fork kernel

zmbc:
Building on work by @maartenbreddels and @edisongustavo in #410, this is my first pass at trying to add the feature I requested in #1236.

The tests pass 🎉 However, this involved a fair amount of trial and error and I still don't fully understand why it works the way it does -- see comments inline. I would really appreciate some ideas here.

Requires a corresponding change to `jupyter_client`: https://github.com/jupyter/jupyter_client/pull/1031.

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -117,6 +117,23 @@ class IPythonKernel(KernelBase):
                 self.debug_just_my_code,
             )
 
+        self.init_shell()
+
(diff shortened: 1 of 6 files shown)


bluss:
`fork()` unfortunately interacts very badly with threading (OS threads). Is it possible that the user has kicked off some running threads?

See also discussion https://discuss.python.org/t/switching-default-multiprocessing-context-to-spawn-on-posix-as-well/21868

PR #1262, opened 2024-08-06
Ignore warning from trio

ianthomas23:
CI is failing with
```
ERROR: while parsing the following warning configuration:

  ignore:trio.MultiError is deprecated since Trio 0.22.0:trio.TrioDeprecationWarning

This error occurred:

DeprecationWarning: The `hash` argument is deprecated in favor of `unsafe_hash` and will be removed in or after August 2025.
```
using `attrs` 24.1.0 and `trio` 0.26.1. Related issue in `trio` is python-trio/trio#3053. This PR ignores the new warning.


ianthomas23:
This fixes the local tests, the downstream tests were already broken and remain so. Merging to clear CI.

PR #1237, opened 2024-05-09
Fix use of "%matplotlib osx"

ianthomas23:
Fixes #1124 which is that the use of `%matplotlib osx` crashes `ipykernel` and downstream libraries whereas `ipython` is fine.

The fix is to specify the `argtypes` of each `objc_msgSend` call just before the call. This is consistent with how `ipython` successfully handles it. I have tested this manually using `spyder` and `jupyter console`, `qtconsole`, `lab` and `nbclassic`: using `ipykernel 6.29.4` (the latest release) they all crash, using this branch they all work as expected.

This PR is opened against the `6.x` branch rather than `main` as issue #1235 means it is not possible to test any `matplotlib` GUI loops on `main`. It would be good to get this into a release (6.29.5?) as soon as possible and not have to wait for #1235 and the other AnyIO fallout issues to be fixed beforehand. This PR will therefore need to be forward ported to `main`.

There is no explicit test for this as you need to actually render a macos window to trigger the failure, and this is not an easy thing to do on headless CI. Essentially this functionality sits in the grey area between `matplotlib` and `ipython/ipykernel/jupyter` rather than in any individual project.

Issue #1235, opened 2024-05-03
AnyIO commit breaks Matplotlib display in separate windows


ianthomas23:
With the commit before the AnyIO PR on the `main` branch (830829fd) and using `ipython 8.23.0` and `matplotlib 3.8.4`, the displaying of Matplotlib plots in separate windows (e.g. with `qt` backend) using jupyter works fine. With the AnyIO commit (772dfb8a) onwards the plot windows are no longer displayed.

Code to reproduce (I've been using `jupyter qtconsole` but `jupyter lab`, etc, all show the same):
```python
%matplotlib qt
import matplotlib.pyplot as plt
fig, ax = plt.subplots()
ax.plot([1, 3, 2])
```

In `ipython` the same example works fine.

davidbrochart:
Probably because of external event loop integration.

ianthomas23:
`enter_eventloop` is only ever called if `kernel.eventloop` is set at startup:

https://github.com/ipython/ipykernel/blob/a7d66ae2197e0d7471ba160542cf5ff7713084b5/ipykernel/kernelapp.py#L732-L736

If the `eventloop` is set later, e.g. via `%matplotlib qt`, it is never called.

Plus there are 7 uses of `kernel.shell_stream` in `eventloops.py` and the `shell_stream` was replaced with `shell_port` in the AnyIO changes. So evidently this code isn't covered by any `ipykernel` tests.

minrk:
I've been looking into this today, and it is definitely the case that the way the event loop integration expects to return control has assumptions about socket use that are not satisfied by the changes in how sockets are used.

In particular, the event loop integration pattern is that the event loops block and hold control until there is a message waiting to be received on the shell socket, at which point control is returned to the kernel. But with the new async architecture, there could be any number of events waiting to process, including the coroutines involved in actually handling a message.

It is not hard to test this, because this test:

```python
@pytest.mark.parametrize("gui", [
    "qt",
    "osx",
])
def test_gui_responsive(kernel_client, gui):
    if gui == 'osx':
        if sys.platform == 'darwin':
            pytest.importorskip("matplotlib")
        else:
            pytest.skip("osx test only on mac")
    if gui == 'qt' and len(qt_guis_avail) == 0:
        pytest.skip("No qt backend available")
      
    kc = kernel_client

    # enable gui
    msg_id, reply = execute(f"%gui {gui}", kc=kc)
    assert reply['status'] == 'ok'
    flush_channels(kc)

[...]

PR #1264, opened 2024-08-06
Specify argtypes when using macos msg

ianthomas23:
This is a forward port of PR #1237 from the `6.x` branch to `main`. I think it was supposed to happen automatically as part of the semi-automated release process but there were a number of problems with that.

Note that this isn't actually usable until #1235 is fixed.


ianthomas23:
Self-merging as this is a forward port.

PR #1265, opened 2024-08-09
Fix eventloop integration with anyio

### LINKED ISSUE #1235: AnyIO commit breaks Matplotlib display in separate windows

ianthomas23:
Fixes #1235.

This fixes GUI event loop integration so that it works following the recent switch to using anyio, so that Matplotlib output correctly displays in separate Windows and both the kernel and plot respond to subsequent input. Such functionality has never been explicitly tested in ipykernel, so I have tested it manually on Linux, macOS and Windows with various combinations of `qt`, `tk` and `osx` Matplotlib backends and Jupyter `lab`, `console`, `qtconsole` and `spyder`.

Here's a screenshot:

Screenshot 2024-08-09 at 15 36 27

The plot can be panned and zoomed for example, which cannot be shown in such a screenshot.

Summary of changes:
1. Switch from using `shell_stream` to `shell_socket` as the former is no longer used.
2. To check if there are ZMQ messages pending switch from using `shell_stream.flush(limit=1)` to `shell_socket.get(zmq.EVENTS) & zmq.POLLIN) > 0`.
3. If an eventloop is set (via `enable_gui`) after the kernel has started, a new `anyio.Event` is used to trigger calling `enter_eventloop`. This allows the synchronous `enable_gui` to trigger the asynchronous `enter_eventloop`.
4. The same event is used to close the waiting task if the kernel is stopped without an eventloop being set.
5. The nested calls of `advance_eventloop` and `schedule_next` are replaced with a simple async loop.

These changes are intentionally minimal to reduce the danger of breaking downstream code.

I have disabled `test_trio_loop` as the other `trio`-based tests are disabled and we don't appear to support it yet, although as it is one of the async backends supported by `anyio` this shouldn't be too onerous. That is separate work to this though. I also noted that `TrioRunner` still exists but is not used any where.

### DIFF
--- a/ipykernel/eventloops.py
+++ b/ipykernel/eventloops.py
@@ -10,6 +10,7 @@ from functools import partial
 
 import zmq
 from packaging.version import Version as V
+from sniffio import AsyncLibraryNotFoundError
 from traitlets.config.application import Application
 
 
@@ -93,11 +94,11 @@ def _notify_stream_qt(kernel):
         # due to our consuming of the edge-triggered FD
         # flush returns the number of events consumed.
         # if there were any, wake it up
-        if kernel.shell_stream.flush(limit=1):
+        if (kernel.shell_socket.get(zmq.EVENTS) & zmq.POLLIN) > 0:
             exit_loop()
 
     if not hasattr(kernel, "_qt_notifier"):
-        fd = kernel.shell_stream.getsockopt(zmq.FD)
+        fd = kernel.shell_socket.getsockopt(zmq.FD)
         kernel._qt_notifier = QtCore.QSocketNotifier(
             fd, enum_helper("QtCore.QSocketNotifier.Type").Read, kernel.app.qt_event_loop
         )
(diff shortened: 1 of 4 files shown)


ianthomas23:
Minimum version tests are failing because in `anyio` prior to 4.2.0 an `Event` can only be instantiated from within the async loop (agronholm/anyio#651).

ianthomas23:
I have bumped the minimum anyio version from 4.0.0 (August 2023) to 4.2.0 (December 2023). Now the only test failures are the same as those occurring on the `main` branch, so this is now ready for review.

blink1073 (approved):
LGTM!

ianthomas23:
Thanks @blink1073

PR #1270, opened 2024-10-03
Run Mypy by default.

Carreau:
Otherwise the types annotations are less usefull, even with ruff.

Ignore/fix types errors.

Also the mypy config was way to lax, ignoring errors project wide intead of per-file. 
This make it much stricter and only ignore error on each file, so this types of errors are still confined, and it is easy-ish, to make it a bit more stricter progressively


Carreau:
I'm still also super-confused as running mypy directly and running via hatch/precommit gives different results

PR #1271, opened 2024-10-05
Detect parent change in more cases on unix

### LINKED ISSUE #517: PPID is not necessarily 1 when parent frontend is killed
Hi, I have raised this issue separately on [jupyter-console](https://github.com/jupyter/jupyter_console/issues/221)

The problem I noticed is that when I directly close the frontend program that called "jupyter-console" (e.g. my terminal), it leaves the "ipykernel_launcher" running in the background. What I expected to happen is that the kernel session created by "jupyter-console" will be killed whenever its host frontend program is killed. I think I have some clues as to the reason why.

First to provide some of the relevant observations:
1. On a remote machine without interactive login session, killing the parent frontend (e.g. a Tmux session or terminal) hosting the Jupyter console will lead to the expected killing of ipykernel on the background. 
2. On another machine running Manjaro Gnome, killing the parent frontend (e.g. a Tmux session or terminal) will not kill the ipykernel subprocess. I can see that it's still dangling in the background.

Chasing down the rabbit hole, I found additional observations:
1. "ipykernel/parentpoller.py" : is responsible for continuous monitoring of parent process that spawned ipykernel. "Line 36" makes the assumption that if the parent of orphaned process is killed, its ID gets set to "1", in which case the kernel will exit itself.
parentpoller_relevant_code
2. After adding in a line to log the parent process ID, I found that on the Manjaro Gnome machine, the initial parent process ID transfers from the "jupyer-console":
initial_ppid
after killing the initial parent process to:
post_parent_killed_ppid

ppid

A few additional [posts](https://unix.stackexchange.com/questions/149319/new-parent-process-when-the-parent-process-dies/177361#177361) suggest that the parent ID of orphaned processes may not be necessarily set to 1 if there are ancestor processes. Since [Gnome 3.34](https://blogs.gnome.org/benzea/2019/10/01/gnome-3-34-is-now-managed-using-systemd/) (and maybe a few more other distributions), there is a shift towards handing off more processes to systemd. Thus when orphaned child arises, it does not get handed off to "init" with a PID of 1. It seems ipykernel may require a special handling to prevent undead backends that consume system resources. A suggestion would be to kill the ipykernel whenever PPID == 1 or PPID == "that of systemd".

bluss:
New behaviour: poll getppid and exit if this value is no longer equal to the original parent pid.

For backwards compatibility: keep old behaviour when parent_pid is 0 or not given.

Parentpoller only watches for parent change to pid 1 (init), but in modern linux environments processes are usually reparented to the systemd user session process. In this situation, ipykernel never noticies that its parent has exited.

jupyter-client already sets JPY_PARENT_PID with the pid of the parent process. We could trust this environment variable directly, but in this change, it is only used if it is equal to the actual parent id when the poller starts. (A side effect: if the parent dies quickly enough, ipykernel does not notice.)

Fixes #517

### DIFF
--- a/ipykernel/parentpoller.py
+++ b/ipykernel/parentpoller.py
@@ -22,9 +22,17 @@ class ParentPollerUnix(Thread):
     when the parent process no longer exists.
     """
 
-    def __init__(self):
-        """Initialize the poller."""
+    def __init__(self, parent_pid=0):
+        """Initialize the poller.
+
+        Parameters
+        ----------
+        parent_handle : int, optional
+            If provided, the program will terminate immediately when
+            process parent is no longer this original parent.
+        """
         super().__init__()
+        self.parent_pid = parent_pid
         self.daemon = True
 
     def run(self):
@@ -32,9 +40,23 @@ class ParentPollerUnix(Thread):
         # We cannot use os.waitpid because it works only for child processes.
         from errno import EINTR
 
+        # before start, check if the passed-in parent pid is valid
+        original_ppid = os.getppid()
+        if original_ppid != self.parent_pid:
+            self.parent_pid = 0
+
+        get_logger().debug(
+            "%s: poll for parent change with original parent pid=%d",
+            type(self).__name__,
+            self.parent_pid,
+        )
+
         while True:
             try:
-                if os.getppid() == 1:
+                ppid = os.getppid()
+                parent_is_init = not self.parent_pid and ppid == 1
+                parent_has_changed = self.parent_pid and ppid != self.parent_pid
+                if parent_is_init or parent_has_changed:
                     get_logger().warning("Parent appears to have exited, shutting down.")
                     os._exit(1)
                 time.sleep(1.0)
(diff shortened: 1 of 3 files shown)


bluss:
First PR, tell me if I did something wrong

bluss:
The ci failures are similar to failures on other PRs here, so it's possible - not certain - that they are unrelated errors.

ipykernel/parentpoller.py
```diff
@@ -22,19 +22,41 @@ class ParentPollerUnix(Thread):
...
         super().__init__()
+        self.parent_pid = parent_pid
         self.daemon = True
 
     def run(self):
         """Run the poller."""
         # We cannot use os.waitpid because it works only for child processes.
         from errno import EINTR
 
+        # before start, check if the passed-in parent pid is valid
```
bluss, line 43: Maybe this whole thing is overly complicated? We could also elect to just trust JPY_PARENT_PID directly - the only problem is if it's a stale environment variable from somewhere.

Carreau:
That looks reasonable to me.

Carreau:
Downstream failures seem unrelated. Merging.

ipykernel/parentpoller.py
bluss: Thanks for merging!

My updated thoughts on this comment here is that it's good that we check if JPY_PARENT_PID is the actual parent. It's actually possible (is it common?) to have situations like  `papermill -> bash -> ipykernel` or with similar intermediate processes, where JPY_PARENT_PID will point to papermill but the ipykernel ppid will be bash.

PR #1272, opened 2024-10-15
Misc cleanup and types updates

Carreau:


ipykernel/kernelapp.py
```diff
@@ -142,9 +143,9 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp, ConnectionFileMix
     debug_shell_socket = Any()
     stdin_socket = Any()
     iopub_socket = Any()
-    iopub_thread = Any()
-    control_thread = Any()
-    shell_channel_thread = Any()
+    iopub_thread: Thread
+    control_thread: Thread
+    shell_channel_thread: Thread
```
davidbrochart, line 148: [BaseThread](https://github.com/ipython/ipykernel/blob/3089438d84447bf859556b89f528ef59332803c5/ipykernel/thread.py#L11)?

ipykernel/kernelbase.py
```diff
@@ -217,6 +217,9 @@ def _parent_header(self):
         "shutdown_request",
         "is_complete_request",
         "interrupt_request",
+        # I have the impression that there is amix/match debug_request and do_debug_request
+        # former was from ipyparallel but is marked as deprecated, but now call do_debug_request
+        # "do_debug_request"
```
davidbrochart, line 222: Maybe not leave this in the codebase?

ipykernel/kernelbase.py
```diff
@@ -258,6 +261,17 @@ def __init__(self, **kwargs):
             self.do_execute, ["cell_meta", "cell_id"]
         )
 
+        if not inspect.iscoroutinefunction(self.do_debug_request):
+            # warning at init time, as we want to warn early enough, even if the
+            # function is not called
+            warnings.warn(
+                "`do_debug_request` will be required to be a coroutine "
+                "functions in the future. coroutine functions have ben supported "
```
davidbrochart, line 269:
```suggestion
                "function in the future. Coroutine functions have been supported "
```

ipykernel/kernelbase.py
```diff
@@ -1007,13 +1021,35 @@ def do_is_complete(self, code):
...
     async def debug_request(self, socket, ident, parent):
-        """Handle a debug request."""
+        """Handle a debug request.
+
+        Seem Deprecated
+        """
+        # If this is incorrect, remove `debug_request` from the deprecated message types
+        # at the beginning of this class
+        warnings.warn(
+            "debug_request is deprecated in kernel_base since ipykernel 6.10"
```
davidbrochart, line 1031:
```suggestion
            "`debug_request` is deprecated in `kernelbase.py` since ipykernel 6.10"
```

ipykernel/kernelbase.py
```diff
@@ -1007,13 +1021,35 @@ def do_is_complete(self, code):
...
+            # repeat the warning at run
+            reply_content = self.do_debug_request(content)
+            warnings.warn(
+                "`do_debug_request` will be required to be a coroutine "
+                "functions in the future. coroutine functions have ben supported "
+                "since ipykernel 6.0 (2021)",
+                DeprecationWarning,
+                stacklevel=1,
+            )
+            if inspect.isawaitable(reply_content):
```
davidbrochart, line 1049: Why not throwing the warning if `reply_content` is not awaitable?

ipykernel/kernelapp.py
Carreau: I'm not sure if it's overridden in subclasses somewhere, but sure I can try so if mypy ever complains.

ipykernel/kernelbase.py
Carreau: Because testing for awaitable is/was the incorrect things to do.

I believe at some point sync functions were returning Futures, which are awaitable. 

We do not want the return a `do_debug_request` to be an awaitble; we want `do_debug_request` to be a coroutine function.

ipykernel/kernelbase.py
Carreau: Well, it's an actual question I have, and I don't think it does belong in an issue. 
I'm  not too sure where to leave it, and I'm certain if I remove it, I will spend  again an hour to figure that out next time I try to clean this up.

minrk: `debug_request` is not related to ipyparallel. ipyparallel doesn't use or support the debugging messages.

ipykernel/kernelbase.py
```diff
@@ -1132,7 +1168,12 @@ async def list_subshell_request(self, socket, ident, parent) -> None:
 
     async def apply_request(self, socket, ident, parent):  # pragma: no cover
         """Handle an apply request."""
-        self.log.warning("apply_request is deprecated in kernel_base, moving to ipyparallel.")
+        warnings.warn(
```
minrk, line 1171: Since the async changes totally break all IPython Parallel versions that didn't have these methods, it probably makes sense to remove the IPython Parallel methods in ipykernel 7, as keeping them while breaking them at the same time doesn't make a lot of sense.

Methods/attributes to remove:

- `aborted`
- `_send_abort_reply`
- `apply_request`
- `abort_request`
- `clear_request`
- `do_clear`

I can work on that, if you want, since I'm working on the ipyparallel updates now anyway

ipykernel/kernelbase.py
```diff
@@ -1007,13 +1021,35 @@ def do_is_complete(self, code):
...
         if not self.session:
             return
         content = parent["content"]
-        reply_content = self.do_debug_request(content)
-        if inspect.isawaitable(reply_content):
-            reply_content = await reply_content
+        if not inspect.iscoroutinefunction(self.do_debug_request):
+            # repeat the warning at run
+            reply_content = self.do_debug_request(content)
+            warnings.warn(
```
minrk, line 1042: Not deprecated. This is part of the debugger protocol, not part of IPython Parallel

ipykernel/kernelbase.py
minrk: done in #1282

ipykernel/kernelbase.py
Carreau: Then something in wrong on like 228 of the file (before modifications), 
it does state that some of those are ipyparallel deprecated mesages:

https://github.com/ipython/ipykernel/blob/2ca799276f4458d50085bb1ef8e2ddeb08a283a6/ipykernel/kernelbase.py#L223-L233

ipykernel/kernelbase.py
minrk: There are good reasons to return an awaitable rather than using a coroutine (it can have dramatic performance benefits), and cutting that off due to deficiencies in tooling does not seem like a positive change to me.

mypy [appears to understand](https://mypy-play.net/?mypy=1.11.0&python=3.10&gist=5ab30e161d5e83a16538d790c9c2deb9) `inspect.isawaitable` since 1.11, so this check is correctly handled for type narrowing, but we can also use `isinstance(thing, typing.Awaitable)`, which is equivalent and I think has worked in mypy for longer.

ipykernel/kernelbase.py
minrk: Yes, the comment is outdated, it only applies to the first two message types here. From `debug_request` onward are control-channel-only messages. #1282 removes the deprecated messages and fixes the comment.

ipykernel/kernelbase.py
Carreau: Fine; we can leave it is awaitable; but should we still support both sync and async ?

minrk: `do_xyz` methods were sync-only for a long time, so most implementations are sync because they have no need to be async. It costs us nothing to keep it that I can see, and it breaks users for no benefit, so I don't understand the trade off. In general, I think you need a really good reason to break APIs, and version numbers don't really have any effect on that.

Carreau: > It costs us nothing to keep it that I can see, and it breaks users for no benefit, so 
I don't understand the trade off

It never "cost nothing" to keep optional things like this – it's not the first time you've used this justification, I used to agree, but I've grown.

It cost not much, but on a really long tail – and it accumulates. Keeping it means that every other consumer present or future of `do_debug_request` must keep in mind that it might be sync or async, and test for it – meaning that `await do_debug_request(...)` is technically wrong at it may be sync and consumer then _must_ check for `isawaitable(result)` and await instead.

This also make forgetting an `async` in subclass a subtly incorrect but working implementations:

```
class Sup:

    async def do_debug(self):
        return 'res'

class Derived(Sup):

    def do_debug(self):
        print('Log')
        return super().do_debug()
```

Well you see the kind of issues.

The non-awaitable/non async path is BTW not tested, and the test in this repo assume the method is always a coroutine:

```
tests/test_ipkernel_direct.py
208:async def test_do_debug_request(ipkernel: IPythonKernel) -> None:
211:    await ipkernel.do_debug_request(msg)


tests/conftest.py
106:    async def do_debug_request(self, msg):
```

Personally I got bitten too many time by "it costs nothing" (think shim modules of IPython, and ipython_genutils and alike from a decade ago), that are still regularly used in client projects giving me headache because we should have ripped them out earlier instead of other folks (or us in the future) paying costs for a long time.

Same things with the it can be X or Y  "for convenience" (Api with strings or bytes, $element or iterable of $element where a consumer ends up passing $element which is iterable and tnigs break,... etc) where you always end up chasing a bug for hours and cursing the API.

Hey up until last week I was chasing a bug in reformatting code in darker,  because In IPython we have a test case in IPython for cyrillic, due to some lingering `py3compat.cast_unicode` in IPython ...

And again, I'm not asking to break now; I'm suggesting we tell consumer "hey this is a small but long term cost we will likely not want to incur forever"

minrk: I don't agree with your assessment of the trade-offs in this case and think it shifts more cost to users than is appropriate, but since you feel strongly, if you want to keep the deprecation message for `if not isawaitable(result)` I can accept that.

I do _super_ appreciate your diligence in adding "since when" info to deprecation messages. That has been a huge help for me.

Once all the deprecations about message handler functions are removed (the actually deprecated messages are now gone, debug_request is part of the [standard kernel protocol](https://jupyter-client.readthedocs.io/en/latest/messaging.html#debug-request), not deprecated or related to ipyparallel), this should be all set.

Carreau: > I don't agree with your assessment of the trade-offs in this case and think it shifts more cost to users than is appropriate, but since you feel strongly, if you want to keep the deprecation message for `if not isawaitable(result)` I can accept that.

Would a PendingDeprecationWarning, or a FutureWarning with a softer message ease your concerns ? saying that we are _considering_ making it mandatory, and that we would _prefer_ implementer to always return an awaitable ?



> I do _super_ appreciate your diligence in adding "since when" info to deprecation messages. That has been a huge help for me.

Thanks, it feel so frustrating when message don't give you this information and just give you the stick with "this will be removed in..."; Maybe one day I'll give a talk on why/how you should do that and other things in deprecation warnings and how to make it easier to push users for new API; or maybe just an blog post ? idk.



> Once all the deprecations about message handler functions are removed (the actually deprecated messages are now gone, debug_request is part of the [standard kernel protocol](https://jupyter-client.readthedocs.io/en/latest/messaging.html#debug-request), not deprecated or related to ipyparallel), this should be all set.

I need to redo this PR anyway, some of the deprecation were due to improper comments and documentation; I'll try to get things in a more correct state; and maybe split that into multiple PR

minrk: I think DeprecationWarning or PendingDeprecationWarning is fine, either way.

PR #1275, opened 2024-10-21
fix mixture of sync/async sockets in IOPubThread

minrk:
all sockets are explicitly sync until/except we are in the coroutines that will await them, whereas there was an ambiguous mixture of sync and async sockets before

- consistent behavior of send for child pipe and main process sockets
- avoids unsafe assumption that send is greedy on async sockets
- avoids potential issues creating async objects in one thread, then using them in another in a different event loop
- always creates/uses the right types, regardless of input socket
- address some typing lint

found while working on https://github.com/ipython/ipyparallel/issues/895

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -3,6 +3,8 @@
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+from __future__ import annotations
+
 import atexit
 import contextvars
 import io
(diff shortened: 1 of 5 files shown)


ipykernel/iostream.py
```diff
@@ -199,6 +206,8 @@ def _setup_pipe_in(self):
 
     async def _handle_pipe_msgs(self):
         """handle pipe messages from a subprocess"""
+        # create async wrapper within coroutine
+        self._async_pipe_in1 = zmq.asyncio.Socket(self._pipe_in1)
```
davidbrochart, line 210: Just wondering why it is different than line 170 when there is `Socket.shadow()`?

minrk:
ok, actually all correct now, I think

PR #1280, opened 2024-10-22
restore zero-copy recv on shell messages

### LINKED ISSUE #1278: use of zmq zero-copy is no longer correct
the switch from ZMQStream to async Socket in #1079 has resulted in always receiving all messages with `copy=True`, but there are [checks](https://github.com/ipython/ipykernel/blob/a991039a2e25493bf11744688aedb490e00719b1/ipykernel/kernelbase.py#L277) for `isinstance(always_bytes_object, zmq.Message)` which always results in `copy=True`.

This can get costly because inproc sends with `copy=False` are particularly lightweight, whereas the new unconditional copy _plus_ additional inproc send/recv for subshells result in at least two additional in-memory copies of every message frame received by the kernel.

minrk:
control channel always copies

closes #1278

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -274,10 +274,9 @@ class Kernel(SingletonConfigurable):
         assert self.control_thread is None or threading.current_thread() == self.control_thread
 
         msg = msg or await self.control_socket.recv_multipart()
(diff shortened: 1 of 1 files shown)


ianthomas23 (approved):
Looks fine to me and I defer to @minrk's zmq knowledge here.

PR #1281, opened 2024-10-22
socket must be None, not shell_socket for default shell

minrk:
`process_shell_message` has

```python
assert socket is None
```

for the default subshell, so passing `shell_socket` here results in:

```python
socket = self.shell_socket
...
assert socket is None
socket = self.shell_socket
```


minrk:
alternative change would be to update the assert to `assert socket in {self.shell_socket, None}`, but this seems to match other calls to `shell_main`

ianthomas23 (approved):
Thanks @minrk

PR #1284, opened 2024-10-25
Drop support for Python 3.8

ianthomas23:
Drop support for Python 3.8 as it reached end-of-life on 2024-10-07.


ianthomas23:
I am going to have to modify `test_run_concurrently_sequence`. Because it is based on timing of two different threads/subshells it is sensitive to runtime conditions such as whether coverage is being recorded. I'll do it separately.

PR #1285, opened 2024-10-25
Improve robustness of subshell concurrency tests

ianthomas23:
Improve robustness of subshell concurrency tests. This is difficult to do as the tests run reliably fine on my dev machine so I am looking at what could theoretically improve the situation and trying it out in CI.

I have moved the `import time` to before starting the time-sensitive subshell code as this should be better than importing it in one subshell (taking an unknown amount of time) or importing it in both subshells at the same time (GIL contention?). If this is not sufficient I could add some threaded synchronisation to the start of the subshell test code to remove any timing variability in sending/receiving zmq messages.


minrk (approved):

ianthomas23:
Merging as it should make things better and I can revisit later to improve further.

PR #1288, opened 2024-10-29
Improve robustness of subshell concurrency tests using Barrier

ianthomas23:
Second attempt to improve robustness of subshell concurrency tests, building on #1285. Using a `threading.Barrier` at start of subshell code execution in tests that are sensitive to timing so that they are tolerant of, for example, a shell message that takes slightly longer to send/receive than normal.

I am hoping to fix ipython/ipython#14545 eventually.

### DIFF
--- a/tests/test_subshells.py
+++ b/tests/test_subshells.py
@@ -134,19 +134,22 @@ def test_run_concurrently_sequence(are_subshells, overlap):
             for is_subshell in are_subshells
         ]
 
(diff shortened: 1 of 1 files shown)


Carreau (approved):
+1, should we merge as is ?

ianthomas23:
> +1, should we merge as is ?

Yes please, and I can iterate further if necessary.

PR #1290, opened 2024-11-07
Rely on intrsphinx_registry to keep intersphinx up to date.

Carreau:


minrk:
Nice, I didn't know about this one

Carreau:
> Nice, I didn't know about this one

Who do you think made the package ?

PR #1291, opened 2024-11-07
Use zmq-anyio

davidbrochart:
@minrk this is using an AnyIO-compatible pyzmq API (from https://github.com/davidbrochart/zmq-anyio), as discussed in https://github.com/zeromq/pyzmq/pull/2045.


davidbrochart:
I can see with this PR that ipykernel runs on trio when hard-coding  the backend [here](https://github.com/ipython/ipykernel/blob/fe527201e9377ad4d9e0c497d3a373f0c8dd261c/ipykernel/kernelapp.py#L699), but how can we choose e.g. from JupyterLab to set `trio_loop`?

minrk:
Nice! So it looks like for compatibility with Windows, you've gone with spawning a `select` thread for every waiting socket?

> I can see with this PR that ipykernel runs on trio when hard-coding the backend [here](https://github.com/ipython/ipykernel/blob/fe527201e9377ad4d9e0c497d3a373f0c8dd261c/ipykernel/kernelapp.py#L699), but how can we choose e.g. from JupyterLab to set trio_loop?

This should presumably be the same as any other Kernel configuration option, so a `config=True` traitlet, set e.g. from a CLI arg or env var in the kernelspec (or [parameterized kernel](https://github.com/jupyter/enhancement-proposals/pull/87), or (most compatible) in the kernel config file (`~/.ipython/profile_default/ipython_kernel_config.py`).

davidbrochart:
> So it looks like for compatibility with Windows, you've gone with spawning a `select` thread for every waiting socket?

Yes, and I think Tornado does something similar [here](https://github.com/tornadoweb/tornado/blob/65a9e48f8ce645f104e3e0aa772222e70b0376d9/tornado/platform/asyncio.py#L438). This means a `zmq_anyio.Socket` has to be started asynchronously with an async context manager (or by calling its `start()` method), but that's consistent with an AnyIO socket.
AnyIO's [wait_socket_readable](https://anyio.readthedocs.io/en/stable/api.html#anyio.wait_socket_readable) "does NOT work on Windows when using the asyncio backend with a proactor event loop (default on py3.8+)".

Thanks for the kernel configuration, I'll try that :+1:

minrk:
> I think Tornado does something similar

It does. The big difference is Tornado starts one selector thread per event loop which is far more scalable, whereas zmq-anyio starts one per socket. This makes sense from a library simplicity standpoint since there isn't a thread running when you are done with a socket, but definitely isn't scalable and probably isn't what we should do long term (and why I think anyio should have this built-in, just like tornado).

As I understand it, this means anyio will spawn [up to 40 threads by default](https://anyio.readthedocs.io/en/stable/threads.html#adjusting-the-default-maximum-worker-thread-count). That's _might_ be okay for ipykernel, but I'd say it does mean we shouldn't use zmq-anyio in any client places like jupyter-client or jupyter-server. Because as soon as you've got 40 idle zmq sockets waiting for a message (what they spend most of their time doing), any subsequent calls to `to_thread` will block forever. I do think the first priority here should be universal `wait_socket_readable` support in anyio itself.

You might be able to provoke this in ipykernel by spawning 41 subshells and not using them, since I think each one adds a socket that will be idle.

You could limit the starvation by making the `to_thread` function one with a timeout that you re-enter all the time from the main async thread. Idle sockets will still starve the thread pool, just with timeout-sized bubbles instead of forever.

davidbrochart:
Interesting, I hadn't thought about that.
Maybe before moving that part to AnyIO we could have async `zmq_anyio.start()` and `zmq_anyio.stop()` that would start/stop the thread?

minrk:
> Maybe before moving that part to AnyIO we could have async zmq_anyio.start() and zmq_anyio.stop() that would start/stop the thread?

Sure! I think that's sensible. I don't have enough experience with the task group hierarchy stuff to know what that should look like. I think it's probably appropriate to have some tests in zmq-anyio with a lot of idle sockets (at least more than the thread count, which I think can be set to 1 or 2) to probe this stuff.

If I were the one writing it, I'd implement a `zmq_anyio.wait_readable` that:

1. invokes tornado's SelectorThread logic as needed on asyncio [as done in pyzmq](https://github.com/zeromq/pyzmq/blob/163acca57058ef0dafd3c6bde87c68083c1379f6/zmq/asyncio.py#L30), and
2. calls `trio.lowlevel.wait_readable` on trio
3. (ideally) accepts integer FD so it doesn't need to call `socket.fromfd` (like `trio.lowlevel.wait_readable` and `asyncio.add_reader`, unlike `anyio.wait_socket_readable`) which I think doubles the number of open FDs

You should be able to base it on [anyio.wait_socket_readable](https://github.com/agronholm/anyio/blob/4d3dd2697a8985216bad5538fbfa4be4c9ebcae3/src/anyio/_backends/_asyncio.py#L2674-L2698) which assumes `socket` objects, despite there being no actual need for that restriction (I'm guessing it's inherited because trio used to have this same issue, but [doesn't anymore](https://trio.readthedocs.io/en/stable/reference-lowlevel.html#trio.lowlevel.wait_readable)).

A smaller, but maybe less clean and less efficient version with a one-time monkeypatch:

```python
if windows and asyncio and proactor:
    # only needed once per asyncio event loop, this is the only situation where a patch is needed
    loop = asyncio.get_running_loop()
    loop.add_reader = selector_add_reader # from tornado's AddThreadSelector
    loop.remove_reader = selector_remove_reader # from tornado's AddThreadSelector

...
# assume wait_socket_readable works, which it should now
await anyio.wait_socket_readable(socket.fromfd(zmq_sock.FD))
# hopefully anyio will fix integer FD support to match underlying asyncio and trio
```

If you did any of those, there would be the advantage that no actual thread is spawned except in the Windows + Proactor + asyncio case, which would get exactly one thread.

fwiw, I started to extract the tornado feature into its own package, but haven't tested it enough to publish a release if there's some reason to not depend on tornado for this feature (I don't think there is): https://github.com/minrk/async-selector-thread. Requiring tornado for this doesn't mean the tornado IOLoop object ever needs to be created, as the SelectorThread logic is pure asyncio, so there's really no reason not to require tornado for this as long as it's the only package with the required feature.

minrk:
I had another thought where you could shutdown the thread if nothing is waiting (when remove_reader is called). This might play nicer with anyio's design of shutting things down when they aren't in use, and you don't need anything hooked up to close unless it's called while waiting on a socket. But it comes at a performance cost because you are probably going to recreate the thread a whole bunch of times (once per message if you only have one socket). I don't actually think we should do that, but it's an idea if there are objections to leaving an idle thread running.

But really by far the most efficient approach is ZMQStream's event-driven on_recv, which registers the FD exactly once and calls handle_events whenever there might be a message, rather than calling add_reader and remove_reader for every message.

davidbrochart:
Still a few tests failing, and trio is not enabled in tests (more failures), but this is taking shape.

davidbrochart:
I'm not sure about the tests that time out, it never happens locally on my machine.

davidbrochart:
Tests don't hang anymore, the remaining failures seem to be for [test_print_to_correct_cell_from_child_thread](https://github.com/ipython/ipykernel/blob/b11d42c1f85e30dc87b73b71e79f994106ab82d4/tests/test_kernel.py#L62) on Windows and Ubuntu/PyPy (there are also some `ResourceWarning: unclosed file <_io.BufferedReader name=14>` treated as errors).

davidbrochart:
I think I'll need some help from @krassowski for the failing `test_print_to_correct_cell_*`, and @Carreau for the failures related to IPython:
```
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\ipykernel\iZc2KLS5\test\lib\site-packages\IPython\utils\_process_win32.py", line 124, in system
    return process_handler(cmd, _system_body)
ResourceWarning: unclosed file <_io.BufferedReader name=14>
```

Carreau:
I think we can tracemalloc.enable() and the resource warning will tell us where the leasked thing is created.

Carreau:
Oh, I had forgotten about `PYTHONTRACEMALLOC` env variable.

davidbrochart:
But it doesn't show more information :thinking:

Carreau:
I'm poking at stuff in https://github.com/ipython/ipython/pull/14621 ;

.github/workflows/ci.yml
```diff
@@ -32,30 +32,44 @@ jobs:
...
+          PYTHONTRACEMALLOC=20 hatch run test:nowarn || PYTHONTRACEMALLOC=20 hatch run test:nowarn --lf
 
       - name: Run the tests on Windows
         timeout-minutes: 15
         if: ${{  startsWith(matrix.os, 'windows') }}
         run: |
-          hatch run cov:nowarn || hatch run test:nowarn --lf
+          hatch run test:pip list
+          hatch run test:python --version
+          PYTHONTRACEMALLOC=20 hatch run test:pytest -v
```
Carreau, line 72:
```suggestion
          hatch run test:pip list
          hatch run test:python --version
          hatch run test:python -m pip install pip install git+https://github.com/ipython/ipython@92dd9e47fe8862ee38770744c165b680cb5241b1
          PYTHONTRACEMALLOC=20 hatch run test:pytest -v
```

Let's try that,

.github/workflows/ci.yml
```diff
@@ -32,30 +32,45 @@ jobs:
...
+          PYTHONTRACEMALLOC=20 hatch run test:nowarn || PYTHONTRACEMALLOC=20 hatch run test:nowarn --lf
 
       - name: Run the tests on Windows
         timeout-minutes: 15
         if: ${{  startsWith(matrix.os, 'windows') }}
         run: |
-          hatch run cov:nowarn || hatch run test:nowarn --lf
+          hatch run test:pip list
+          hatch run test:python --version
+          hatch run test:python -m pip install pip install git+https://github.com/ipython/ipython@92dd9e47fe8862ee38770744c165b680cb5241b1
```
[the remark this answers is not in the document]
Carreau:
```suggestion
          hatch run test:pip list
          hatch run test:python --version
          hatch run test:python -m pip install git+https://github.com/ipython/ipython@92dd9e47fe8862ee38770744c165b680cb5241b1
```

Carreau:
that ... was not it

I will see if I can create a fast branch that only run the failing test, and iterate on commenting things in IPython to see where the buffer can come from. But I do not have windows, so it's a painful process.

davidbrochart:
Thanks @Carreau.
I installed a Windows VM, I'll see if I can reproduce locally.

Carreau:
I would also not be surprised for this kind of errors to depend on previous tests.

krassowski:
I would like to understand how this PR fits into v7 of ipykernel and the motivation for this change, is there an issue describing what are the benefits for users/developers and the plan in more general?

davidbrochart:
Ipykernel is now based on AnyIO, but since [pyzmq](https://github.com/zeromq/pyzmq) doesn't support AnyIO, it prevents ipykernel using a Trio event loop. After talking with @minrk, I started [zmq-anyio](https://github.com/davidbrochart/zmq-anyio) to fill that gap and we upstream Windows support for reading reading sockets in the `ProactorEventLoop` in https://github.com/agronholm/anyio/pull/831, which is required by `zmq`.
Without this PR, we would loose Trio support in ipykernel, which is too bad. I think I'll need help to fix the tests, if we want to have this PR in.
What do you think @minrk and @krassowski ?

krassowski:
Apologies for rather fundamental questions, I am just not very familiar with the topic.

> it prevents ipykernel using a Trio event loop

Does using Trio event loop in 6.x require users to opt-in? How widely is it used? What are its advantages?

davidbrochart:
> Does using Trio event loop in 6.x require users to opt-in?

Yes, it was [opt-in](https://github.com/ipython/ipykernel/blob/830829fd344849c69027a7a198f745e7b84bcb28/ipykernel/kernelapp.py#L172).

> How widely is it used?

Hard to say if it is used a lot in ipykernel, but Trio is getting more and more popular:

star-history-2025211

> What are its advantages?

Trio pioneered [structured concurrency](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful), which is what AnyIO brings to asyncio. We use it [pycrdt-websocket](https://github.com/jupyter-server/pycrdt-websocket), [jupyverse](https://github.com/jupyter-server/jupyverse/pull/453)... In general, level-based cancellation is considered to be a superior solution to edge-based cancellation.

Carreau:
I also strongly recommends the other blogposts that accompany trio release 

https://vorpus.org/blog/archives.html, starting with https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/ 

IPython tries to support trio as well so +1 for trio.

davidbrochart:
The difficulty with this PR is that some tests fail on some platforms but pass on others, and all tests pass on at least one platform, so it's not like it's not working at all, which makes it hard to find the issue. It could be in `zmq-anyio` or in `ipykernel`, I don't know.

davidbrochart:
https://github.com/ipython/ipykernel/pull/1291/commits/b55589244926cbc88cecddd419094b3beaa0b188 is a workaround for not suspending the AnyIO socket selector thread (on Windows with a ProactorEventLoop) when stopped on a breakpoint (I opened https://github.com/agronholm/anyio/issues/869). Now the debugger works on Windows too.
Remaining failures:
- Unclosed files in `IPython\utils\_process_win32.py` (cc @Carreau).
- `zmq.error.ZMQError: Socket operation on non-socket`
- `Fatal Python error: Aborted`
- timeout...

But it could be about the way tests are written too.

.github/workflows/ci.yml
```diff
@@ -49,19 +55,22 @@ jobs:
...
+          PYTHONTRACEMALLOC=20 hatch run test:nowarn || PYTHONTRACEMALLOC=20 hatch run test:nowarn --lf
 
       - name: Run the tests on Windows
         timeout-minutes: 15
         if: ${{  startsWith(matrix.os, 'windows') }}
         run: |
-          hatch run cov:nowarn || hatch run test:nowarn --lf
+          hatch run test:pip list
+          hatch run test:python --version
+          #hatch run test:python -m pip install git+https://github.com/ipython/ipython@92dd9e47fe8862ee38770744c165b680cb5241b1
```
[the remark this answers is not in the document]
davidbrochart: With this it looks like we can't test e.g. Python 3.9 because of IPython.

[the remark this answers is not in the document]
Carreau: Let's already see if this passes on non-python 3.9, we can see after.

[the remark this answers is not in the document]
Carreau: But it seem this hangs anyway, I'll try something else.

.github/workflows/ci.yml
```diff
@@ -49,19 +55,22 @@ jobs:
...
+          PYTHONTRACEMALLOC=20 hatch run test:nowarn || PYTHONTRACEMALLOC=20 hatch run test:nowarn --lf
 
       - name: Run the tests on Windows
         timeout-minutes: 15
         if: ${{  startsWith(matrix.os, 'windows') }}
         run: |
-          hatch run cov:nowarn || hatch run test:nowarn --lf
+          hatch run test:pip list
+          hatch run test:python --version
+          hatch run test:python -m pip install git+https://github.com/ipython/ipython@f9899071478fab03d9c7cf5c982141f2fa0af7d7
```
Carreau, line 72:
```suggestion
          hatch run test:python -m pip install git+https://github.com/ipython/ipython@ea303f8330fddbc786e1fcbf31e881feb7a606b3
```

This just adds a gc.collect to try to narrow down where the unclosed things are.

Carreau:
While you are on this, 

The `hatch run typing:test` will miss a bunch of improper types as anyio (and other libs), are missing from the pre-commit-yaml config files for extra types. Maybe that can help you to catch some issues.

davidbrochart:
Thanks for working on this @Carreau. It looks like the remaining failures are a timeout or caused by IPython, only on Windows (if we exclude PyPy)?

Carreau:
yeah, if we can find a way to ignore the windows failure, i can postpone trying to debug it. otherwise Incan try this week. 

One thing i don't like is the jupyter core still I think uses run_sync via asyncio when calling shutdown kernel, so even in trio, there is an asyncio loop running. 

To verify, and not due to this PR, so not blocking. 
l

davidbrochart:
> One thing i don't like is the jupyter core still I think uses run_sync via asyncio when calling shutdown kernel, so even in trio, there is an asyncio loop running.

You mean in `jupyter-client`? But this is not the same process as ipykernel, so having its own event loop should be fine?
Otherwise `jupyter-client` could use zmq-anyio at some point.

Carreau:
Hum, you are likely right. I would need to recheck, but it does make sens.

Carreau:
https://github.com/ipython/ipython/pull/14749 seems to fix the windows issue. I've push a commit here to ignore the warning in the meantime.

Carreau:
> [ipython/ipython#14749](https://github.com/ipython/ipython/pull/14749) seems to fix the windows issue

Well it fixes it, but creates more issues when magics can't be interrupted by Ctrl-C on win anymore...

davidbrochart:
So it seems that the remaining failures are because of a timeout, or related to subshells?

Carreau:
> So it seems that the remaining failures are because of a timeout, or related to subshells?

Outside of PyPy, it seem so. There are some things that show stuck history threads, but those are just bad cleanup on IPython side and should not create deadlock – and should be less aparent with https://github.com/ipython/ipython/pull/14739 – there will always be one, as there is always at least 1 history mananager.

davidbrochart:
Should we continue testing PyPy? I think it's a great project, but I'm seeing projects giving less and less importance to it, now that CPython is getting a JIT.

ianthomas23:
Note that the `inprocess` tests were disabled in the original `anyio` PR and haven't been reenabled yet:

https://github.com/ipython/ipykernel/blob/f9cfd11b0751125a584a3545cceac7cf0e9b6651/pyproject.toml#L160

davidbrochart:
@ianthomas23 do you think some subshell tests could be flaky?

ianthomas23:
> @ianthomas23 do you think some subshell tests could be flaky?

Yes, some are based on timings which can be risky in CI. Probably I need to write some different tests that do not rely on timings. But also you may have changed some implementation detail that subshells rely upon.

davidbrochart:
Yes zmq-anyio probably changes the order of events, but tests should not depend on that.
The other test failures are either:
- related to IPython,
- a timeout,
- on PyPy.

I'm not sure the failures on PyPy should be ignored or if they can give a hint to some real issues, like the timeout.

Carreau:
for IPython's 

```
ResourceWarning: unclosed file <_io.BufferedReader name=14>
```

I'm aware, but that's beyond my windows knowledge on how to fix. 

Either we close the srtream, but then magics can't get interrupted;
of we leave them dangling, and are not sure if they leak.

Best is to ignore for now.

davidbrochart:
All tests pass except for [ipyparallel](https://github.com/ipython/ipykernel/actions/runs/13436581861/job/37540173368?pr=1291) :eyes:

davidbrochart:
It seems that the only relevant change was removing these architectures from the CI:
```
          - os: windows-latest
            python-version: "3.10"
          - os: windows-latest
            python-version: "3.11"
          - os: windows-latest
            python-version: "3.12"
```
I'll try adding them back just to see.

davidbrochart:
Nevermind, it was running plain `pytest` instead of `test:nowarn`.

ipykernel/heartbeat.py
```diff
@@ -92,13 +92,17 @@ def _bind_socket(self):
     def run(self):
         """Run the heartbeat thread."""
         self.name = "Heartbeat"
-        self.socket = self.context.socket(zmq.ROUTER)
-        self.socket.linger = 1000
+
         try:
+            self.socket = self.context.socket(zmq.ROUTER)
```
Carreau, line 97: it may be a real dumb question, but could/should this be refactored (in another PR maybe) into a context manager:

```
with self.context.socket(zmq.ROUTER) as self.socket:
    ...
```

ipykernel/iostream.py
```diff
@@ -48,7 +49,7 @@ class IOPubThread:
     whose IO is always run in a thread.
     """
 
-    def __init__(self, socket, pipe=False):
+    def __init__(self, socket: zmq_anyio.Socket, pipe=False):
```
Carreau, line 52:
```suggestion
    def __init__(self, socket: zmq_anyio.Socket, pipe:bool=False):
```

tests/test_async.py
```diff
@@ -30,24 +30,23 @@ def test_async_await():
     assert content["status"] == "ok", content
 
 
-# FIXME: @pytest.mark.parametrize("asynclib", ["asyncio", "trio", "curio"])
 @pytest.mark.skipif(os.name == "nt", reason="Cannot interrupt on Windows")
-@pytest.mark.parametrize("asynclib", ["asyncio"])
-def test_async_interrupt(asynclib, request):
+@pytest.mark.parametrize("anyio_backend", ["asyncio"])  # FIXME: %autoawait trio
+def test_async_interrupt(anyio_backend, request):
```
Carreau, line 35: Just a note that there is a TODO here.

tests/test_embed_kernel.py
```diff
@@ -109,7 +109,10 @@ def test_embed_kernel_basic():
     with setup_kernel(cmd) as client:
         # oinfo a (int)
         client.inspect("a")
-        msg = client.get_shell_msg(timeout=TIMEOUT)
+        while True:
+            msg = client.get_shell_msg(timeout=TIMEOUT)
+            if msg["msg_type"] == "inspect_reply":
+                break
```
Carreau, line 115: does this have risk of hanging forever ?

tests/test_io.py
```diff
@@ -12,31 +12,41 @@
...
 import pytest
 import zmq
-import zmq.asyncio
+import zmq_anyio
+from anyio import create_task_group
 from jupyter_client.session import Session
 
 from ipykernel.iostream import _PARENT, BackgroundSocket, IOPubThread, OutStream
 
+pytestmark = pytest.mark.anyio
```
Carreau, line 21:
```suggestion
```

I think this is unused now.

Of if it's a module fixture, there should be a comment.

ipykernel/heartbeat.py
davidbrochart: Maybe, I was just trying to get rid of a failure in [test_merge_connection_file](https://github.com/ipython/ipykernel/blob/28d096d57e4750ab4a92a69e3c6544140b384729/tests/test_kernelapp.py#L50), which creates an `IPKernelApp`, initializes its heartbeat socket and thread, and closes the ZMQ context, leaving the thread working on a closed ZMQ context :cry: 
It was just working by chance before.

Carreau (commented):
+1

tests/test_io.py
davidbrochart: I had to add it back in https://github.com/ipython/ipykernel/pull/1291/commits/192d62edd314199b15fba0bac766f3428c83724c.

tests/test_embed_kernel.py
davidbrochart: I removed it, it seems to be fine now.

davidbrochart:
I had to rerun failing tests, which means that there's some flakiness.
Failures in ipyparallel are expected because it will have to use `zmq_anyio.Socket` instead of `zmq.anyio.Socket`.
How do you feel about merging and focusing on resolving the flakiness?

davidbrochart:
Something is not clear to me, `%autoawait` not only accepts `on` or `off`, but also `asyncio` or `trio`. But whether the kernel is running with an asyncio or a Trio event loop is a [configuration](https://github.com/ipython/ipykernel/blob/72d7ca8247bafde33e693a03545dcfb452c79b25/ipykernel/kernelapp.py#L182), so it cannot be switched at runtime, right?

davidbrochart:
Yeah in ipykernel the event loop used for ipykernel itself to run (handle ZMQ messages, etc.) is the same event loop that runs user code. I think this is the main difference with IPython, which as you said doesn't have an outer event loop.
This means that in ipykernel if the event loop is asyncio, one cannot use `%autoawait trio`, and if the event loop is Trio, one cannot use `%autoawait asyncio`. Basically, in ipykernel one can just use `%autoawait on` or `%autoawait off`, and if `on` then the configured event loop is used. Am I right?

Carreau:
With this PR, my guess is yes, if we want multiple loop and `autoawait anotherlib` we need to update this:

```diff
async def do_execute(
        ...
        async def run(execution: Execution) -> None:
+           # here it's await if same same loop
+           if same_loop:
                execution.result = await coro
+           else:
+                # or the otherlib.run(coro) if another loop
            if not execution.interrupt:
                self.shell_interrupt.put(False)

        res = None
        try:
            async with create_task_group() as tg:
                execution = Execution()
                self.shell_is_awaiting = True
                tg.start_soon(run, execution)
                execution.interrupt = await to_thread.run_sync(self.shell_interrupt.get)
                self.shell_is_awaiting = False
                if execution.interrupt:
                    tg.cancel_scope.cancel()

                res = execution.result
        finally:
            shell.events.trigger("post_execute")
            if not silent:
                shell.events.trigger("post_run_cell", res)
```

I think.

But if we do that, may not have tasks that can span more than the duration of 1 cell.

davidbrochart:
I think mixing event loops in the same notebook is a bad idea anyway, that should be forbidden.

davidbrochart:
In [this timeout failure](https://github.com/ipython/ipykernel/actions/runs/13516469444/job/37766070718?pr=1291#step:5:982) I see 28 `IPythonHistorySavingThread` threads, isn't that a lot?

Carreau:
that should be fixed in Ipython main. It's just that tests create a lot of Shell instances, each with a history saving thread, that was clean only on atexit. now they should be cleaned on garbage collect.

davidbrochart:
That would be interesting to try in this PR, do you expect to release IPython soon?

Carreau:
I published a beta yesterday, and the relevant PR is https://github.com/ipython/ipython/pull/14739, I doubt it will have an effect – or at least I don't see how the history thread could block anything else.

davidbrochart:
I get this error, but maybe I'm doing it wrong?
```
  File "/Users/runner/Library/Application Support/hatch/env/virtual/ipykernel/nycfXNUk/cov.qt5/lib/python3.11/site-packages/IPython/core/history.py", line 688, in __init__
    assert len(HistoryManager._instances) <= HistoryManager._max_inst, (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: (2, 1.0)
```

Carreau:
Yeah, that's on purpose, it tells you you have a history manager leaking, likely in one of the tests that just precede.

Carreau:
with tracemalloc I can tell you that the leaking history_managers are contructed like that:

```
 |  File "/Users/bussonniermatthias/dev/ipykernel/tests/conftest.py", line 213
 |    kernel = MockIPyKernel()
 |  File "/Users/bussonniermatthias/dev/ipykernel/tests/conftest.py", line 198
 |    super().__init__(*args, **kwargs)
 |  File "/Users/bussonniermatthias/dev/ipykernel/ipykernel/ipkernel.py", line 144
 |    self.shell = self.shell_class.instance(
 |  File "/Users/bussonniermatthias/dev/traitlets/traitlets/config/configurable.py", line 583
 |    inst = cls(*args, **kwargs)
 |  File "/Users/bussonniermatthias/dev/ipython/IPython/core/interactiveshell.py", line 623
 |    self.init_history()
 |  File "/Users/bussonniermatthias/dev/ipython/IPython/core/interactiveshell.py", line 1906
 |    self.history_manager = HistoryManager(shell=self, parent=self)
 |  File "/Users/bussonniermatthias/dev/traitlets/traitlets/traitlets.py", line 1289
 |    inst = new_meth(cls)
```

There is likeley something keeping them alive.  I doubt they are the cause of any failure her though.

Carreau:
the fixture were yielding outside of a try/except, and were not properly destroying kernels. 

I pushed a commit that should close many of the leaks.

At least locally tests/test_ipkernel_direct.py almost passes.

More failure will be fixed once we have #1352

davidbrochart:
Thanks, it seems that the only remaining failure is a timeout in `test_kernelapp.py::test_start_app` on Windows?

Carreau:
Just pushed another commit that fix 2 leaks in tests/test_zmq_shell.py

Carreau:
There are no more leaks, or history thread waiting. I hope this will help you track the last errors.

Carreau:
Oh, well, that last issues a re because `_atexit_once` was introduced in https://github.com/ipython/ipython/pull/12898 I'll let you either figure it out, or bump min IPython to 8.0 that was relased more than 3 years ago.

davidbrochart:
I'm not sure why more tests hang now, but being on IPython `main` complicates the picture.
On the other hand tests fail on ipykernel `main` too, so maybe it's not because of zmq-anyio.

minrk:
awesome work, @davidbrochart!

PR #1294, opened 2024-11-13
Misc type annotations

Carreau:
Extracted from #1272

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -54,6 +54,7 @@ from .iostream import IOPubThread
 from .ipkernel import IPythonKernel
 from .parentpoller import ParentPollerUnix, ParentPollerWindows
 from .shellchannel import ShellChannelThread
+from .thread import BaseThread
 from .zmqshell import ZMQInteractiveShell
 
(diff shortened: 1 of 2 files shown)


ipykernel/kernelapp.py
```diff
@@ -142,9 +143,10 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp, ConnectionFileMix
     debug_shell_socket = Any()
     stdin_socket = Any()
     iopub_socket = Any()
-    iopub_thread = Any()
-    control_thread = Any()
-    shell_channel_thread = Any()
+
+    iopub_thread: BaseThread
```
minrk, line 147: I think these still need to be defined as traits, e.g. `: BaseThread = Any()` or perhaps even `Instance(BaseThread, allow_none=True)`.

Carreau: I think that will need a `# type: ignore[assign]` as well, or mypy will be unhappy

Carreau:
Updated, and fixed the corresponding error shown by mypy.

minrk:
Thanks!

Carreau:
thanks for the review and merge !

PR #1295, opened 2024-11-13
Suggest to make implementations of some function always return awaitable

Carreau:
See discussion in #1272; It is not deprecated, but being able to always know you can (and must) await should be simpler in the long run.

Deprecating now is not the point, but I want to cover our bases, so that we are more confident later when and if we want to enforce those await.

In particular many of those branches are not covered in our tests – and I don't even know wether they were ever taken;

After changing some of the base methods to be async, it breaks qtconsole, spyder kernel and a few things– so only emitting warnings for now.

A few other things use the `if awaitable(...):` pattern but are a bit more complicated, and some do not date back to 2021, so those will be dealt with separately.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -60,7 +60,7 @@ from .iostream import OutStream
 _AWAITABLE_MESSAGE: str = (
     "For consistency across implementations, it is recommended that `{func_name}`"
     " either be a coroutine function (`async def`) or return an awaitable object"
-    " (like a Future). It might become a requirement in the future."
+    " (like an `asyncio.Future`). It might become a requirement in the future."
     " Coroutine functions and awaitables have been supported since"
     " ipykernel 6.0 (2021)."
 )
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -57,6 +57,14 @@ from ipykernel.jsonutil import json_clean
 from ._version import kernel_protocol_version
 from .iostream import OutStream
 
+_AWAITABLE_MESSAGE: str = (
+    "For consistency across implementations, it is recommended that `{func_name}`"
+    " either be a coroutine function (`async def`) or return an awaitable object"
+    " (like a Future). It might become a requirement in the future."
+    " Coroutine functions and awaitables have been supported since"
+    " ipykernel 6.0 (2021)."
+)
+
 
 def _accepts_parameters(meth, param_names):
     parameters = inspect.signature(meth).parameters
@@ -742,6 +750,12 @@ class Kernel(SingletonConfigurable):
 
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="execute_request"),
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
 
         # Flush output before sending the reply.
         if sys.stdout is not None:
@@ -802,6 +816,12 @@ class Kernel(SingletonConfigurable):
         matches = self.do_complete(code, cursor_pos)
         if inspect.isawaitable(matches):
             matches = await matches
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="do_complete"),
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
 
         matches = json_clean(matches)
         self.session.send(socket, "complete_reply", matches, parent, ident)
@@ -830,6 +850,12 @@ class Kernel(SingletonConfigurable):
         )
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="do_inspect"),
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
 
         # Before we send this object over, we scrub it for JSON usage
         reply_content = json_clean(reply_content)
@@ -849,6 +875,12 @@ class Kernel(SingletonConfigurable):
         reply_content = self.do_history(**content)
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="do_history"),
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
 
         reply_content = json_clean(reply_content)
         msg = self.session.send(socket, "history_reply", reply_content, parent, ident)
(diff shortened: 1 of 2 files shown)


Carreau:
I'm going to revert making the default implementation awaitable later, but currently this breaks some downstream things in the way I was expecting as consumers don't check wether the return is an awaitable or not.

minrk:
I don't object to using 7.0 to make all base class `do_` methods async. It is a breaking change, but a lot is already broken by 7.0. It is a difficult balance to strike, and I know I tend to come down on the side of not breaking things. The plus side is that subclasses can make these methods async without losing backward compatibility since async has been _allowed_ for quite some time.

One thing I'll note is that allowing subclasses to implement methods as sync or async is one thing (what we do in ipykernel), whereas changing a base class implementation to be async or not (has happened some places in 7.0 I think, this PR makes it more consistent) is definitely a straightforwardly breaking change. It is not the case that the current API where subclasses may define methods as async means that the base class can switch between sync and async without it being considered a major breaking change. Allowing multiple implementations of these methods is really the point of the Kernel class, but so is a stable base implementation.

Carreau:
@minrk can you review the new wording and if you agree ? 
The sometime async/sometime sync is the reasons qtconsole is failing downstream (https://github.com/jupyter/qtconsole/pull/629)

ipykernel/kernelbase.py
```diff
@@ -830,6 +850,12 @@ async def inspect_request(self, socket, ident, parent):
         )
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="inspect_request"),
```
minrk, line 855:
```suggestion
                _AWAITABLE_MESSAGE.format(func_name="do_inspect"),
```

ipykernel/kernelbase.py
```diff
@@ -849,6 +875,12 @@ async def history_request(self, socket, ident, parent):
         reply_content = self.do_history(**content)
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="history_request"),
```
minrk, line 880:
```suggestion
                _AWAITABLE_MESSAGE.format(func_name="do_history"),
```

ipykernel/kernelbase.py
```diff
@@ -990,6 +1028,12 @@ async def is_complete_request(self, socket, ident, parent):
         reply_content = self.do_is_complete(code)
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="do_execute"),
```
minrk, line 1033:
```suggestion
                _AWAITABLE_MESSAGE.format(func_name="do_is_complete"),
```

ipykernel/kernelbase.py
```diff
@@ -1006,6 +1050,12 @@ async def debug_request(self, socket, ident, parent):
         reply_content = self.do_debug_request(content)
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="debug_request"),
```
minrk, line 1055:
```suggestion
                _AWAITABLE_MESSAGE.format(func_name="do_debug_request"),
```

minrk (approved):
LGTM, thanks! I suggested some fixes to the function names in the warnings, but the warning text looks great.

I'll reiterate that it is _not_ the optional support for async subclasses that caused a problem in Qt, it is the change in the _base class_ implementation, which is not implied by the flexibility afforded to subclasses (there was never an implication that Qt needed to handle this, ipykernel just made a breaking change).

But I'm still +1 on a consistent switch to async with the warning as in this PR.

ipykernel/kernelbase.py
```diff
@@ -802,11 +816,17 @@ async def complete_request(self, socket, ident, parent):
...
+                _AWAITABLE_MESSAGE.format(func_name="do_complete"),
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
 
         matches = json_clean(matches)
         self.session.send(socket, "complete_reply", matches, parent, ident)
 
-    def do_complete(self, code, cursor_pos):
+    async def do_complete(self, code, cursor_pos):
```
Carreau, line 829:
```suggestion
    def do_complete(self, code, cursor_pos):
```

ipykernel/kernelbase.py
```diff
@@ -974,7 +1012,7 @@ async def shutdown_request(self, socket, ident, parent):
 
         self.stop()
 
-    def do_shutdown(self, restart):
+    async def do_shutdown(self, restart):
```
Carreau, line 1015:
```suggestion
    def do_shutdown(self, restart):
```

ipykernel/kernelbase.py
```diff
@@ -57,6 +57,14 @@
 from ._version import kernel_protocol_version
 from .iostream import OutStream
 
+_AWAITABLE_MESSAGE: str = (
+    "For consistency across implementations, it is recommended that `{func_name}`"
+    " either be a coroutine function (`async def`) or return an awaitable object"
+    " (like a Future). It might become a requirement in the future."
```
davidbrochart, line 63:
```suggestion
    " (like an `asyncio.Future`). It might become a requirement in the future."
```

ipykernel/kernelbase.py
```diff
@@ -849,6 +875,12 @@ async def history_request(self, socket, ident, parent):
         reply_content = self.do_history(**content)
         if inspect.isawaitable(reply_content):
             reply_content = await reply_content
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="do_history"),
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
```
Carreau, line 883:
```suggestion
        else:
            pass
            # commenting this to see if printing is the issue with saving thread blocing in test.
            #warnings.warn(
            #    _AWAITABLE_MESSAGE.format(func_name="do_history"),
            #    PendingDeprecationWarning,
            #    stacklevel=1,
            #)
```

ipykernel/kernelbase.py
```diff
@@ -802,6 +816,12 @@ async def complete_request(self, socket, ident, parent):
         matches = self.do_complete(code, cursor_pos)
         if inspect.isawaitable(matches):
             matches = await matches
+        else:
+            warnings.warn(
+                _AWAITABLE_MESSAGE.format(func_name="do_complete"),
+                PendingDeprecationWarning,
+                stacklevel=1,
+            )
```
Carreau, line 824:
```suggestion
        else:
            pass
            # warnings.warn(
            #     _AWAITABLE_MESSAGE.format(func_name="do_complete"),
            #     PendingDeprecationWarning,
            #     stacklevel=1,
            # )
```

Carreau:
Ok, as everything seem to be passing and there is no objection, I'm going to self-merge to move forward. 

The only problem is tests was that Pytest enable deprecation into errors, and so test were failing.

PR #1296, opened 2024-11-18
Use supported_features=['debugger'] in kernel info reply

ianthomas23:
[JEP92](https://github.com/jupyter/enhancement-proposals/pull/92/files) added support for optional features to be included in kernel info reply messages via the `supported_features` attribute, with subshells and debugger being examples. We are already using it for subshells, this PR also adds `debugger` to the list of `supported_features` if `debugpy` is importable.

Corresponding documentation PR in `jupyter_client` for messaging protocol is jupyter/jupyter_client#1045.

This replaces the use of `kernel_info_reply['debugger']` which should be considered deprecated but cannot be removed until downstream libraries adopt reading the `supported_features` rather than `debugger` attribute directly.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -895,7 +895,7 @@ class Kernel(SingletonConfigurable):
             "language_info": self.language_info,
             "banner": self.banner,
             "help_links": self.help_links,
-            "supported_features": supported_features
+            "supported_features": supported_features,
         }
 
     async def kernel_info_request(self, socket, ident, parent):
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -891,6 +891,12 @@ class Kernel(SingletonConfigurable):
         }
         if self._supports_kernel_subshells:
             info["supported_features"] = ["kernel subshells"]
+
+        from .debugger import _is_debugpy_available
+
+        if _is_debugpy_available:
+            info["supported_features"].append("debugger")
+
         return info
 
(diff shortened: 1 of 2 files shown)


ipykernel/kernelbase.py
```diff
@@ -880,18 +880,23 @@ async def connect_request(self, socket, ident, parent):
 
     @property
     def kernel_info(self):
-        info = {
+        from .debugger import _is_debugpy_available
+
+        supported_features: list[str] = []
+        if self._supports_kernel_subshells:
+            supported_features.append("kernel subshells")
```
davidbrochart, line 887: Is it necessary to have "kernel", or could we just put "subshells"?

davidbrochart:
> We are already using it for subshells, this PR also adds `debugger` to the list of `supported_features` if `debugpy` is importable.

I don't understand, it seems that this PR does both add `"debugger"` and `"kernel subshells"` to the list of `supported_features`, right?

ipykernel/kernelbase.py
ianthomas23: "kernel subshells" is what is says in the JEP (https://github.com/jupyter/enhancement-proposals/blob/master/kernel-subshells/kernel-subshells.md#identify-optional-feature) so that is what it must be.

ianthomas23:
> I don't understand, it seems that this PR does both add `"debugger"` and `"kernel subshells"` to the list of `supported_features`, right?

No, "kernel subshells" was already there but I've rearranged the code a little.

davidbrochart:
Ah yes I see.

PR #1297, opened 2024-11-20
Show Python version in CI

davidbrochart:


davidbrochart:
It looks like Pyhton 3.9.13 is installed in `build (windows-latest, 3.13)` (see [here](https://github.com/ipython/ipykernel/actions/runs/11932248022/job/33256829004?pr=1297)).

PR #1300, opened 2024-11-21
Replace BaseThread's add_task with start_soon

davidbrochart:
This PR replaces a `BaseThread`'s `add_task()` method with `start_soon()`. The new name is less confusing as it's the same as in AnyIO, and it allows to start a task in the thread even after the thread has been started.
We also get rid of `_IOPubThread`, which has no reason to be different than a `BaseThread`.

(from #1291)


davidbrochart:
I updated the PR with my latest changes in #1291.

ipykernel/thread.py
```diff
@@ -14,29 +19,59 @@ class BaseThread(Thread):
...
+        self._tasks.put(("run_sync", func))
+        return self._result.get()
+
+    def run(self) -> None:
         """Run the thread."""
-        return run(self._main)
+        try:
+            run(self._main)
+        except Exception:
+            pass
```
Carreau, line 49: Do you want to at least log something if there is an exception ? It seems weird to me to swallow exception, but I'll trust you there.

davidbrochart: True, but I'm not sure if this would play well with threads like iostream, which does something with writing to `stdout`/`stderr`?
Maybe the exception could be made available through a method instead?

PR #1305, opened 2024-12-12
properly close OutStream and various fixes

### LINKED ISSUE #1283: Cleanup file descriptors watcher thread
It seems that IOStreams are not correcly calling their `close` method as in:

https://github.com/ipython/ipykernel/blob/2ca799276f4458d50085bb1ef8e2ddeb08a283a6/ipykernel/iostream.py#L579-L596

Thus, even though `reset_io` is called when closing the IPKernelApp:

https://github.com/ipython/ipykernel/blob/2ca799276f4458d50085bb1ef8e2ddeb08a283a6/ipykernel/kernelapp.py#L512-L519

`sys.stdout` is still not functioning as expected after IPyKernelApp closes, because the writes are being redirected.

limwz01:
In order to support proper execution after resuming from an embedded kernel, we have to make sure that we can revert the IO alterations by the kernelapp using the `close()` method. The following fixes were made:
* Call `close` before `embed` returns
* Don't write the extra character on `OutStream` `close()` to pipe for watching thread if it's already dead or missing
* Add new variable `closed` so that "close" can be called multiple times without any error
* Don't use `__stdout__` and `__stderr__` at all
* Keep `blackhole` reference around so that the watching thread's pipe file descriptors do not get accidentally closed when the object is garbage collected (it happens now because it is not stored in `__stdout__`)
* Keep old IO variables and restore them when exiting
* More robust StreamHandler fixup and restore (but I think these should actually be completely removed)

Fixes #1283

### DIFF
--- a/ipykernel/_version.py
+++ b/ipykernel/_version.py
@@ -1,6 +1,8 @@
 """
 store the current version info of the server.
 """
+from __future__ import annotations
+
 import re
 
 # Version string must appear intact for hatch versioning
--- a/ipykernel/embed.py
+++ b/ipykernel/embed.py
@@ -55,3 +55,4 @@ def embed_kernel(module=None, local_ns=None, **kwargs):
     app.kernel.user_ns = local_ns
     app.shell.set_completer_frame()  # type:ignore[union-attr]
     app.start()
+    app.close()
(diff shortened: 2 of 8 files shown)


limwz01:
Fixes https://github.com/ipython/ipykernel/issues/1283

limwz01:
I have no idea why the tests are failing and I currently do not have the appropriate python version to run the tests. But I'll work on it.

limwz01:
ok it's all fixed

minrk:
Thank you!

PR #1223, opened 2024-03-25
Do not import debugger/debugpy unless needed

krassowski:
Follow-up to https://github.com/ipython/ipykernel/pull/1213.

This avoid spurious warnings on kernel management operations, like calling `python -m ipykernel install --sys-prefix`.

What do you think?

Issue #1198, opened 2024-01-16
python 3.11 debugpy error messages


mlucool:
Hi,

Using python 3.11 we see the following error:
```
$ python -m ipykernel install --name mykernel --user
0.00s - Debugger warning: It seems that frozen modules are being used, which may
0.00s - make the debugger miss breakpoints. Please pass -Xfrozen_modules=off
0.00s - to python to disable frozen modules.
0.00s - Note: Debugging will proceed. Set PYDEVD_DISABLE_FILE_VALIDATION=1 to disable this validation.
Installed kernelspec mykernel in /u/myuser/.local/share/jupyter/kernels/mykernel
```

This comes form https://github.com/ipython/ipykernel/blob/93a63fb7b8752899ed95118fa35e56f74eedd0c6/ipykernel/debugger.py#L24.

This was also posted on stackoverflow here: https://stackoverflow.com/questions/75114841/debugger-warning-from-ipython-frozen-modules

Can we fix ipykernel to only show this message when it's applicable (i.e. when trying to debug)?

krassowski:
Thank you for opening this issue! This was previously reported a few times on Jupyter Discourse and it is really confusing for users (who sometimes mis-attribute other issues to this warning as it is common and easy to blame).

In particular there is some relevant discussion in https://discourse.jupyter.org/t/debugger-warning-it-seems-that-frozen-modules-are-being-used-python-3-11-0/16544

> Can we fix ipykernel to only show this message when it's applicable (i.e. when trying to debug)?

ipykernel imports debugpy to check if it is available; this is hard to check without import, because the import can error out even if debugpy is installed see https://github.com/ipython/ipykernel/issues/986.

I previously suggested always appending `-Xfrozen_modules=off` on a higher level (on ipykernel startup). It is unclear to me if there is any downside to doing that, maybe we should give it a try?

mlucool:
>I previously suggested always appending -Xfrozen_modules=off on a higher level (on ipykernel startup). It is unclear to me if there is any downside to doing that, maybe we should give it a try?

Seems like a better default, but maybe have some way to opt-out in case it does break someone's workflow?

krassowski:
Yes, it looks that we could add this in the default kernelspec and make it configurable by having an argument for kernelspec installation app. If it breaks anything we can quickly revert and switch the default.

We should condition adding this argument on ipykernel being run in CPython rather than any other Python implementation which might not support this argument.

PR #1307, opened 2024-12-13
make debugger class configurable

smacke:
In this PR, add a field debugger_class defaulting to the normal Debugger type. Kernel subclasses can override this field with their own debugger subclass types to provide additional customizations on top of the original Debugger class.

### DIFF
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -23,6 +23,7 @@ from traitlets import Any, Bool, HasTraits, Instance, List, Type, default, obser
 from .comm.comm import BaseComm
 from .comm.manager import CommManager
 from .compiler import XCachingCompiler
+from .debugger import Debugger, _is_debugpy_available
 from .eventloops import _use_appnope
 from .iostream import OutStream
 from .kernelbase import Kernel as KernelBase
@@ -71,6 +72,8 @@ class IPythonKernel(KernelBase):
     shell = Instance("IPython.core.interactiveshell.InteractiveShellABC", allow_none=True)
     shell_class = Type(ZMQInteractiveShell)
 
+    debugger_class = Type(Debugger)
+
     use_experimental_completions = Bool(
         True,
         help="Set this flag to False to deactivate the use of experimental IPython completion APIs.",
@@ -110,11 +113,9 @@ class IPythonKernel(KernelBase):
 
         self.executing_blocking_code_in_main_shell = False
 
-        from .debugger import Debugger, _is_debugpy_available
-
         # Initialize the Debugger
         if _is_debugpy_available:
-            self.debugger = Debugger(
+            self.debugger = self.debugger_class(
                 self.log,
                 self.debugpy_socket,
                 self._publish_debug_event,


ipykernel/ipkernel.py
```diff
@@ -110,11 +113,9 @@ def __init__(self, **kwargs):
 
         self.executing_blocking_code_in_main_shell = False
 
-        from .debugger import Debugger, _is_debugpy_available
```
jasongrout, line 113: It looks like https://github.com/ipython/ipykernel/pull/1223 intentionally introduced this lazy import. Let's make sure we're addressing the need there as well before making this an eager import at the top of the file.

jasongrout: In particular, does this change mean that `python -m ipykernel install --sys-prefix` has spurious warnings?

jasongrout: CC @krassowski who made this lazy import chnage earlier this year.

krassowski: Thanks for spotting, yes this is intentional to address https://github.com/ipython/ipykernel/issues/1198

ipykernel/ipkernel.py
```diff
@@ -71,6 +72,8 @@ class IPythonKernel(KernelBase):
     shell = Instance("IPython.core.interactiveshell.InteractiveShellABC", allow_none=True)
     shell_class = Type(ZMQInteractiveShell)
 
+    debugger_class = Type(Debugger)
```
jasongrout, line 75: You can also give this class as a full-specified dotted string, so we could move the import above back to being an on-demand import in the class definition.

jasongrout: I think this should work:
```suggestion
    debugger_class = Type("ipykernel.debugger.Debugger")
```

smacke: Thanks for the suggestion; I think if I do this I can omit the import altogether?

jasongrout (approved):
Awesome, thanks! In full disclosure, we'd like this at Databricks (where Stephen and I both work) because we customize the debugger class.

davidbrochart (approved):

krassowski (commented):
It seems fine, but not 100% sure as I get an error when trying to install kernel: https://github.com/ipython/ipykernel/issues/1309

Carreau:
Can this get rebased to trigger CI (I try to do it, but there is no allow push from maintainers).

smacke:
Hey @Carreau just rebased + pushed

Carreau:
There is one minor lint to fix (unused mypy ignore), and I restarted the test on 3.9. 

I would actually be in favor of dropping 3.9 support (per https://scientific-python.org/specs/spec-0000/), and FYI IPython CLI already dropped 3.10.

smacke:
@Carreau did you want me to fix those lint failures? They don't seem related with this change but happy to do it anyway

Carreau:
No, it was just a "the failures are know and fixed on master". I haven't been involved in ipykernel in a while and was reluctant to merge, but I think that's ok.

Issue #1289, opened 2024-10-30
Need public opt-out API for output routing from threads


minrk:
We need an easier opt-out of the behavior introduced in #1186, because most output I'm seeing come from threads is really confusing or even lost, where the previous behavior of sending all thread output to the current cell would be more appropriate. Either a kernel-level opt-out or a per-thread opt-out or both should be required, and we should perhaps reconsider the default behavior.

Two examples that I don't think are that unusual, demonstrating where the current behavior is incorrect and currently impossible to avoid with public APIs:

### ipyparallel

For example, IPython Parallel runs its IO in a background thread, and sometimes this produces output (e.g. when streaming output from engines). It produces output in response to direct, blocking action in the main thread, but the output is _produced_ in a long-running background thread. Sending this output to the cell that created the Client is not desirable. Producing the output in the main thread is not particularly feasible either, because it's used in e.g. streaming output in user code:

```python
with async_result.stream_output(): <- this instructs a background thread to start streaming to sys.stdout, etc.
    do_blocking_things() <- blocks the main thread, not controlled by ipyparallel
```

In all situations, the right thing to do for output produced by this thread is to go to the current cell.

The same is true for some log output, e.g. stopping clusters - which again produces output in a background thread as a direct result of synchronized action taken on the main thread, where placing the output near the action taken that triggered the output is less surprising than the far disconnected initial launch of the thread.

### ThreadPoolExecutor

Another, fully standard library situation is ThreadPoolExecutor, where outputs from all tasks will go to the initial thread-spawning Cell. It's not just the cell that creates the pool, since spawning threads may actually be deferred until the first task submission requiring the thread, producing this very surprising output order:

![Image](https://github.com/user-attachments/assets/f90d0606-a2c7-4b41-831e-99a391da6140)

which may be related to the joblib issue in #402 which is closed, but appears to actually be unresolved.

I think there is an assumption in #1186 that threads, once spawned, are long running and do not interact with the main thread. This holds for some examples of "fire and forget" type threads, but is definitely not true in general, and I'm not even sure it's true more often than not.

We at least need a way for packages/libraries to indicate that a thread producing output shouldn't be routed to the originating cell.

[...]

PR #1312, opened 2024-12-20
Fix test_print_to_correct_cell_from_child_thread

davidbrochart:
The parent thread has to live for the duration of the child thread, otherwise the parent thread exits before the child thread and disappears from the active threads, causing it to be removed from `thread_to_parent` and eventually leading to messages not being routed to the right cell:
https://github.com/ipython/ipykernel/blob/41a965eb35b3364a56e5f1e827d7e3a9580c0213/ipykernel/ipkernel.py#L781-L794
@krassowski Could you confirm? I saw that while working on #1291.

### DIFF
--- a/tests/test_kernel.py
+++ b/tests/test_kernel.py
@@ -105,8 +105,10 @@ def test_print_to_correct_cell_from_child_thread():
             sleep({interval})
 
     def parent_target():
-        Thread(target=child_target).start()
-        sleep({interval * iterations})
+        sleep({interval})
+        thread = Thread(target=child_target)
+        thread.start()
+        thread.join()
 
     Thread(target=parent_target).start()
     """
--- a/tests/test_kernel.py
+++ b/tests/test_kernel.py
@@ -105,8 +105,8 @@ def test_print_to_correct_cell_from_child_thread():
             sleep({interval})
 
     def parent_target():
-        sleep({interval})
         Thread(target=child_target).start()
+        sleep({interval * iterations})
 
     Thread(target=parent_target).start()
     """


tests/test_kernel.py
```diff
@@ -105,8 +105,8 @@ def child_target():
             sleep({interval})
 
     def parent_target():
-        sleep({interval})
         Thread(target=child_target).start()
+        sleep({interval * iterations})
```
krassowski, line 109: Does it also work if in addition to this sleep you keep the `sleep({interval})` above? I wonder if that was here for a reason. In other words, I think that adding the sleep at the end is fine, I am less confident about the function of the sleep that was removed (as this can change some edge case related to execution order).

Further, should we save a reference to the thread in a local variable before calling `sleep()`?

minrk:
I think also related to this is #1289 . Most of the time, in my experience, routing output to the cell that happened to spawn the thread _not_ the right thing to do, and results in quite a bit of mysteriously lost and out-of-order output.

krassowski:
Thanks for linking #1289, I missed it when it was opened (despite the mention!). I see your point. I think that usually it is the right thing to do, but will happily agree that there should be a way to opt out for the scenarios that you present.

tests/test_kernel.py
davidbrochart: In https://github.com/ipython/ipykernel/pull/1312/commits/41b31c5ca070bae713d80b15739011143ac58483 I added back the first `sleep`, and replaced the other `sleep` with a `thread.join()`, since it's essentially the same.

davidbrochart:
[Failure](https://github.com/ipython/ipykernel/actions/runs/12432129343/job/34710902367?pr=1312) is unrelated:
```
FAILED tests/test_subshells.py::test_run_concurrently_sequence[True-are_subshells0] - AssertionError: assert 'error' == 'ok'
```

krassowski (approved):
LGTM!

PR #1314, opened 2025-01-21
More Informative assert

Carreau:
This  keep failing downstream in IPython. It's annoying. I hope this will give an idea as to why. Otherwise can we mark the test as xfail ?

If not I"ll stop testing in IPython if I break ipykernel.

### DIFF
--- a/tests/test_subshells.py
+++ b/tests/test_subshells.py
@@ -125,6 +125,9 @@ def test_thread_ids():
         delete_subshell_helper(kc, subshell_id)
 
 
+@pytest.mark.xfail(
+    strict=False, reason="this randomly fail and make downstream testing less useful"
+)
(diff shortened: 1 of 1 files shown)


ianthomas23 (approved):
I am happy to make this change permanently, or the `xfail` alternative if you'd prefer on a temporary basis.

There seem to be quite a lot of upstream and downstream failures in various related projects CI. This test here is somewhat risky as it is timing based and is probably best to relax it a bit, but I don't want to spend any more time on the subshells work here until the `anyio` changes are completed.

Carreau:
xfail, or make them with a specific flag so that they are not ran by default for now ?

Carreau:
I've added an xfail (non strict)

ianthomas23:
Thanks @Carreau.

PR #1316, opened 2025-02-12
Stop support for Python 3.9

Carreau:
New major version are a good time to drop support for older versions. I thihnk 3.9 should not be an issue, I woudl argue for removal of 3.10 as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,10 +22,10 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.9", "3.13"]
+        python-version: ["3.10", "3.13"]
         include:
           - os: ubuntu-latest
-            python-version: "pypy-3.9"
+            python-version: "pypy-3.10"
           - os: macos-latest
             python-version: "3.10"
           - os: ubuntu-latest
--- a/ipykernel/ipkernel.py
+++ b/ipykernel/ipkernel.py
@@ -160,10 +160,7 @@ class IPythonKernel(KernelBase):
         self._new_threads_parent_header = {}
         self._initialize_thread_hooks()
 
-        if hasattr(gc, "callbacks"):
-            # while `gc.callbacks` exists since Python 3.3, pypy does not
-            # implement it even as of 3.9.
-            gc.callbacks.append(self._clean_thread_parent_frames)
+        gc.callbacks.append(self._clean_thread_parent_frames)
 
     help_links = List(
         [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ classifiers = [
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
 ]
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dependencies = [
     "debugpy>=1.8.1",
     "ipython>=7.23.1",


davidbrochart:
Why should we drop Python 3.9, since it has not reached EOL?

Carreau:
Please read the description of the PR:
- We are doing a major version, so it's a good time to drop.

Also:
  - IPython itself has already dropped 3.10
  - https://scientific-python.org/specs/spec-0000/
  - It makes CI faster.
  - We can start using 3.10 only features/types without conditional code.

Waiting for EOL is not a good reason, EOL means that _upsrtream_ have stopped supporting _their_ package, in general it's better to stop before their EOL so that they can stop their support knowing they don't people in the dust. 

Saying you need to wait for EOL, is like saying "no it's fine you can use a deprecated feature, it has not been removed.

Carreau:
Plus explicitly for 3.9, some test are flaky and annoying only on 3.9

davidbrochart:
These are still not valid reasons IMO.
Also, AFAIK we don't do that in other Jupyter projects.

Carreau:
Meh, if you don't want that's fine.

I find it a bit strange as for example main numpy is 3.11+, matplotlib is 3.10+ to lag more than 2 years behind and I have a hard time seing how someone that still python 3.9 can justify that they must have ipykernel 7+. It's not like dropping 3.9 support in 7.0 is still going to prevent 6.x to be installed...

> Also, AFAIK we don't do that in other Jupyter projects.

This is factually false as IPython does it. Unless you do not consider IPython as a Jupyter project ?

I think you are also adding maintenance burden to yourself and potentially pushing some contributors away.

I personally only have so much free time I'm willing to invest to fix bugs on older Pythons, and there are also limited time I can justify to clients WRT python versions they don't use anymore. The "oh I can't do that because we still support 3.9", writing conditional code for older Python is becoming tiresome.

Feel free to close.

davidbrochart:
I'm just saying my opinion, thus not closing.
I'm also no saying IPython is not a Jupyter project, but I don't know any other Jupyter project that does what you suggest.

davidbrochart:
I feel this is something that has to be discussed at the Jupyter level, because one Jupyter project will affect many others, and this is especially true for ipykernel.
My understanding is that we support Python versions until EOL, but I may be wrong.

ianthomas23:
Personally I am +1 on this.

Carreau:
For what it is worth, I just got annoyed by 3.9, because `some_builtin_object|None` is invalid at runtime on 3.9, while we use `|None` everywhere for the rest.

and I really don't see why something released in 2025 need to support something that was release in 2020. There is still the 6.x branch for that.

Carreau:
Can we maybe, just for 7.0 get the thought of @ipython/jupyterkernelteam ?

krassowski:
I still have a mental image of Python 3.9 as a "recent" version, probably because it was a very nicely feature-packed release and subsequent releases were smaller in terms of features (literally, due to shortened development cycle, more on that later).

> I'm also not saying IPython is not a Jupyter project, but I don't know any other Jupyter project that does what you suggest.

There is a precedent for releasing major versions (which we will need to support for a long time afterwards) with a requirement of using a Python version that already met its EOM but not EOL. For example, Python 3.7 was EOL on 2023-06-27 but:
- JupyterLab 4.0 was released on May 15, 2023 requiring Python 3.8 or newer
- jupyter-server 2.0 was released Dec 6, 2022 requiring Python 3.8 or newer

Notably, Python release cadence has increased starting with Python 3.9 from once every 18 months to once every 12 months (https://peps.python.org/pep-0602/). One of the arguments for it was to allow more Linux distributions (case of Fedora is explicitly mentioned in the PEP) to synchronise their release schedule and avoid the shipped Python version from falling behind. I think this makes it easier to drop older versions because nowadays operating systems will ship (or offer) more recent versions.

The downside is that dropping Python 3.9 will put downstreams in a position where they would need to test against multiple ipykernel versions as:
- JupyterLab supports Python 3.9+
- Jupyter Notebook follows what JupyterLab does
- Spyder allows Python 3.8 but tests only on 3.9+ (https://github.com/spyder-ide/spyder/pull/23286)


> IPython itself has already dropped 3.10

I find this a bit of a challenge for downstreams, because moving this fast means:
- frontends (Jupyter, Spyder) will need to move their main to match the Python version OR ignore ipykernel 7
- users will continue using older versions of frontends which support Python 3.10 because they cannot update and they cannot tell their admin to update as these versions are still receiving security updates
- when a new critical security issue gets discovered Jupyter  maintainers will be asked to backport it to Jupyter versions which support 3.10; we may not have resources to do so and users will have a choice of not using Jupyter or using a vulnerable version

which is why I am against removal of Python 3.10 support. I think this is much less of problem for Python 3.9 with the current JupyterLab release cadence.

CC @ccordoba12 for opinions.

Carreau:
> The downside is that dropping Python 3.9 will put downstreams in a position where they would need to test against multiple ipykernel versions as: [...]

I'm confused of what the extra changes are here.
This is already the case that downstream are testing with multiple version of  dependencies that are not compatible with some Python version in the CI matrix? That the point of CI matrix and `python_requires` ? 


In CI the python matrix will already install IPython 8.18.1 on 3.9, but 8.x (8.32 currently) on 3.10+ and latest on 3.11+ (which should) soon be 9.0). And that will be true for any direct or indirect (or conditional) dependencies as well.

I'm genuinely confused on what it changes downstream, it's like literally the life of CI that versions of dependencies changes across CI matrix and why ipykernel would need to be different ?

Do you mean that downstream _may_ need to have conditional code based on IPykernel version ? 
I agree with that, but that's unrelated to IPykernel being 3.8+, 3.10+ or 3.11+, it just depends on whether we have API change in ipykernel.

> frontends (Jupyter, Spyder) will need to move their main to match the Python version OR ignore ipykernel 7


I'm confused as well here. I don't see why users can't use newer version of frontends if they can't update ipykernel?  Don't/won't jupyterlab, server and all still work with ipykernel 6 ? 
Isn't the point of Jupyter to decouple frontend and kernels ?  Did they had to had to "move their main to match the Python " when IPython dropped support for 3.9 ? For 3.10 ? I did not see anyone stop support for IPython 8.19+, so I'm genuinely confused of what you are implying downstream will need to do.

>    users will continue using older versions of frontends which support Python 3.10 because they cannot update and they cannot tell their admin to update as these versions are still receiving security updates

How is dropping support for something in ipykernel affects security issue in other projects ? 
 - If it's an issue in IPykernel than it's ipykernel problem to backport security patches to say the 6.x branch. 
 - If Jupyter Server is still compatible 3.10, or even 3.9 and has a security issue it is not affected by ipykernel. It can use ipykernel 6 on 3.9 and ipykernel 7 on 3.12.

I mean we are not speaking of jupyter_client here – which I agree would have a much bigger impact.

All of the above seem to be all the same point that were made for the "Drop python 2" transition; and the reason `requires_python` was added:

- Publishing IPykernel 7.x with `python_requires = '>=3.whathever'` will just mean that on python 3.whatever and below pip and alike will just install 6.x, it does not force anyone downstream to make any changes, and CI will automatically adapt. And because it's a kernel it should not impact any clients.

---- 
>    when a new critical security issue gets discovered Jupyter maintainers will be asked to backport it to Jupyter versions which support 3.10; we may not have resources to do so and users will have a choice of not using Jupyter or using a vulnerable version

Those are different arguments, and those are imho opinions, I can understand those but:

I find the "weed won't have enough resources" also a double edge swords and also a bit of a straw man argument:

 1) You don't **have to** match CPython security timeline, that's a choice, you can decide to be shorter, other python projects are shorter. And again here I think that kernel and security are one of the most flexible and with the least attack surface compared to say server/lab or hub.

 2) The more python version you support the more work it is. In my own experience the time I spend fixing CI or writing code that is compatible 3.9 and 3.10 is wayyyy more that the time I spend backporting and working on security fixes. So I'll return it the opposite argument: dropping 3.9 – and 3.10 – will free much more time from maintainers to be able to actually spend time backporting security fixes if needed.

krassowski:
Yes, the misunderstanding and my poor points come from an implicit assumption that jupyterlab and other frontends will soon require the latest version of ipykernel. Of course, this does not need to be the case. Maybe my thinking here was biased by the failing downstream tests and this release being not backward compatible.

In two points:
- yes, JupyterLab may need to expand CI matrix because testing subshells integration (currently done by installing alpha) will no longer be possible on older Python version; not a huge deal on one hand, pretty annoying on the other
- yes, JupyterLab could decide not to upgrade its pin of `ipykernel` to `ipykernel>=7.0` (currently `ipykernel>=6.5`). Supporting the smallest range of `ipykernel` makes life easier for JupyterLab.

I agree someone needs to pay some extra maintenance cost and the decision to be made will just shift it between maintainers of ipykernel/users/maintainers of downstream. Some choices might allow to minimize the total cost.

> You don't have to match CPython security timeline, that's a choice, you can decide to be shorter, other python projects are shorter. And again here I think that kernel and security are one of the most flexible and with the least attack surface compared to say server/lab

Of course! I just explained what I thought was likely to happen, in a sense worse case scenario. It does not mean that this is a strong argument, or that I personally think this PR should not be merged, but I find value in exploring these scenarios to understand trade-offs. Spelling it out means others, like you, can correct errors in my reasoning.

Carreau:
> Yes, the misunderstanding and my poor points come from an implicit assumption that jupyterlab and other frontends will soon require the latest version of ipykernel. Of course, this does not need to be the case. Maybe my thinking here was biased by the failing downstream tests and this release being not backward compatible.

Ah with this assumption I understand, and if JupyterLab wants to pin ipykernel, then I agree with your points.

> * yes, JupyterLab may need to expand CI matrix because testing subshells integration (currently done by installing alpha) will no longer be possible on older Python version; not a huge deal on one hand, pretty annoying on the other

It seem that JupyterLab _should_ anyway test kernels that do and do not support subshells, it's understandable that it's more work. 

Now this goes into a slight different direction of JupyterLab and Kernel vertical integration. 
Jupyter Frontends tries to be kernel and language agnostic, but is it time to start thinking about tighter integration and make the assumption that the kernel is almost always a Python one.

Like if jupyterlab pins ipykernel, then I also expect older jupyterlab to not support ipykernel>7; in which case this is a vertical integration and we should likely merge back ipykernel and the server code into the same, or have releases in lockstep.


> Of course! I just explained what I thought was likely to happen, in a sense worse case scenario. It does not mean that this is a strong argument, or that I personally think this PR should not be merged, but I find value in exploring these scenarios to understand trade-offs. Spelling it out means others, like you, can correct errors in my reasoning.

I understand, and I think 3.10 is on the fence, so I won't push for it.

PR #1317, opened 2025-02-12
Copy payloadpage.page from IPython

### LINKED ISSUE #1304: Move IPython.core.payloadpage to this package ?
```
ipykernel/ipykernel/zmqshell.py
 23:from IPython.core import page, payloadpage
547:        self.set_hook("show_in_pager", page.as_hook(payloadpage.page), 99)
```

Seem to be the only thing that refer to IPython.core.payloadpage submodule, even in IPython itself; 
so I think it should be moved here.

Carreau:
IPykernel is the only user, and this will let us remove it from IPython.

Closes #1304

### DIFF
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -20,7 +20,7 @@ import threading
 import warnings
 from pathlib import Path
 
-from IPython.core import page, payloadpage
+from IPython.core import page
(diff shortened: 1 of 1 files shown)


minrk (approved):

Carreau:
THanks !

PR #1325, opened 2025-02-13
Try to force precommit-ci to send autoupdate PRs.

Carreau:
It has not in a loong time, but I'm not sure why.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,7 @@
 ci:
-  autoupdate_schedule: monthly
   autoupdate_commit_msg: "chore: update pre-commit hooks"
+  autoupdate_branch: ""
+  autoupdate_schedule: weekly
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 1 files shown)


minrk:
Looking at branch protection rules for #1318, I think the rules were prohibiting pre-commit.ci from pushing to its branch. I think I fixed it by excluding `pre-commit-ci-update-config` and dependabot brnaches, but hard to say how to tell, since there's no way to force a PR.

Carreau:
Then I suggest we at least merge this to have an update next monday instead of waiting a month.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,7 @@
 ci:
-  autoupdate_schedule: monthly
   autoupdate_commit_msg: "chore: update pre-commit hooks"
+  autoupdate_branch: ""
```
minrk, line 3: I don't think setting this to empty makes sense, since it's already the default

```suggestion
```

Carreau:
thanks !

PR #1327, opened 2025-02-13
Manually update mdformat pre-commit and run it.

Carreau:
I think some if those changes are wrong, as `[foo]` is valid markdown when the link is defined later, but here we actually want square brackets.


Carreau:
thanks.

PR #1333, opened 2025-02-13
Bump mypy

Carreau:


Carreau:
Everything (except downstream as usual) is passing.

ipykernel/zmqshell.py
```diff
@@ -78,12 +78,16 @@ def _hooks(self):
             self._thread_local.hooks = []
         return self._thread_local.hooks
 
-    def publish(
+    # Feb: 2025IPython has a deprecated, `source` parameter, marked for removal that
```
krassowski, line 81:
```suggestion
    # Feb: 2025 IPython has a deprecated, `source` parameter, marked for removal that
```

ipykernel/inprocess/client.py
```diff
@@ -114,7 +126,9 @@ async def execute(
         )
         msg = self.session.msg("execute_request", content)
         await self._dispatch_to_kernel(msg)
-        return msg["header"]["msg_id"]
+        res = msg["header"]["msg_id"]
+        assert isinstance(res, str)
```
krassowski, line 130: I wondered if `typing.cast` could be safer, but per protocol it looks like a valid assertion

krassowski (approved):

ipykernel/inprocess/client.py
Carreau: I agree, my hope is that if it's wrong we can at least trigger failures with nightlies/pre-release, whcih a cat would not

Carreau (commented):
Thanks  for the review.

Carreau:
docs failure unrelated.

PR #1337, opened 2025-02-17
try to fix spyder kernel install

Carreau:


Carreau:
That fixes the install step so we can see actual failure. 

I'll self merge soon

PR #1339, opened 2025-02-17
Test changing base method to async after #1295

Carreau:
This is not meant to be released in 7.0 finale but iss used as a test that:
 - 1. Our test suite is working is eferything is async
 - 2. I'like to think of making a canary release (alpha, or beta) with something similar to get feedback from what we break upstream.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -794,7 +794,7 @@ class Kernel(SingletonConfigurable):
             self._aborted_time = time.monotonic()
             self.log.info("Aborting queue")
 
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -116,6 +116,7 @@ def destroy(self):
     @no_type_check
     async def test_shell_message(self, *args, **kwargs):
         msg_list = self._prep_msg(*args, **kwargs)
+        print(msg_list)
```
Carreau, line 119:
```suggestion
```

PR #1345, opened 2025-02-18
Try to debug non-closed iopub socket

Carreau:
We sometime get

```
 FAILED tests/test_connect.py::test_port_bind_failure_gives_up_retries - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <function Socket.__del__ at 0x7efe22e4d760>

    File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/zmq/sugar/socket.py", line 185, in __del__
      warn(
  ResourceWarning: Unclosed socket <zmq.asyncio.Socket(zmq.PUB) at 0x7efe1eb32820><F48>
```

This adds a tracemalloc fixture for the crashing test module in order to try to narrow that down. I'm thorn between restart the tests here until we have the failure; or merge and see once we get the error in another PR and then remove the fixture once we've fund the culprit.

--- 

The only zmq.PUB I can find is iopub, and we don't seem to be closing it

In addition : should app be a contextmanager ?
And should the atexit close registring be replaced with a weakref finalizer ?

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -427,7 +427,7 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp, ConnectionFileMix
         if self.debug_shell_socket and not self.debug_shell_socket.closed:
             self.debug_shell_socket.close()
 
-        for channel in ("shell", "control", "stdin", "iopub"):
+        for channel in ("shell", "control", "stdin"):
             self.log.debug("Closing %s channel", channel)
             socket = getattr(self, channel + "_socket", None)
             if socket and not socket.closed:
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -427,7 +427,7 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp, ConnectionFileMix
         if self.debug_shell_socket and not self.debug_shell_socket.closed:
             self.debug_shell_socket.close()
 
-        for channel in ("shell", "control", "stdin"):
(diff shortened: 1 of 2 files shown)


Carreau:
Ah ! No, in the real application the iopub_socket is closed when you close the thread, but in conftest it is not !

Carreau:
well, in conf.py it is also closed...

Carreau:
@minrk do you believe that using tracemalloc in the __del__ in pyzmq to print where the socket was created could help and would you accept a PR ?

minrk:
@Carreau 100%! I thought you had to `-X tracemalloc` / `PYTHONTRACEMALLOC=1` to get this output from ResourceWarnings. Can we run the tests with that?

Carreau:
> @Carreau 100%! I thought you had to `-X tracemalloc` / `PYTHONTRACEMALLOC=1` to get this output from ResourceWarnings. Can we run the tests with that?

I think you need both ? My understanding is that `-X tracemalloc` / `PYTHONTRACEMALLOC=1` just enable tracemalloc, but that you still need to call [tracemalloc.get_object_traceback(obj)](https://docs.python.org/3/library/tracemalloc.html#tracemalloc.get_object_traceback) in the code that emit the warnings, and `tracemalloc.get_object_traceback(obj)` ?

Maybe I'm wrong.

Carreau:
Seem that [you are right](https://github.com/python/cpython/blob/388e1ca9f08ee5caefd1dd946dc6e236ce73d46f/Lib/warnings.py#L35-L92) it shoudl show up automatically. I try differently.

PR #1346, opened 2025-02-19
Make our own mock kernel methods async

Carreau:
As we are recommending upstream to use asyn, mught as well do it ourselves a well.

Plus this makes sure our test suite tests async


minrk (approved):

krassowski:
I think there is a typo in the title, should be `Make our own mock kernel methods async`

Carreau:
> I think there is a typo in the title, should be `Make our own mock kernel methods async`

Oops ! Fixed.

PR #1348, opened 2025-02-19
Licence :: * trove classifers are deprecated

Carreau:
It is replaced by License-expression (which is exclusive with license that we used, so can't just replace for now).

https://packaging.python.org/en/latest/specifications/core-metadata/#license-expression


minrk:
> which is exclusive with license that we used, so can't just replace for now

I don't follow this. Shouldn't we use `License-Expression: BSD-3-Clause`?

PR #1349, opened 2025-02-19
Disable 3 failing downstream tests, but keep testing the rest.

Carreau:
Also remove the -x, because if there is many failures, we want to knw that.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -116,7 +116,7 @@ jobs:
         shell: bash -l {0}
         run: |
           cd ${GITHUB_WORKSPACE}/../qtconsole
-          xvfb-run --auto-servernum ${pythonLocation}/bin/python -m pytest -x -vv -s --full-trace --color=yes -k "not test_execute" qtconsole
+          xvfb-run --auto-servernum ${pythonLocation}/bin/python -m pytest -vv -s --full-trace --color=yes -k "not test_execute" qtconsole
 
   spyder_kernels:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


ccordoba12:
Just a quick comment to say that I don't agree with this change because it'd basically leave this in the cold to support IPykernel 7.0, which seems to include some big changes (like the usage of AnyIO) for which we have zero experience with.

Furthermore, last time I checked there are not many failing tests in both Qtconsole and Spyder-kernels that need to be fixed.

davidbrochart:
> which seems to include some big changes (like the usage of AnyIO) for which we have zero experience with

The public API is just async, you don't need experience with AnyIO.

Carreau:
This disable _just_ the failing tests that have been failing for months:
 - only 2 tests in spyder_kernel
 - and 1 test in qtconsole, not the full test suite)

Currently every PRs here need an admin to overwrite merge because of only these 3 failures, so it does not change problem, it will just make it easier to spot if there are any other regressions.

Carreau:
In addition, if you look, I removed the `-x`, which meant the test were stopping after the first failure, so if anything this PRs tests qtconsole and spyder_kernel *more* than current main

ccordoba12:
> The public API is just async, you don't need experience with AnyIO.

We have little experience with async stuff because it's not easy to integrate it with Qt and it doesn't bring a significant advantage for us due to the GIL.

> This disable just the failing tests that have been failing for months

Ok, sorry for the noise then. I thought you were going to disable the entire test suite.

> only 2 tests in spyder_kernel
> and 1 test in qtconsole

Ok, at least in spyder-kernels we have an upper constraint on IPykernel 7, so we can take our time to address those failures. But that's not the case in Qtconsole, and the test marked as failing checks the in-process kernel, which is used by several projects downstream.

So hopefully that can be addressed before releasing the new version.

Carreau:
No worries, I've rephrased the PR title.

Let me merge this and open an issue marked 7.0 to remind us to un-skip those tests.

PR #1350, opened 2025-02-19
Try to reenable tests from downstream ipywidgets

Carreau:
Have been disabled from 2 years.


Carreau:
well, ipywidgets is passing now.

PR #1352, opened 2025-02-20
Remove deprecated modules since 4.3 (2016).

Carreau:
This is annoying as this makes downstream tests fail in IPython, while the culprit is ipyparallel.

Ipyparallel itself is tested, and main has been fixed, so we just get a false failure in IPython downstream tests.

It's also been deprecated for 10 years, now, so at least we can have that removed from main; so that we can see if downstream breaks.


davidbrochart (approved):

krassowski (commented):
Yup, if this break a lot of downstreams it is easy to revert as this just removed whole files.

Should we increase

https://github.com/ipython/ipykernel/blob/dd07f5272eaf3c613a284421989985256a18e4ab/.github/workflows/ci.yml#L80

Since we are removing some files which did not have test coverage?

krassowski:
Huh, it looks like coverage reporting is broken. I opened https://github.com/ipython/ipykernel/issues/1365, I guess this is good to merge then.

PR #1358, opened 2025-02-25
Test more python versions

davidbrochart:


Carreau:
I think you might try `3.14-dev`, also, but that may be pushing it.

PR #1361, opened 2025-02-26
Ensure IPython's HistoryManager threads don't leak

davidbrochart:


tests/conftest.py
```diff
@@ -34,6 +35,9 @@ def _garbage_collection(request):
 except ModuleNotFoundError:
     tracemalloc = None
 
+# ensure we don't leak history managers
+HistoryManager._max_inst = 1
```
Carreau:
```suggestion
HistoryManager._max_inst = 2
```

Let's give it some wiggle room

PR #1362, opened 2025-02-26
Test test_kernelapp.py only

davidbrochart:


davidbrochart:
This is to show that `tests/test_kernelapp.py::test_start_app` hangs if run separately, while it passes when the whole test suite is run:
https://github.com/ipython/ipykernel/actions/runs/13539944292/job/37838614614?pr=1362

davidbrochart:
Fixed in #1366.

PR #1363, opened 2025-02-26
TQDM workaround due to unresponsive maintainer

Carreau:
`tqdm` has an incorrect detection of ZMQInteractiveShell when launch via a scheduler that bypass IPKernelApp. Think of JupyterHub cluster spawners and co.

As of end of Feb 2025, the maintainer has been unresponsive for 5 months, to our fix, so we implement a workaround. I don't like it but we have few other choices.

See https://github.com/tqdm/tqdm/pull/1628

### DIFF
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -34,7 +34,6 @@ from IPython.utils.process import arg_split, system  # type:ignore[attr-defined]
 from jupyter_client.session import Session, extract_header
 from jupyter_core.paths import jupyter_runtime_dir
 from traitlets import Any, CBool, CBytes, Dict, Instance, Type, default, observe
-from traitlets.config import Config
 
 from ipykernel import connect_qtconsole, get_connection_file, get_connection_info
 from ipykernel.displayhook import ZMQShellDisplayHook
--- a/ipykernel/zmqshell.py
+++ b/ipykernel/zmqshell.py
@@ -34,6 +34,7 @@ from IPython.utils.process import arg_split, system  # type:ignore[attr-defined]
 from jupyter_client.session import Session, extract_header
 from jupyter_core.paths import jupyter_runtime_dir
 from traitlets import Any, CBool, CBytes, Dict, Instance, Type, default, observe
+from traitlets.config import Config
 
 from ipykernel import connect_qtconsole, get_connection_file, get_connection_info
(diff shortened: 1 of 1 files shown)


krassowski (approved):
I can see that even once the maintainer of `tqdm` has some more time, they can take a while to respone as there is a a few upstream PRs related to improving notebook autodetection:
- https://github.com/tqdm/tqdm/pull/1628 (ours)
- https://github.com/tqdm/tqdm/pull/1655
- https://github.com/tqdm/tqdm/pull/1559

Not sure if we can provide any feedback on the other PRs, but if there are any thoughts we could share to help with review upstream, this might help to move it forward.

ipykernel/zmqshell.py
```diff
@@ -472,6 +473,20 @@ def subshell(self, arg_s):
...
+        # tqdm has an incorrect detection of ZMQInteractiveShell when launch via
+        # a scheduler that bypass IPKernelApp Think of JupyterHub cluster
+        # spawners and co. as of end of Feb 2025, the maintainer has been
+        # unresponsive for 5 months, to our fix, so we implement a workaround. I
+        # don't like it but we have few other choices.
+        # See https://github.com/tqdm/tqdm/pull/1628
+        if "IPKernelApp" not in self.config:
+            self.config["IPKernelApp"] = Config(
+                {"tqdm": "dummy value for https://github.com/tqdm/tqdm/pull/1628"}
+            )
```
minrk, line 488: slightly simpler:

```suggestion
            self.config.IPKernelApp.tqdm = "dummy value for https://github.com/tqdm/tqdm/pull/1628"
```

Carreau:
To backport.

ipykernel/zmqshell.py
Carreau: Yep. Make sens.

ipykernel/zmqshell.py
```diff
@@ -34,6 +34,7 @@
 from jupyter_client.session import Session, extract_header
 from jupyter_core.paths import jupyter_runtime_dir
 from traitlets import Any, CBool, CBytes, Dict, Instance, Type, default, observe
+from traitlets.config import Config
```
Carreau, line 37:
```suggestion
```

davidbrochart (approved):

Carreau:
I hope this will get merged one day.

PR #1364, opened 2025-02-26
Ensure test_start_app takes 1s to stop kernel

davidbrochart:


davidbrochart:
The test fails, meaning that it shouldn't pass on `main` (see #1362).

davidbrochart:
This just makes the test useful. Let's merge it and fix the test?

PR #1366, opened 2025-02-27
Make kernelbase._eventloop_set event thread-safe

davidbrochart:
See #1364.


davidbrochart:
I can see locally that `test_kernelapp.py::test_start_app` passes with this PR if run independently, but it fails if the whole test suit is run.

PR #1367, opened 2025-02-27
Fix OutStream hooks not being called

steovd:
#1110 introduced `OutStream` hooks, which are [stored in a thread-local variable](https://github.com/ipython/ipykernel/blob/700c26f29126d2ced5480f0314642b12cff0d493/ipykernel/iostream.py#L733) and [invoked in `_flush`](https://github.com/ipython/ipykernel/blob/700c26f29126d2ced5480f0314642b12cff0d493/ipykernel/iostream.py#L651). However, since `_flush` [usually runs in a background thread](https://github.com/ipython/ipykernel/blob/700c26f29126d2ced5480f0314642b12cff0d493/ipykernel/iostream.py#L611), the hooks only work when registered from this thread.

This PR makes it so that the hooks are accessed from the thread which calls `flush`, which I assume is the expected behavior.

### DIFF
--- a/ipykernel/iostream.py
+++ b/ipykernel/iostream.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import atexit
 import contextvars
+import functools
 import io
 import os
 import sys
@@ -620,10 +621,19 @@ class OutStream(TextIOBase):
         else:
             self._flush()
 
+    @property
     def _flush(self):
+        """Prepare _flush_impl partial to be scheduled on the IO thread.
+
+        This indirection is necessary to ensure _flush_impl calls hooks
+        registered from the current thread (as they are thread-local).
+        """
+        return functools.partial(self._flush_impl, self._hooks)
+
+    def _flush_impl(self, hooks=()):
         """This is where the actual send happens.
 
-        _flush should generally be called in the IO thread,
+        _flush_impl should generally be called in the IO thread,
         unless the thread has been destroyed (e.g. forked subprocess).
         """
         self._flush_pending = False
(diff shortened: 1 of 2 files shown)


davidbrochart (commented):
Thanks @steovd, do you think this can be related to https://github.com/ipython/ipykernel/pull/1369#issuecomment-2687591429?

steovd:
> Thanks @steovd, do you think this can be related to [#1369 (comment)](https://github.com/ipython/ipykernel/pull/1369#issuecomment-2687591429)?

I don't think either this patch or #1110 have anything to do with it. The same happens on main even when I revert [the only possibly relevant change from #1110](https://github.com/ipython/ipykernel/pull/1110/files#diff-50efd52d050465d24b5b923d8b8ee556f91142cf44a9209244db2a4a75dd7f6eL530-R542). I'm not that familiar with the code though, so I could be wrong.

nathanmcavoy:
Just checking in - any update on this? This fix would resolve a frustrating bug I'm currently facing. Thanks!

ipykernel/iostream.py
```diff
@@ -620,10 +621,19 @@ def flush(self):
         else:
             self._flush()
 
+    @property
```
Zentrik, line 624: took me a hot second to figure out why this was being used.

steovd:
> Just checking in - any update on this? This fix would resolve a frustrating bug I'm currently facing. Thanks!

You can work around this bug by registering your hooks from the background thread. I don't have the code at hand but I think it went something like this:
```py
outstream.pub_thread.schedule(functools.partial(outstream.register_hook, myhook))
```

nathanmcavoy:
> > Just checking in - any update on this? This fix would resolve a frustrating bug I'm currently facing. Thanks!
> 
> You can work around this bug by registering your hooks from the background thread. I don't have the code at hand but I think it went something like this:
> 
> ```python
> outstream.pub_thread.schedule(functools.partial(outstream.register_hook, myhook))
> ```

Thanks! I ended up doing something similar, but looks like your version is neater.

Adding the hook itself is already a hacky fix to what I am trying to do so I still think this PR is useful, but glad there is a workaround.

PR #1369, opened 2025-02-27
Fix test_outstream

davidbrochart:


davidbrochart:
[test_outstream](https://github.com/ipython/ipykernel/blob/bf104470d2efe50437523697b736ac66b2c2fc16/tests/test_io.py#L95) passes but it shouldn't, here is what is printed if [`pytest -s` is used](https://github.com/ipython/ipykernel/actions/runs/13564635258/job/37914872716?pr=1369):
```
  File "/home/runner/work/ipykernel/ipykernel/ipykernel/iostream.py", line 392, in _watch_pipe_fd
    bts = os.read(self._fid, PIPE_BUFFER_SIZE)
AttributeError: 'OutStream' object has no attribute '_fid'
```

PR #1414, opened 2025-07-30
Cache separate headers on subshell threads

ianthomas23:
This fixes a bug in the caching of received message headers at various places in the code which are later used for reply or iopub messages. The existing code on `main` branch assumes that the shell processes a single request at a time so that the cached header is valid until that request is replied to, but now of course there can be multiple subshells in different threads executing code concurrently.

The solution is to cache a different header per subshell, or to be more precise a different header per thread. This is accomplished using a [ContextVar](https://docs.python.org/3/library/contextvars.html) which is essentially thread-local storage that also supports asynchronous execution.  We are already using a `ContextVar` at https://github.com/ipython/ipykernel/blob/4c218005f790057f44a2af0be29d33018572490a/ipykernel/iostream.py#L454
so it is already a dependency of `ipykernel`.

I found this whilst debugging use of `ipympl` in `jupyterlab`, trying to create multiple `matplotlib` plots using the option of one subshell per comm-target (the default option). The best way to view this bug is to use [kernelspy](https://github.com/jupyterlab-contrib/jupyterlab-kernelspy) as it shows that in this scenerio some messages (e.g. some idle status messages) have the wrong parent header and are therefore displayed in the wrong section of the `kernelspy` output.

All new `parent_header` attributes are private (leading underscore). Where they were previously public I have added a public `parent_header` getter to extract the appropriate value from the `ContextVar` so that downstream libraries that have been reading the `parent_header` in the main shelll should have the same functionality as before.

### DIFF
--- a/ipykernel/displayhook.py
+++ b/ipykernel/displayhook.py
@@ -7,10 +7,11 @@ from __future__ import annotations
 import builtins
 import sys
 import typing as t
+from contextvars import ContextVar
 
 from IPython.core.displayhook import DisplayHook
 from jupyter_client.session import Session, extract_header
-from traitlets import Any, Dict, Instance
+from traitlets import Any, Instance
 
 from ipykernel.jsonutil import encode_images, json_clean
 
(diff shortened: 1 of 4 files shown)


ianthomas23:
I'm merging this as is, with the downstream `jupyter_client` and `ipyparallel` workflows disabled as they are waiting on PRs in those repositories. I'll create an issue to track the downstream changes.

ianthomas23:
Note `Kernel._parent_ident` reinstated in PR #1427 for backward compatibility.

minrk:
Looks like there's more to it - we need to resolve parent_header to the main shell thread from _any_ thread that could be spawned (#1450).

I think this actually relates to #1289 because we do a lot wrong with output from threads right now.

PR #1419, opened 2025-08-06
Use correct `__version__` on `main` branch after branch manipulations

ianthomas23:
After the branch manipulations a few weeks ago, we need to update the `__version__` to be the latest in the `7.x` series not `6.x` series.

Fixes ipython/ipython#14953.

### DIFF
--- a/ipykernel/_version.py
+++ b/ipykernel/_version.py
@@ -5,7 +5,7 @@ store the current version info of the server.
 import re
 
 # Version string must appear intact for hatch versioning
-__version__ = "6.30.0a0"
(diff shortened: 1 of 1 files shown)


ianthomas23:
Downstream failure in `jupyter_client` is due to use if `anyio` that needs to be removed, see jupyter/jupyter_client#1066.

Merging this so that the `ipython`'s downstream `ipykernel` tests should pass, the downstream `jupyter_client` tests here will fail until I have submitted a PR to fix the issue and it is merged there.

PR #1431, opened 2025-09-16
Drop support for Python 3.9

ianthomas23:
Drop support for Python 3.9 on `main` branch. We would normally continue to support a Python version until its end of life, which is next month, but the next release from `main` will be 7.0.0 and I don't intend to make bug fix releases on the 7.x branch for Python 3.9 so it is cleaner to not do a release in the first place. Note that `ipython` has already dropped support for Python 3.9 and 3.10.

The `6.x` branch will continue to support Python 3.9 if any further bug fix releases are required on that branch.


ianthomas23:
Thanks @minrk

PR #1434, opened 2025-09-20
Manually reset interrupt evemt

ptosco:
Manually reset interrupt event at the receiver end rather than relying on auto reset. This makes it possible for other waiting receivers to be signaled.
In particular, C Python modules running on Windows on a Jupyter Kernel can be interrupted via "Interrupt kernel" without requiring to send the "Interrupt kernel" signal twice as currently happens (see more detailed description in the PR to the `jupyter_client repo).

### DIFF
--- a/ipykernel/parentpoller.py
+++ b/ipykernel/parentpoller.py
@@ -121,6 +121,7 @@ class ParentPollerWindows(Thread):
 
             if WAIT_OBJECT_0 <= result < len(handles):
                 handle = handles[result - WAIT_OBJECT_0]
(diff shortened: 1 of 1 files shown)


minrk:
@ptosco can you confirm my understanding that this is safe and right to do regardless of the parent process behavior (i.e. without the jupyter-client PR)?

ptosco:
@minrk Let me double-check that; I’ll get back to you by tomorrow. Thanks.

ptosco:
@minrk I can confirm this is safe regardless of the parent process behavior (i.e. without the jupyter-client PR).

minrk:
@ptosco thank you!

PR #1437, opened 2025-10-08
start to test on 3.14 and 3.14-free-threaded

Carreau:


ianthomas23:
Closing this as the changes here were included in #1447 so that fix could be tested against python 3.14 in CI.

PR #1446, opened 2025-10-14
6.x: Update PEP-639 license values in pyproject.toml

bollwyvl:
Thanks all!

While wading through the various CI PRs, saw errors relating to this.

I didn't fully investigate _when_ `hatchling` would properly, fully, correctly support the various fields, so I updated to what is [claimed upstream](https://hatch.pypa.io/dev/history/hatchling/#hatchling-v1.5.0) as a minimum. As a `build-system`, it's likely to have been tested with just about every version that has come out.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling>=1.12", "jupyter_client>=6"]
+requires = ["hatchling>=1.22", "jupyter_client>=6"]
 build-backend = "hatchling.build"
 
 [project]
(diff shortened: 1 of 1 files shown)


bollwyvl:
Oof, upstream on `hatch` it actually was updated all the way to [latest](https://github.com/pypa/hatch/pull/2064/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711). Guess i'll try that.

bollwyvl:
I'll do a forward-port in a jiffy...

PR #1447, opened 2025-10-14
Avoid overriding Thread._context in Python 3.14

### LINKED ISSUE #1445: [BUG] In version 7.0.0, on executing a cell, the kernel status is stuck at connecting indefinitely.
On installing ipykernel version 7.0.0, executing a cell causes the kernel status to be stuck at connecting indefinitely. No output is produced.

Image

On installing ipykernel==6.30.1, it works fine.

My python version is 3.14.0
OS: MacOS

ianthomas23:
Cannot set `threading.Thread._context` in Python 3.14 as it is used for a new `context` argument passed to the constructor. Hence here changing use of `ShellChannelThread._context` to `ShellChannelThread._zmq_context` instead.

Fixes #1445.


ianthomas23:
This is mostly working now. The failures look the same as on the `6.x` branch for python 3.14. They are real issues that need fixing, but I'd like to release 7.0.1 as soon as possible with this and #1448 to avoid annoying python 3.14 users any more.

ianthomas23:
Evidence that this works locally for me using python 3.14:

Screenshot 2025-10-14 at 16 22 58

Issue #1442, opened 2025-10-14
bugreport: event loop integration not working for qt (pyside6)


RubendeBruin:
The integration of the Qt event loop with notebooks seems to be broken. Anything I can do to fix it? 

<img width="585" height="475" alt="Image" src="https://github.com/user-attachments/assets/5dc3167c-7936-4e43-ad81-8ebd20016fdf" />

Reproduce:

[qt_eventloop_integration.ipynb](https://github.com/user-attachments/files/22898805/qt_eventloop_integration.ipynb)

I used uv to set up the environment using the following toml:
```
[project]
name = "testerdetest"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "jupyterlab>=4.4.9",
    "pyside6-essentials",
]
```

packages:
```
Package                   Version
------------------------- --------------
anyio                     4.11.0
argon2-cffi               25.1.0
argon2-cffi-bindings      25.1.0
arrow                     1.3.0
asttokens                 3.0.0
async-lru                 2.0.5
attrs                     25.4.0
babel                     2.17.0
beautifulsoup4            4.14.2
bleach                    6.2.0
certifi                   2025.10.5
cffi                      2.0.0
charset-normalizer        3.4.3
colorama                  0.4.6
comm                      0.2.3
debugpy                   1.8.17
decorator                 5.2.1
defusedxml                0.7.1
executing                 2.2.1
fastjsonschema            2.21.2
fqdn                      1.5.1
h11                       0.16.0
httpcore                  1.0.9
httpx                     0.28.1
idna                      3.11
ipykernel                 7.0.0
ipython                   9.6.0
ipython-pygments-lexers   1.1.1
isoduration               20.11.0
jedi                      0.19.2
jinja2                    3.1.6
json5                     0.12.1
jsonpointer               3.0.0
jsonschema                4.25.1
jsonschema-specifications 2025.9.1
jupyter-client            8.6.3
jupyter-core              5.8.1
jupyter-events            0.12.0
jupyter-lsp               2.3.0
jupyter-server            2.17.0
jupyter-server-terminals  0.5.3
jupyterlab                4.4.9
jupyterlab-pygments       0.3.0
jupyterlab-server         2.27.3
lark                      1.3.0
markupsafe                3.0.3
matplotlib-inline         0.1.7
mistune                   3.1.4
nbclient                  0.10.2
nbconvert                 7.16.6
[...]

PR #1451, opened 2025-10-14
Fix routing of background thread output when no parent is set explicitly

### LINKED ISSUE #1289: Need public opt-out API for output routing from threads

### LINKED ISSUE #1450: `ipykernel` 7.0.0 cannot send display data from threads
Minimal python code to put in notebook:
```py
import logging

import ipyparallel as ipp


def hello_mpi():
    # We define all imports inside the function as they have to be launched on the remote engines
    from mpi4py import MPI

    print(f"Hello from rank {MPI.COMM_WORLD.rank}/{MPI.COMM_WORLD.size - 1}")


with ipp.Cluster(engines="mpi", n=3, log_level=logging.ERROR) as cluster:
    # We send the query to run the function `hello_mpi` on all engines
    query = cluster[:].apply_async(hello_mpi)
    # We wait for all engines to finish
    query.wait()
    # We check that all engines exited successfully
    assert query.successful(), query.error
    # We print the output from each engine
    print("".join(query.stdout))
```
Using conda to install the packages:
```yml
name: test_ipp
channels:
  - conda-forge
dependencies:
  - mpi4py
  - ipyparallel
  - python=3.12
  - ipykernel>=7.0.0
  - jupyter-book
  - ipywidgets
  - jupytext
```
results in
```bash
ERROR:tornado.general:Uncaught exception in ZMQStream callback
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/tqdm/notebook.py", line 262, in update
    return super().update(n=n)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/tqdm/std.py", line 1242, in update
    self.refresh(lock_args=self.lock_args)
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/tqdm/std.py", line 1347, in refresh
    self.display()
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/tqdm/notebook.py", line 157, in display
    pbar.value = self.n
    ^^^^^^^^^^
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/traitlets/traitlets.py", line 716, in __set__
    self.set(obj, value)
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/traitlets/traitlets.py", line 706, in set
    obj._notify_trait(self.name, old_value, new_value)
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/traitlets/traitlets.py", line 1513, in _notify_trait
    self.notify_change(
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/ipywidgets/widgets/widget.py", line 700, in notify_change
    self.send_state(key=name)
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/ipywidgets/widgets/widget.py", line 586, in send_state
    self._send(msg, buffers=buffers)
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/ipywidgets/widgets/widget.py", line 825, in _send
    self.comm.send(data=msg, buffers=buffers)
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/comm/base_comm.py", line 144, in send
    self.publish_msg(
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/ipykernel/comm/comm.py", line 42, in publish_msg
    parent=self.kernel.get_parent(),
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dokken/src/mambaforge/envs/test_ipp/lib/python3.12/site-packages/ipykernel/kernelbase.py", line 797, in get_parent
    return self._shell_parent.get()
           ^^^^^^^^^^^^^^^^^^^^^^^^
LookupError: <ContextVar name='shell_parent' at 0x7dd1814c94e0>
```
Only happens in notebooks, and if `ipywidgets` is installed, which might indicate that the problem lies elsewhere.

First observed in: https://github.com/jorgensd/adios4dolfinx/actions/runs/18489789453/job/52692118274
Work if one bounds ipykernel, i.e.
```yml
  - ipykernel<7.0.0
```

minrk:
- removes guessing about whether threads should route output to their originating cell (see #1289)
- thread-local value is persisted, if set explicitly
- if no thread-local value is set, fallback on a global value (usually the latest cell), restores longstanding default behavior

There's lots of duplication around, but I couldn't see a way to resolve that without breaking things. That can be a later goal.

closes #1450 

improves but does not quite close #1289 because there is still no explicit opt-out once a thread has set its parent (there is no `clear_parent()`), but at least the parent isn't set by default anymore, so the opt-out is less urgently needed.

To send a thread's output to a specific cell, this must now be done explicitly, which I think is a big improvement and simplification. No thread inheritance is tracked, no cell guessing occurs.

To route output to a specific cell from a thread, explicitly call `set_parent` in the thread with the parent from the outer context that should be kept for the duration of the thread:

```python
current_cell_parent = get_ipython().get_parent()

def thread_task():
    get_ipython().set_parent(current_cell_parent)
    print("...")

Thread(target=thread_task).start()
```

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -72,6 +72,8 @@ _AWAITABLE_MESSAGE: str = (
     " ipykernel 6.0 (2021). {target} does not seem to return an awaitable"
 )
 
+T = t.TypeVar("T")
+
 
 def _accepts_parameters(meth, param_names):
     parameters = inspect.signature(meth).parameters
@@ -801,7 +803,7 @@ class Kernel(SingletonConfigurable):
 
         return self._get_shell_context_var(self._shell_parent)
 
-    def _get_shell_context_var(self, var: ContextVar):
+    def _get_shell_context_var(self, var: ContextVar[T]) -> T:
         """Lookup a ContextVar, falling back on the shell context
 
         Allows for user-launched Threads to still resolve to the shell's mai context
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -806,7 +806,7 @@ class Kernel(SingletonConfigurable):
     def _get_shell_context_var(self, var: ContextVar[T]) -> T:
         """Lookup a ContextVar, falling back on the shell context
 
-        Allows for user-launched Threads to still resolve to the shell's mai context
+        Allows for user-launched Threads to still resolve to the shell's main context
 
         necessary for e.g. display from threads.
         """
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -17,7 +17,7 @@ import typing as t
 import uuid
 import warnings
 from collections.abc import Mapping
-from contextvars import ContextVar
+from contextvars import Context, ContextVar, copy_context
 from datetime import datetime
 from functools import partial
 from signal import SIGINT, SIGTERM, Signals, default_int_handler, signal
@@ -201,6 +201,7 @@ class Kernel(SingletonConfigurable):
     _control_parent_ident: bytes = b""
     _shell_parent: ContextVar[dict[str, Any]]
     _shell_parent_ident: ContextVar[bytes]
+    _shell_context: Context
     # Kept for backward-compatibility, accesses _control_parent_ident and _shell_parent_ident,
     # see https://github.com/jupyterlab/jupyterlab/issues/17785
     _parent_ident: Mapping[str, bytes]
@@ -320,13 +321,14 @@ class Kernel(SingletonConfigurable):
         self._shell_parent.set({})
         self._shell_parent_ident = ContextVar("shell_parent_ident")
         self._shell_parent_ident.set(b"")
+        self._shell_context = copy_context()
 
         # For backward compatibility so that _parent_ident["shell"] and _parent_ident["control"]
         # work as they used to for ipykernel >= 7
         self._parent_ident = LazyDict(
             {
                 "control": lambda: self._control_parent_ident,
-                "shell": lambda: self._shell_parent_ident.get(),
+                "shell": lambda: self._get_shell_context_var(self._shell_parent_ident),
             }
         )
 
@@ -768,6 +770,8 @@ class Kernel(SingletonConfigurable):
         else:
             self._shell_parent_ident.set(ident)
             self._shell_parent.set(parent)
+            # preserve the last call to set_parent
+            self._shell_context = copy_context()
 
     def get_parent(self, channel=None):
         """Get the parent request associated with a channel.
@@ -794,7 +798,20 @@ class Kernel(SingletonConfigurable):
 
         if channel == "control":
             return self._control_parent
-        return self._shell_parent.get()
+
+        return self._get_shell_context_var(self._shell_parent)
+
+    def _get_shell_context_var(self, var: ContextVar):
+        """Lookup a ContextVar, falling back on the shell context
+
+        Allows for user-launched Threads to still resolve to the shell's mai context
+
+        necessary for e.g. display from threads.
+        """
+        try:
+            return var.get()
+        except LookupError:
+            return self._shell_context[var]
 
     def send_response(
         self,
@@ -1455,7 +1472,7 @@ class Kernel(SingletonConfigurable):
             )
         return self._input_request(
             prompt,
-            self._shell_parent_ident.get(),
+            self._get_shell_context_var(self._shell_parent_ident),
             self.get_parent("shell"),
             password=True,
         )
@@ -1472,7 +1489,7 @@ class Kernel(SingletonConfigurable):
             raise StdinNotImplementedError(msg)
         return self._input_request(
             str(prompt),
-            self._shell_parent_ident.get(),
+            self._get_shell_context_var(self._shell_parent_ident),
             self.get_parent("shell"),
             password=False,
         )
--- a/ipykernel/displayhook.py
+++ b/ipykernel/displayhook.py
@@ -29,6 +29,7 @@ class ZMQDisplayHook:
 
         self._parent_header: ContextVar[dict[str, Any]] = ContextVar("parent_header")
         self._parent_header.set({})
+        self._parent_header_global = {}
 
     def get_execution_count(self):
         """This method is replaced in kernelapp"""
@@ -57,11 +58,16 @@ class ZMQDisplayHook:
 
     @property
     def parent_header(self):
-        return self._parent_header.get()
+        try:
+            return self._parent_header.get()
+        except LookupError:
+            return self._parent_header_global
 
     def set_parent(self, parent):
         """Set the parent header."""
-        self._parent_header.set(extract_header(parent))
+        parent_header = extract_header(parent)
+        self._parent_header.set(parent_header)
+        self._parent_header_global = parent_header
 
 
 class ZMQShellDisplayHook(DisplayHook):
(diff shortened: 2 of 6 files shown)


minrk:
passing everywhere but 3.14, includes missing coverage for #1450

ipykernel/kernelbase.py
```diff
@@ -794,7 +800,20 @@ def get_parent(self, channel=None):
...
         if channel == "control":
             return self._control_parent
-        return self._shell_parent.get()
+
+        return self._get_shell_context_var(self._shell_parent)
+
+    def _get_shell_context_var(self, var: ContextVar[T]) -> T:
+        """Lookup a ContextVar, falling back on the shell context
+
+        Allows for user-launched Threads to still resolve to the shell's mai context
```
ianthomas23: Typo `mai` to `main` ?

ianthomas23 (approved):
LGTM, just one typo.

ianthomas23:
The failing qtconsole test
```
FAILED qtconsole/tests/test_00_console_widget.py::test_scroll[False] - pytestqt.exceptions.TimeoutError: waitUntil timed out in 20000 milliseconds
```
is labelled as flaky in qtconsole. The subshells changes in this repo made it more flaky, and it looks this PR might be making it more flaky still. I'd be inclined to merge this as it is, and if the flaky test in too annoying we could just exclude it from CI here until it can be investigated downstream. What do you think @minrk?

minrk:
That sounds good to me

ianthomas23 (approved):
Thanks @minrk

ianthomas23:
I'm aiming for a 7.0.2 release soon with #1453, this PR, and hopefully a fix that I am working on for #1442/#1454.

krassowski:
Just to understand, this made the behaviour from https://github.com/ipython/ipykernel/pull/1186 opt-in rather than opt-out?

krassowski:
> but at least the parent isn't set by default anymore, so the opt-out is less urgently needed.

Yeah, but it kind of brings back https://github.com/jupyter-widgets/ipywidgets/issues/2358 and https://github.com/jupyterlab/jupyterlab/issues/3936?

I understand the motivation for making it opt-out, and I understand that making it opt-in for now might be easier. If the plan is to release in 7.0.2 I would suggest highlighting this in the release notes, and documenting the workaround of using `set_parent` and `get_parent` in general.

EwoutH:
Thanks for working on this. When will it be included in a new release?

minrk:
@krassowski yes, though it is opt-in instead of _unconditional_, it never was 'opt-out' and since the guess was wrong very often, I think, having a clear opt-in is better than having no opt out at all, which is the choice we were presented with. Getting back to the default behavior once opted in is still a lot trickier than it should be, so room to improve there.

https://github.com/jupyter-widgets/ipywidgets/issues/2358 should be addressed by `Output.__enter__()` setting the appropriate thread-local state. This is an explicit opt-in to capture output for the current thread, so it makes sense that it should do the capture. Note that #1186 also _did not_ fix https://github.com/jupyter-widgets/ipywidgets/issues/2358 either - it still routed output to the wrong place (cell output, not the widget OutputArea), but the same cell, rather than being captured by the OutputWidget.
 
I think having the thread-local state is still a great fix and the right thing to do, it's just the initial guess that was the problem. I think there are still improvements to be made on this front.

I 100% agree on documenting the change, I'll look into that.

ianthomas23:
I've prepared a draft release of 7.1.0 as I realised that #1435 is an enhancement and hence needs to be in a minor not patch release. The draft changelog notes are at https://github.com/ipython/ipykernel/releases; I've written a little summary at the top for the changes but I have not covered this PR. @minrk Would you like to add a line or two of documentation there, or a link to wherever else you're planning to write it?

DonJayamanne:
@ianthomas23  @krassowski
(apologies if I've pinged the wrong people)
Hi, I'm one of the maintainers of Jupyter extension in VS Code here.

Any idea when a new version of IPyKernel will be shipped out with this fix?
Asking as we have quite few a users running into issues due to this.

Thanks for promptly fixing this issue

minrk:
@ianthomas23 thanks, I added a little note to the 7.1 release notes draft. Good to go from my end once I skip the most failing tests in #1463

ianthomas23:
> @ianthomas23 @krassowski (apologies if I've pinged the wrong people) Hi, I'm one of the maintainers of Jupyter extension in VS Code here.
> 
> Any idea when a new version of IPyKernel will be shipped out with this fix? Asking as we have quite few a users running into issues due to this.
> 
> Thanks for promptly fixing this issue

@DonJayamanne ipykernel 7.1.0 has been released.

minrk:
@ianthomas23 thanks for shipping this!

krassowski:
It looks like this was felt across ecosystem. Would anyone like to review https://github.com/jupyter-widgets/ipywidgets/pull/4021?

PR #1453, opened 2025-10-15
update tests for 3.14

minrk:
- move qt matrix out of hatch to CI because qt6 doesn't work on 3.14 and hatch doesn't seem to be able to exclude matrix entries (and we don't use the hatch matrix for anything else)
- skip 3.14t which needs a jupyter-core release (coming soon)
- test less than all combinations of Python + qt
- the subprocess output capture tests seem to not work on 3.14. I don't understand why because the same code definitely does work perfectly in a notebook. Maybe conflicting somehow with pytest's FD output capture. So those tests are now skipped, but it's the _test_ that is broken, not the actual behavior.

### DIFF
--- a/tests/test_kernel.py
+++ b/tests/test_kernel.py
@@ -211,10 +211,14 @@ def test_sys_path_profile_dir():
     assert "" in sys_path
 
 
+# the subprocess print tests fail in pytest,
+# but manual tests in notebooks work fine...
+
+
 @pytest.mark.flaky(max_runs=3)
 @pytest.mark.skipif(
-    sys.platform == "win32" or (sys.platform == "darwin"),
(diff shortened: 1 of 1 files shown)


minrk:
note: this doesn't increase the size of the build matrix, it actually shrinks it. What it does is shift a dimension of the matrix (qt5 or qt6) up from hatch to github, so we can do things like skip qt6 on 3.14 (and skip some others because we don't need every qt on every version, as long as we have upper/lower bounds covered)

ianthomas23:
Note to self (or someone else if they beat me to it): needs backport to 6.x branch

ianthomas23:
@meeseeksdev please backport to 6.x

ianthomas23:
Manually backported by #1460.

PR #1463, opened 2025-10-20
test that matplotlib event loop integration is responsive

minrk:
also fixes `%matplotlib osx` which gets converted to `macosx` on its way through matplotlib and back

This should hopefully be the regression test for #1458

### DIFF
--- a/tests/test_matplotlib_eventloops.py
+++ b/tests/test_matplotlib_eventloops.py
@@ -9,9 +9,10 @@ from .test_eventloop import qt_guis_avail
 from .utils import assemble_output
 
 # these tests don't seem to work with xvfb yet
+# these tests seem to be a problem on CI in general
 pytestmark = pytest.mark.skipif(
-    sys.platform == "linux" and bool(os.getenv("CI")),
-    reason="tests not working yet with xvfb on linux CI",
+    bool(os.getenv("CI")),
+    reason="tests not working yet reliably on CI",
 )
 
 guis = []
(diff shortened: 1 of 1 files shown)


minrk:
all the matplotlib tests are passing on mac and windows, need to figure out why linux isn't enabling the event loops

ccordoba12:
> need to figure out why linux isn't enabling the event loops

Perhaps you need to run that test with `xvfb`?

minrk:
> Perhaps you need to run that test with xvfb?

Probably right, looking at that.

ianthomas23:
Changes look good to me, and tests pass locally on both macos and linux. I don't understand the CI failures but I assume they are related to `xvfb`? Matplotlib itself pip installs `pytest-xvfb` which might be worth a try as with it installed `pytest` automatically uses `xvfb` without having to set the `PRE_TEST` stuff.

minrk:
with xvfb, the tests are _hanging_ after `embed_kernel`, which makes me think it might be during setup of  `test_eventloops`

ianthomas23:
Is it worth just disabling the reliably-failing `test_matplotlib_eventloops.py` on qt5/python3.14 so we can merge and come back to this later?

minrk:
Yes, I think so. Sorry I couldn't figure out how to get these more reliable. They pass 100% every time I test locally, so I think it's something about how slow ci environments are

minrk:
I give up, I just skipped the tests altogether on CI. They are too flaky. I'm sure there's some race somewhere, perhaps in where the kernel runs the cell activating the event loop and then enters the eventloop mode, maybe a message can get dropped or stuck waiting somewhere? It could be a real bug, but I don't want to hold up the 7.1 release anymore, and this does have the `%matplotlib osx` fix that should get in there.

But these tests pass 100% for me locally, so I at least know that they verify `%matplotlib osx` work, so I don't want to delete them. I'll come back and try to track down the problem another time with less pressure.

PR #1464, opened 2025-10-24
Test changing base method to async after #1295

Carreau:
Re-do of #1339, the goal is to detect downstream breakage

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -862,7 +862,7 @@ class Kernel(SingletonConfigurable):
             subshell_id = parent["header"].get("subshell_id")
             self._abort_queues(subshell_id)
 
-    def do_execute(
+    async def do_execute(
         self,
         code,
         silent,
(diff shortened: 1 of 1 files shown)


Carreau:
@ianthomas23 thoughts ? 

As for previous one, mostly as a canary to see if nightly would break stuff, and potentially revert it and redo in a couple of month if that's the case.

ianthomas23:
> @ianthomas23 thoughts ?
> 
> As for previous one, mostly as a canary to see if nightly would break stuff, and potentially revert it and redo in a couple of month if that's the case.

I'm happy to merge this to see if it causes problems elsewhere.

ianthomas23 (approved):

Carreau:
Ok, let's try then, and revert if/when needed.

ianthomas23:
@Carreau I'd like to make a `7.2.0` release as there have been a few changes recently, there is a [7.2.0a0 prerelease](https://pypi.org/project/ipykernel/7.2.0a0/) on PyPI for testing purposes. Do you think these base method `async` changes can be safely included in that, or do we need to revert this before the actual release?

Carreau:
#1339 was opened less than a year ago so I would say to maybe revert for a 7.2.0 minor. At some point though we need to bite the bullet make a release that breaks a few things even if this means re-releasing a patch a few days after.

ianthomas23:
Thanks, I'll remove this before the 7.2.0 release and then re-apply it afterwards.

PR #1475, opened 2025-11-12
pass cell_meta from do_execute to run_cell

erawn:
Other half of [https://github.com/ipython/ipython/pull/15071](https://github.com/ipython/ipython/pull/15071 )

In order to avoid an `if-elif-elif-else` sequence I copied the strategy already used for `execute_request` and made a list of the arguments rather than conditioning different method calls. Happy to switch it back or do something else if others prefer a different style.


P4X-ng (approved):

ianthomas23:
Can you rebase this on `main` as there were some CI failures that I think I have fixed.

ianthomas23 (approved):
Thanks @erawn

There are some linting and docs errors but I will merge it now and fix the errors myself afterwards, as you have been waiting quite a while for this to be accepted.

erawn:
Thank you so much @ianthomas23 ! Apologies for missing your earlier message, and thanks for handling it anyways!

Issue #1381, opened 2025-03-21
Heads up: nest-asyncio is no longer actively maintained


frenzymadness:
Unfortunately, Ewald de Wit, the maintainer of `nest-asyncio`, passed away in March 2024. The project is now archived on GitHub.

It should still work as before—only one test is currently failing with Python 3.14 alpha 6: `test_timeout`.

Since it's a dependency here, I thought I'd let you know. The project may eventually need new maintainers.

frenzymadness:
> Thanks for the heads up, and sorry to hear that. I think we should not use `nest-asyncio` anymore.

If that would be possible, having less dependencies is always better.

Chaoses-Ib:
In case there is still a need of it: I've made a fork of `nest-asyncio` to fix support for Python 3.12 and 3.14: [nest-asyncio2](https://github.com/Chaoses-Ib/nest-asyncio2#comparison-with-nest_asyncio). Since it's also used in my own project, I'm going to maintain it for a long time. There is no behavior change on Python < 3.12, and it's already tested with https://github.com/NVIDIA/NeMo-Agent-Toolkit, https://github.com/pyvista/pyvista, https://github.com/Chaoses-Ib/ComfyScript, so it should be a safe replacement.

lahwaacz:
nest-asyncio was already supposed to be removed in https://github.com/ipython/ipykernel/pull/1081 but for some reason it is still specified in pyproject.toml dependencies

ianthomas23:
> nest-asyncio was already supposed to be removed in [#1081](https://github.com/ipython/ipykernel/pull/1081) but for some reason it is still specified in pyproject.toml dependencies

It was restored just after in #1086.

bolinocroustibat:
> In case there is still a need of it: I've made a fork of `nest-asyncio` to fix support for Python 3.12 and 3.14: [nest-asyncio2](https://github.com/Chaoses-Ib/nest-asyncio2#comparison-with-nest_asyncio). Since it's also used in my own project, I'm going to maintain it for a long time. There is no behavior change on Python < 3.12, and it's already tested with https://github.com/NVIDIA/NeMo-Agent-Toolkit, https://github.com/pyvista/pyvista, https://github.com/Chaoses-Ib/ComfyScript, so it should be a safe replacement.

Great news, thank you @Chaoses-Ib . We might use your fork from now on in our team.

ianthomas23:
[...]

PR #1499, opened 2026-02-27
Switch from using nest-asyncio to nest-asyncio2

ianthomas23:
As explained in https://github.com/ipython/ipykernel/issues/1381#issuecomment-3548626525 `nest-asyncio` is  no longer maintained, so here switching to use `nest-asyncio2` which is a fork that is maintained.


ianthomas23:
Looks good, CI is green.

PR #1503, opened 2026-03-06
Remove dead `json_clean` / `jsonutil` module

erikgaas:
/!\ Needs a new release of ipyparallel as of June 10, 2026

## Summary

Some preliminary exploration trying to get more up to speed with the codebase.

`json_clean()` has been a no-op since jupyter_client >= 7 (released 2021). Since ipykernel now requires `jupyter_client>=8.8.0`, this code can never execute.

## Changes

- Remove `ipykernel/jsonutil.py` entirely
- Remove all `json_clean()` call sites in `kernelbase.py` (7 calls + import)
- Remove `tests/test_jsonutil.py` (all tests were already `skipIf(jupyter_client >= 7)`)

No behavioral change — every removed `json_clean()` call was returning its input unchanged.


Carreau:
I like that ! Apparently there are similar utils in jupyter_client (and the one from jupyter_client is used in jupyter_server). Those should also likely go through a cycle of deprecation/removal !

Thanks, (rebasing to try to get tests to pass)

Carreau:
ipyparallel will need fixeed it imports json clean from ipykernel. (changing to draft, if you or someone else want to send a fix to ipyparallel)

PR #1507, opened 2026-03-06
Deduplicate awaitable check in request handlers

erikgaas:
## Summary

Every request handler in `kernelbase.py` had the same 7-line pattern checking if the `do_*` method returned an awaitable and warning if not. This PR extracts that into a single `_await_result()` helper.

## Changes

- Add `_await_result(result, func_name)` method to `Kernel` class
- Replace 7 identical check-and-warn blocks with one-line calls

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -543,6 +543,16 @@ class Kernel(SingletonConfigurable):
         # This can be removed when minimum python increases to 3.10
         self._control_lock = asyncio.Lock()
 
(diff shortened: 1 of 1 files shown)


Carreau:
Duplication in this case is I think better, because _when_ we remove the pending deprecation branches we can do an a case by cases basis. Not everything need to always be DRY. Plus the function _await_result, does not make it clear to the reader that some path are deprecated.

It's a personal decision and thanks you for suggesting it though I don't belive we want to go that route for this particular case.

PR #1515, opened 2026-05-06
Support ZMQ Curve for transport encryption

krassowski:
This PR adds support for ZMQ Curve cryptography to encrypt kernel communications over TCP, addressing the current limitation of all code and outputs being transmitted in plaintext.

What's changed

- Curve key support in IPKernelApp: Loads `curve_publickey` / `curve_secretkey` from   connection files and applies them to all ZMQ sockets (`iopub`, `shell`, `control`,  `stdin`, `heartbeat`) via new `_apply_curve_server_options()` /  `_apply_curve_client_options()` helpers.
- Kernelspec metadata: Advertises `"supported_encryption": "curve"` so frontends can discover and negotiate encryption capability.
- Emits a warning when TCP transport is used without encryption, guiding users toward IPC or CurveZMQ.
- Heartbeat thread: Extended to accept and apply curve keys to its ROUTER socket.

### References

- Depends on https://github.com/jupyter/jupyter_client/pull/1110
- Enables https://github.com/jupyter/jupyter_client/issues/808
- PoC for https://github.com/jupyter/enhancement-proposals/issues/75
- Toggle on server-level: https://github.com/jupyter-server/jupyter_server/pull/1638

### Code changes

- [x] tests
- [x] additions

### DIFF
--- a/ipykernel/heartbeat.py
+++ b/ipykernel/heartbeat.py
@@ -27,14 +27,29 @@ from jupyter_client.localinterfaces import localhost
 class Heartbeat(Thread):
     """A simple ping-pong style heartbeat that runs in a thread."""
 
-    def __init__(self, context, addr=None):
-        """Initialize the heartbeat thread."""
+    def __init__(self, context, addr=None, curve_publickey=None, curve_secretkey=None):
+        """Initialize the heartbeat thread.
+
+        Parameters
+        ----------
+        context : zmq.Context
+        addr : tuple, optional
+            (transport, ip, port)
+        curve_publickey : bytes, optional
+            Z85-encoded CurveZMQ public key.  When provided together with
+            *curve_secretkey*, the heartbeat socket will operate as a
+            CurveZMQ server so that only authenticated clients can connect.
+        curve_secretkey : bytes, optional
+            Z85-encoded CurveZMQ secret key (paired with *curve_publickey*).
+        """
         if addr is None:
             addr = ("tcp", localhost(), 0)
         Thread.__init__(self, name="Heartbeat")
         self.context = context
         self.transport, self.ip, self.port = addr
         self.original_port = self.port
+        self.curve_publickey = curve_publickey
+        self.curve_secretkey = curve_secretkey
         if self.original_port == 0:
             self.pick_port()
         self.addr = (self.ip, self.port)
@@ -94,6 +109,10 @@ class Heartbeat(Thread):
         self.name = "Heartbeat"
         self.socket = self.context.socket(zmq.ROUTER)
         self.socket.linger = 1000
+        if self.curve_secretkey is not None:
+            self.socket.curve_secretkey = self.curve_secretkey
+            self.socket.curve_publickey = self.curve_publickey
+            self.socket.curve_server = True
         try:
             self._bind_socket()
         except Exception:
--- a/ipykernel/heartbeat.py
+++ b/ipykernel/heartbeat.py
@@ -36,11 +36,11 @@ class Heartbeat(Thread):
         addr : tuple, optional
             (transport, ip, port)
         curve_publickey : bytes, optional
-            Z85-encoded CurveZMQ public key.  When provided together with
+            CurveZMQ public key (Z85). When provided together with
             *curve_secretkey*, the heartbeat socket will operate as a
             CurveZMQ server so that only authenticated clients can connect.
         curve_secretkey : bytes, optional
-            Z85-encoded CurveZMQ secret key (paired with *curve_publickey*).
+            CurveZMQ secret key (Z85, paired with *curve_publickey*).
         """
         if addr is None:
             addr = ("tcp", localhost(), 0)
--- a/ipykernel/heartbeat.py
+++ b/ipykernel/heartbeat.py
@@ -48,8 +48,8 @@ class Heartbeat(Thread):
         self.context = context
         self.transport, self.ip, self.port = addr
         self.original_port = self.port
-        self.curve_publickey = curve_publickey
-        self.curve_secretkey = curve_secretkey
+        self._curve_publickey = curve_publickey
+        self._curve_secretkey = curve_secretkey
         if self.original_port == 0:
             self.pick_port()
         self.addr = (self.ip, self.port)
@@ -109,9 +109,9 @@ class Heartbeat(Thread):
         self.name = "Heartbeat"
         self.socket = self.context.socket(zmq.ROUTER)
         self.socket.linger = 1000
-        if self.curve_secretkey is not None:
-            self.socket.curve_secretkey = self.curve_secretkey
-            self.socket.curve_publickey = self.curve_publickey
+        if self._curve_secretkey is not None:
+            self.socket.curve_secretkey = self._curve_secretkey
+            self.socket.curve_publickey = self._curve_publickey
             self.socket.curve_server = True
         try:
             self._bind_socket()
@@ -141,3 +141,6 @@ class Heartbeat(Thread):
                 raise
             else:
                 break
+
+    _curve_publickey: bytes | None
+    _curve_secretkey: bytes | None
--- a/ipykernel/heartbeat.py
+++ b/ipykernel/heartbeat.py
@@ -27,7 +27,7 @@ from jupyter_client.localinterfaces import localhost
 class Heartbeat(Thread):
     """A simple ping-pong style heartbeat that runs in a thread."""
 
-    def __init__(self, context, addr=None, curve_publickey=None, curve_secretkey=None):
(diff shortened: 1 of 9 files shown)


ipykernel/kernelapp.py
```diff
@@ -274,6 +298,12 @@ def write_connection_file(self, **kwargs: Any) -> None:
             iopub_port=self.iopub_port,
             control_port=self.control_port,
         )
+        if self.enable_curve and self._curve_publickey is not None:
+            # Store Z85-encoded keys as ASCII strings alongside the HMAC key.
+            # Clients that understand CurveZMQ will use these to configure
+            # their sockets; legacy clients ignore the unknown fields.
+            connection_info["curve_publickey"] = self._curve_publickey.decode("ascii")
+            connection_info["curve_secretkey"] = self._curve_secretkey.decode("ascii")  # type: ignore[union-attr]
```
krassowski: This should be handled in jupyter-client instead I think...

krassowski: Done in 2f29d52

ipykernel/kernelapp.py
```diff
@@ -328,13 +356,19 @@ def init_sockets(self):
         self.context = context = zmq.Context()
         atexit.register(self.close)
 
+        if self.enable_curve:
+            self._curve_publickey, self._curve_secretkey = zmq.curve_keypair()
+            self.log.debug("CurveZMQ enabled; generated server keypair")
```
krassowski: Maybe we should log a warning:


```suggestion
            self.log.debug("CurveZMQ enabled; generated server keypair")
        elif self.transport == "tcp":
            self.log.warn("Running kernel without encryption; communication is susceptible to eavesdropping; switch to ICP transport or enable CurveZMQ encryption to dismiss this warning")
```

krassowski: Done in af73137

ipykernel/kernelapp.py
```diff
@@ -328,13 +356,27 @@ def init_sockets(self):
...
+        elif self.transport == "tcp":
+            self.log.warning(
+                "Kernel is running over TCP without encryption."
+                " All communication (including code and outputs) is sent in plain text"
+                " and is susceptible to eavesdropping."
+                " Use IPC transport or set IPKernelApp.enable_curve=True to enable"
+                " CurveZMQ encryption."
+            )
```
krassowski, line 368: `nbclient` downstream tests are failing due to addition of this warning, see:

```
  E       AssertionError: assert '[IPKernelApp] WARNING | Kernel is running over TCP without encryption. All communication (including code and outputs) is sent in plain text and is susceptible to eavesdropping. Use IPC transport or set IPKernelApp.enable_curve=True to enable CurveZMQ encryption.\n[IPKernelApp] WARNING | Kernel is running over TCP without encryption. All communication (including code and outputs) is sent in plain text and is susceptible to eavesdropping. Use IPC transport or set IPKernelApp.enable_curve=True to enable CurveZMQ encryption.' == ''
```

I believe we should keep it and update `nbclient` tests, any objections?

ipykernel/heartbeat.py
```diff
@@ -27,14 +27,29 @@
 class Heartbeat(Thread):
     """A simple ping-pong style heartbeat that runs in a thread."""
 
-    def __init__(self, context, addr=None):
-        """Initialize the heartbeat thread."""
+    def __init__(self, context, addr=None, curve_publickey=None, curve_secretkey=None):
```
Carreau:
```suggestion
    def __init__(self, context, addr=None, *, curve_publickey=None, curve_secretkey=None):
```

Force it to be kwargs ?

ipykernel/heartbeat.py
```diff
@@ -27,14 +27,29 @@
...
+            CurveZMQ secret key (Z85, paired with *curve_publickey*).
+        """
         if addr is None:
             addr = ("tcp", localhost(), 0)
         Thread.__init__(self, name="Heartbeat")
         self.context = context
         self.transport, self.ip, self.port = addr
         self.original_port = self.port
+        self.curve_publickey = curve_publickey
+        self.curve_secretkey = curve_secretkey
```
Carreau: Let's also add types at class level for the new fields.

ipykernel/kernelapp.py
```diff
@@ -188,6 +188,19 @@ def abs_connection_file(self):
     """,
     ).tag(config=True)
 
+    enable_curve = Bool(
+        bool(int(os.environ.get("JUPYTER_ENABLE_CURVE", "0"))),
```
Carreau: I don't like the name it's non obvious, i think it's ok to have "curve" in inner variable names, maybe not in env-var and options. ; should we also trim(), the value of environ.get, or be stricter on it's format ?

ipykernel/kernelapp.py
Carreau: +1 to fix nbclient.

tests/test_curve.py
```diff
@@ -0,0 +1,178 @@
...
+
+import pytest
+import zmq
+
+from ipykernel.kernelapp import IPKernelApp
+
+
+@pytest.fixture
+def temp_folder_path(tmp_path):
+    return str(tmp_path)
```
Carreau: Just fix `_make_app`? It's the only utils that will receive the result of this fixture ?

ipykernel/kernelapp.py
krassowski: I agree, I think we could hide "curve" here as an implementation detail; an argument for keeping it as-is would be alignment with ipyparallel (https://github.com/ipython/ipyparallel/pull/553).

I think an obvious choice would be `enable_transport_encryption`?

Same change would need to follow in https://github.com/jupyter-server/jupyter_server/pull/1638.

CC @minrk in case if you have an opinion here

ipykernel/heartbeat.py
krassowski: And they can be private. Done in 0eec39d

tests/test_curve.py
krassowski: Done in 0c71c214c6a7f5d236ee500693ed6a974360a325

ipykernel/kernelapp.py
minrk: yeah, nbclient tests shouldn't fail if warnings are logged from another package, that's a problem in the test suite

krassowski: I opened https://github.com/jupyter/nbclient/pull/341

pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
     "ipython>=7.23.1",
     "comm>=0.1.1",
     "traitlets>=5.4.0",
-    "jupyter_client>=8.8.0",
+    "jupyter_client @ git+https://github.com/krassowski/jupyter_client.git@add-curve-encryption",
```
krassowski: (needs reverting once jupyter-client PR is merged and released)

pyproject.toml
```diff
@@ -71,6 +71,9 @@ cov = [
 pyqt5 = ["pyqt5"]
 pyside6 = ["pyside6"]
 
+[tool.hatch.metadata]
+allow-direct-references = true
```
krassowski: (this too)

ipykernel/kernelapp.py
```diff
@@ -328,13 +348,26 @@ def init_sockets(self):
         self.context = context = zmq.Context()
         atexit.register(self.close)
 
+        if self.curve_secretkey is not None:
+            self.log.debug("Detected CurveZMQ secret key; using transport encryption")
```
minrk: Probably warrants info-level:

```suggestion
            self.log.info("Detected CurveZMQ secret key; using transport encryption")
```

minrk (approved):

Carreau:
I kicked test; everything passing. Do you want to update the PR description extensively before merge ?

pyproject.toml
krassowski: Here

pyproject.toml
krassowski: here

krassowski (commented):
The PR description is fine, but we should release jupyter-client and update `pyproject.toml` before we merge

ipykernel/kernelapp.py
minrk: enable_transport_encryption / `JUPYTER_ENABLE_TRANSPORT_ENCRYPTION` seems fine to me, if a bit verbose. I can update ipyparallel to match, no big deal.

Carreau (approved):

krassowski:
All tests pass :tada:

I do not have merge rights here, can we get this merged and released?

PR #1516, opened 2026-05-06
Fix downstream test: do not use editable install for `ipyparallel`

krassowski:
- Same as https://github.com/jupyter/jupyter_client/pull/1116
- Due to https://github.com/jupyterlab/maintainer-tools/pull/275


Carreau:
Thanks !

Issue #1523, opened 2026-05-26
Make psutil optional for constrained platforms


yves-surrel:
## Feature request

Would it be possible to make `psutil` an optional dependency, or to provide a configuration flag allowing `ipykernel` to run without `psutil` on constrained platforms?

## Context

I am running `ipykernel` inside a Python-for-Android / Kivy application. This setup works with `ipykernel==6.9.1`, but newer versions import `psutil` during kernel startup. On Android / python-for-android, `psutil` is difficult to support because it contains platform-specific native code and relies on process-inspection APIs that are not always available or useful in this environment.

The kernel itself works for the main use case:

- Jupyter messaging
- remote QtConsole control
- normal Python execution
- NumPy operations
- application-side control through the kernel namespace

The blocking point is that newer `ipykernel` versions require `psutil` even though the features using it appear to be related mainly to process/resource reporting and child-process management.

## Observed psutil usage

From inspection, `psutil` appears to be used in areas such as:

- `usage_request`
- `_process_children`
- `_signal_children`

These seem useful on desktop/server platforms, but they are not essential for all embedded/mobile use cases.

## Proposed behavior

One possible approach would be:

- keep full `psutil` behavior when `psutil` is installed;
- if `psutil` is unavailable, allow the kernel to start anyway;
- degrade gracefully for features requiring process introspection;
- return limited or unavailable resource information from `usage_request`;
- treat child-process enumeration as empty or unsupported.

For example, `ipykernel` could catch `ImportError` around `psutil` and use a small internal fallback layer.

## Why this would help

This would make recent `ipykernel` versions usable on constrained or unusual Python targets, including:

- Android / python-for-Android
- embedded Linux systems
- restricted containers
- environments where native extension builds are not practical

Currently, users in those environments may need to pin to older `ipykernel` versions only to avoid the hard `psutil` dependency.

## Current workaround

For now, I can run successfully with:

```text
ipykernel==6.9.1

Carreau:
Yes, that would be resaonable request, we already I belive have it optional for IPython in pyodide (I think we might even have a shim for that) .

Feel free to send PR.

ianthomas23:
Closed by #1527.

PR #1527, opened 2026-06-10
Make psutil optional at runtime

yves-surrel:
This PR makes ipykernel's direct use of psutil optional at runtime.

When psutil is installed, existing behavior is preserved. When ipykernel.kernelbase is running with psutil unavailable:
- usage_request still sends a valid usage_reply with hostname, pid, and cpu_count
- psutil-dependent CPU and memory telemetry is omitted rather than fabricated
- child-process enumeration returns an empty list
- child-process signalling and progressive termination degrade safely

psutil is moved from required runtime dependencies to the test extra so existing subprocess-shutdown coverage remains active in the test environment.

This is intended for constrained platforms where psutil may be unavailable or difficult to build. Runtime optionality is preferable to platform markers here because some constrained builds, such as Python-for-Android, may report sys.platform == "linux" while psutil is unavailable.

Validation:
- python -m pytest tests/test_kernel_direct.py -q
- python -m pytest tests/test_kernel.py -q

Addresses #1523

The remaining PyPy failures appear unrelated to this PR.

This PR only changes ipykernel's direct `psutil` usage in `kernelbase.py`, plus the corresponding dependency/test updates. It does not modify the subshell execution path, threading code, Windows-specific code, or CI workflow configuration.

One failure is in `tests/test_subshells.py::test_run_concurrently_sequence[...]`, and another appears to be a Windows+PyPy timeout while waiting on a threading condition variable. I do not see a direct connection to the changed `usage_request` fallback or child-process cleanup fallback paths.

I will leave these unchanged unless maintainers think they reproduce specifically because of this branch.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -36,13 +36,6 @@ except ImportError:
     # jupyter_client < 5, use local now()
     now = datetime.now
 
-try:
-    import psutil
-except ImportError:
-    psutil = None
-
-_NO_SUCH_PROCESS = () if psutil is None else psutil.NoSuchProcess
-
 import zmq
 from IPython.core.error import StdinNotImplementedError
 from jupyter_client.session import Session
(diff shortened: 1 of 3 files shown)


ianthomas23 (approved):
I've re-run CI and it is all passing.

Thanks @yves-surrel.

Issue #1438, opened 2025-10-09
ipykernel 7.0.0 release with subshells


ianthomas23:
# Kernel Subshells in ipykernel 7.0.0

This is a placeholder issue for the `ipykernel 7.0.0` release which includes kernel subshells for the first time in a full release, and is due for release on Monday 13th October 2025. This is intended to be the first point of call to obtain information about the release, ask questions and report problems with it.

For those not using subshells it is supposed to be backward compatible with the `6.x` branch, but there are some architectural changes such as the use of a separate thread to handle sending and receiving of shell channel messages which may break assumptions made in downstream libraries and hence this is identified as a major releases as it is potentially backwards incompatible. If you experience problems then please report them, and you can pin `ipykernel < 7` if necessary.

There have been many discussions, issues and pull requests about subshells over the last few years. Linking to all of these would be overwhelming, so here are a few salient links only. Searching in the `ipykernel` github repo for "subshells" will give more links.

The original motivation and changes to the Jupyter protocol are in [Jupyter Enhancement Proposal 91](https://github.com/jupyter/enhancement-proposals/pull/91).

The history in `ipykernel` is complicated by the original implementation being included with changes from `tornado` to `anyio`, but the `anyio` changes have since been sidelined and the second subshell implementation in the current `7.x` branch is on top of `tornado`. Version numbers in related issues and PRs are therefore not always correct.

## ipykernel

- Original implementation in PR #1249, this has a useful diagram and a list of future changes, such as to execution counts, that need to be made.
- Current implementation in PR #1396.
- Details of branch manipulations in #1387 and #1399.

## jupyterlab

- Subshell support added in jupyterlab/jupyterlab#16963. This includes a video and explanation of how you can use subshells in kernels that support them.
- Support for comms over subshells added in jupyterlab/jupyterlab#17363 and jupyterlab/jupyterlab#17364, including some useful videos. This fixes a problem in `ipywidgets` (and hence other widget libraries) of widgets not updating whilst the kernel shell is performing a blocking operation.

## ipympl

- `ipympl 0.9.8` includes matplotlib/ipympl#603 to support comms over subshells.

## Future work

There is future work to be done supporting the `ipykernel 7.0.0` release in downstream projects, and in addressing various issues that need to work with subshells such as those enumerated in the first post of #1249.

ianthomas23:
`ipykernel 7.0.0` released: https://github.com/ipython/ipykernel/releases/tag/v7.0.0.

[...]

PR #1529, opened 2026-06-14
Send the shell reply through the ZMQStream instead of raw on its socket

BoykoNeov:
## Summary

On Windows, ipykernel 7 intermittently drops an `execute_request` on the shell channel: the kernel goes **idle (0% CPU) and never replies**, and the client times out waiting for `execute_reply`. In a headless notebook smoke-test we measured this at **~30% of runs**; *which* cell hangs wanders run to run (it is content-innocent), and only a kernel restart recovers. This looks like the same user-visible hang tracked downstream in microsoft/vscode-jupyter#17228.

Forcing `WindowsSelectorEventLoopPolicy` does **not** help — the kernel already runs a selector loop.

## Root cause

The shell ROUTER socket is **dual-use** on the shell-channel thread:

- a `ZMQStream` reads `execute_request`s off it (`init_kernel` builds `ZMQStream(self.shell_socket, …)`, delivered via `on_recv`), while
- replies go back over the **same socket** out-of-band through a raw `send_multipart` that bypasses the stream, in `SubshellManager._send_on_shell_channel`:

```python
def _send_on_shell_channel(self, msg) -> None:
    assert current_thread().name == SHELL_CHANNEL_THREAD_NAME
    self._shell_socket.send_multipart(msg)   # raw send on the ZMQStream's own socket
```

That out-of-band send **drains the socket's edge-triggered `ZMQ_FD` read edge** — the documented libzmq corollary that *after `zmq_send` a socket may become readable without producing a new read event*. Because the send is not `ZMQStream`-mediated, the read side is never re-armed, so an `execute_request` that arrived **concurrently** strands unread on a registered-but-non-readable fd. The strand is **terminal**: while a backlog is already pending there is no `0 → nonzero` `EVENTS` transition, so no later arrival re-edges it, and the kernel sits idle.

Minimal raw-pyzmq reproduction of just the send-drains-read-edge step (no Jupyter), on a ROUTER (the shell socket type):

```python
import select, time, zmq
def readable(fd): return bool(select.select([fd], [], [], 0)[0])

ctx = zmq.Context()
a = ctx.socket(zmq.DEALER); a.setsockopt(zmq.IDENTITY, b"A")
b = ctx.socket(zmq.ROUTER)
b.bind("tcp://127.0.0.1:5704"); a.connect("tcp://127.0.0.1:5704"); time.sleep(0.3)
bfd = b.getsockopt(zmq.FD)

a.send(b"hello"); time.sleep(0.2)                       # warmup: learn route 'A', clear setup edges
assert b.recv_multipart() == [b"A", b"hello"]
b.getsockopt(zmq.EVENTS)
assert not readable(bfd) and not (b.getsockopt(zmq.EVENTS) & zmq.POLLIN)

a.send(b"req1"); time.sleep(0.2)                        # a new request arrives, UNREAD
assert readable(bfd)                                    # read edge is set

b.send_multipart([b"A", b"reply"]); time.sleep(0.05)    # OUT-OF-BAND send on the same ROUTER
assert not readable(bfd)                                # <-- the send DRAINED the read edge
assert b.getsockopt(zmq.EVENTS) & zmq.POLLIN            # ...while POLLIN stays set...
assert b.recv_multipart(zmq.NOBLOCK) == [b"A", b"req1"] # ...and the request was still queued
```

## Fix

After each out-of-band reply send, schedule the shell `ZMQStream`'s read handler on the shell-channel loop — the **same edge-trap reschedule `ZMQStream._update_handler` already runs internally**:

```python
self._shell_channel_io_loop.add_callback(
    lambda: stream._handle_events(stream.socket, 0)
)
```

The `shell_stream` (built in `init_kernel`) is threaded through `ShellChannelThread` into `SubshellManager` so the reply path can reach it; the re-arm is guarded by `stream is not None and stream.socket is self._shell_socket`. No polling and no new dependency — it moves the existing edge-trap reschedule to the one un-mediated operation that drains the edge.

## Validation

Windows, Python 3.13 and 3.14, pyzmq 27.1.0 / libzmq 4.3.5; three arms × 20 real notebook runs, same machine/session:

| Arm | After the reply `send_multipart` | Wedged |
| --- | --- | --- |
| control | none | 6/20 (30%) |
| sham | identical scheduling overhead, `add_callback(lambda: None)` (no re-arm) | 5/20 |
| **fix** | `add_callback(lambda: stream._handle_events(stream.socket, 0))` | **0/20** |

`P(0/20 | p=0.30) ≈ 0.70^20 ≈ 8e-4`. The `sham` arm isolates the re-arm itself from the added wake-ups (it stays at the control rate). With the diff applied the threaded reference was live on every send (551 re-arms, 0 None/mismatch). Validated against 7.2.0 / 7.3.0, where `_send_on_shell_channel` is byte-identical to `main`.

## A public-API alternative

`stream.flush(zmq.POLLIN)` — a *public* `ZMQStream` method ipykernel already calls in `kernelbase.py` — may be preferable to reaching into `_handle_events`. `flush` is a synchronous drain loop whereas `_handle_events(socket, 0)` is the edge-trap reschedule (related but not identical). I only measured `_handle_events`; happy to switch to `flush` if you'd rather stay on the public API.

## Notes

- Still unfixed on `main`: `_send_on_shell_channel` is a bare `send_multipart`.
- Related: microsoft/vscode-jupyter#17228 (downstream symptom), #1438 (7.0.0 problem hub), #307 (the 4.8.2 fix for the same `ZMQ_FD` edge-trigger bug class), zeromq/pyzmq#2173 (same signaler-drain family, different layer).
- A user-side mitigation — retry the kernel subprocess on the `CellTimeoutError` signature — works today and is orthogonal to this fix.
- Happy to add a regression test or open a companion tracking issue if you'd prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -608,6 +608,9 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp, ConnectionFileMix
         """Create the Kernel object itself"""
         if self.shell_channel_thread:
             shell_stream = ZMQStream(self.shell_socket, self.shell_channel_thread.io_loop)
+            # Hand the stream to the shell-channel thread so SubshellManager can send the
+            # out-of-band reply through the stream rather than raw on the socket (the wedge fix).
+            self.shell_channel_thread.shell_stream = shell_stream
         else:
             shell_stream = ZMQStream(self.shell_socket)
         control_stream = ZMQStream(self.control_socket, self.control_thread.io_loop)
--- a/ipykernel/kernelapp.py
+++ b/ipykernel/kernelapp.py
@@ -412,11 +412,7 @@ class IPKernelApp(BaseIPythonApplication, InteractiveShellApp, ConnectionFileMix
             self.control_socket.router_handover = 1
 
         self.control_thread = ControlThread(daemon=True)
-        self.shell_channel_thread = ShellChannelThread(
-            context,
-            self.shell_socket,
-            daemon=True,
-        )
+        self.shell_channel_thread = ShellChannelThread(context, daemon=True)
 
     def init_iopub(self, context):
         """Initialize the iopub channel."""
--- a/ipykernel/shellchannel.py
+++ b/ipykernel/shellchannel.py
@@ -29,6 +29,10 @@ class ShellChannelThread(BaseThread):
         self._manager: SubshellManager | None = None
         self._zmq_context = context  # Avoid use of self._context
         self._shell_socket = shell_socket
+        # Set by kernelapp.init_kernel after it builds the shell ZMQStream (this thread
+        # is created before the stream). Threaded into SubshellManager so the out-of-band
+        # reply is sent through the stream rather than raw on the socket (the wedge fix).
+        self.shell_stream = None
         # Record the parent thread - the thread that started the app (usually the main thread)
         self.parent_thread = current_thread()
 
@@ -43,6 +47,7 @@ class ShellChannelThread(BaseThread):
                 self._zmq_context,
                 self.io_loop,
                 self._shell_socket,
+                self.shell_stream,
             )
         return self._manager
 
--- a/ipykernel/shellchannel.py
+++ b/ipykernel/shellchannel.py
@@ -7,6 +7,7 @@ from threading import current_thread
 from typing import Any
 
 import zmq
+from zmq.eventloop.zmqstream import ZMQStream
 
 from .subshell_manager import SubshellManager
 from .thread import SHELL_CHANNEL_THREAD_NAME, BaseThread
@@ -28,11 +29,12 @@ class ShellChannelThread(BaseThread):
         super().__init__(name=SHELL_CHANNEL_THREAD_NAME, **kwargs)
         self._manager: SubshellManager | None = None
         self._zmq_context = context  # Avoid use of self._context
+        # No longer passed on to SubshellManager, which now sends on the shell channel
+        # through the stream and never touches the socket. Nothing else reads this.
         self._shell_socket = shell_socket
-        # Set by kernelapp.init_kernel after it builds the shell ZMQStream (this thread
-        # is created before the stream). Threaded into SubshellManager so the out-of-band
-        # reply is sent through the stream rather than raw on the socket (the wedge fix).
-        self.shell_stream = None
+        # Set by kernelapp.init_kernel after it builds the shell ZMQStream, since this
+        # thread is created before the stream exists.
+        self.shell_stream: ZMQStream | None = None
         # Record the parent thread - the thread that started the app (usually the main thread)
         self.parent_thread = current_thread()
 
@@ -43,10 +45,11 @@ class ShellChannelThread(BaseThread):
         # Lazy initialisation.
         if self._manager is None:
             assert current_thread() == self.parent_thread
+            # Also narrows the type for the manager, which takes a non-optional stream.
+            assert self.shell_stream is not None
             self._manager = SubshellManager(
                 self._zmq_context,
                 self.io_loop,
-                self._shell_socket,
                 self.shell_stream,
             )
         return self._manager
--- a/ipykernel/shellchannel.py
+++ b/ipykernel/shellchannel.py
@@ -22,16 +22,12 @@ class ShellChannelThread(BaseThread):
     def __init__(
         self,
         context: zmq.Context[Any],
-        shell_socket: zmq.Socket[Any],
         **kwargs,
     ):
         """Initialize the thread."""
         super().__init__(name=SHELL_CHANNEL_THREAD_NAME, **kwargs)
         self._manager: SubshellManager | None = None
         self._zmq_context = context  # Avoid use of self._context
-        # No longer passed on to SubshellManager, which now sends on the shell channel
-        # through the stream and never touches the socket. Nothing else reads this.
-        self._shell_socket = shell_socket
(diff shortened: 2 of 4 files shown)


BoykoNeov:
The failing checks here are pre-existing flakiness on the Windows runners, not this patch — each is on a code path these three files don't touch:

- **`build (windows-latest, qt5, 3.10)`** — `tests/inprocess/test_kernel.py::test_pylab` timed out while matplotlib built its font cache on a cold runner. That's the in-process kernel (no ZMQ socket, no `SubshellManager`, no `kernelapp`), so the shell-channel change can't reach it.
- **`build (windows-latest, qt5, pypy-3.11)`** — `RuntimeError: Kernel died before replying to kernel_info`, a kernel *startup* crash that occurs before any `execute_request`, i.e. before the re-arm in `_send_on_shell_channel` ever runs. The same pypy-3.11 job passes under qt6 and fails under qt5, and `main` fails this same job identically (same error; 141 passed / 42 skipped, attributed to `test_async_interrupt`), so it isn't introduced here.
- **`enforce-label`** — needs a maintainer label.

Everything that actually exercises the change is green: all CPython Windows builds (which run `tests/test_subshells.py` under the patch), the downstream-project tests, and lint. Happy to rebase or re-run if that's useful.

practicusai:
@BoykoNeov not sure it is relevant, but the in our case of https://github.com/microsoft/vscode-jupyter/issues/17228 , the OS is Ubuntu. I hope the fix will work for multiple OSes.

ianthomas23:
I've confirmed the flaky tests passed the second time around.

> - Happy to add a regression test or open a companion tracking issue if you'd prefer.

A regression test is important.

Is there a simpler solution of the `_send_on_shell_channel` using the `shell_stream` rather than `shell_socket` to `send_multipart`?

BoykoNeov:
Thanks @ianthomas23 — and good the re-run came back green.

Agreed, and I think routing through `shell_stream` is the better fix. The raw `shell_socket.send_multipart` is the one op on the dual-use ROUTER that the stream never sees, so it drains the edge-triggered `ZMQ_FD` read edge and nothing re-arms it. Sending through `shell_stream.send_multipart` keeps the stream the sole user of the socket — its own `_handle_events` recvs any concurrently-queued request first, then re-arms `POLLIN` via `_rebuild_io_state`. So it kills the root cause instead of re-arming after it, and drops the poke into the private `_handle_events`. `_send_on_shell_channel` already runs on the shell-channel thread that owns the stream's loop, so no thread-affinity worries. I've pushed that — the PR is now the stream-send fix plus a regression test.

The test is deterministic rather than statistical: it sets up the strand by hand (queue a request on the ROUTER, consume the `ZMQ_FD` read edge so it can't re-fire), then does the out-of-band reply send and asserts the queued request still lands in `on_recv`. It's behavioural, so it passes whether the reply re-arms the stream or routes through it, and times out on a raw-send body. Verified on Windows; the edge-trigger behaviour it leans on is general, so CI should cover the other platforms.

@practicusai — the mechanism is generic libzmq edge-triggering rather than Windows-specific, so this should help the Linux case too.

ZupoLlask:
Dear @BoykoNeov,

We're seeing some issues with flaky tests on Windows that seem to be somewhat related with this subject.

Would you give us some help [here](https://github.com/ipython/ipykernel/pull/1469), please? 🙏

ianthomas23:
@BoykoNeov This is looking great, thanks.

I think we can go slightly further and not pass in a `shell_socket` in `SubshellManager.__init__` and remove `SubshellManager._shell_socket`. As far as I can tell the `shell_stream` is always defined in that constructor; we can `assert shell_stream is not None` there just in case. I've tried this locally (with the other changes that it causes) and CI passes for me.

ianthomas23:
@BoykoNeov are you happy to continue with this? If not I will take it over as I'd like to get it merged before the next release.

BoykoNeov:
Hi, I am currently on vacation until next week, if you want - take over, otherwise I will see what I can do when I return.
Best regards

BoykoNeov:
Done - `SubshellManager` no longer takes or stores the shell socket, so the send is just a plain `shell_stream.send_multipart` with no branch.

I left `ShellChannelThread`'s `shell_socket` argument alone, since changing that signature is a bigger call than the one you asked for. But it was only holding the socket to hand to the manager, so nothing reads it now - it and the `init_sockets` argument can probably both go. Say the word and I'll do it here, or leave it for a follow-up.

Also: there are two asserts, not one. The extra one is in `ShellChannelThread.manager` - `shell_stream` is typed `ZMQStream | None` there (it's set after construction, since the thread exists before `init_kernel` builds the stream), so mypy needs the narrowing at the `SubshellManager` call. Happy to make it a `RuntimeError` instead if you'd rather it survive `python -O`.

While I was in there I also corrected the class docstring, which said reply sends are wrapped in another lock - that lock is gone, `_lock_cache` is the only one left, and sends are already serialised by running in the shell channel thread.

ianthomas23:
Thanks. Please do remove the `shell_socket` argument to `ShellChannelThread`.

BoykoNeov:
Done — removed, along with the field, since nothing read it.

ianthomas23 (approved):
Thanks @BoykoNeov

thomasjm:
Hi all,

I've confirmed this fixes a problem I've been having. Any chance a new release is planned soon that will include this? Thanks!

PR #1537, opened 2026-07-02
validate variable names in copyToGlobals debug handler

sahvx655-wq:
**Expression injection via unvalidated variable names in copyToGlobals**

While reading through the debug handlers I noticed `copyToGlobals` drops the client-supplied `dstVariableName` straight into `globals()['{dst_var_name}']` and forwards that to `setExpression`, and hands `srcVariableName` over as the value expression, with neither of them checked. Its sibling `richInspectVariables` already runs `str.isidentifier` over the name it evaluates, so the omission here stood out. A name such as `x'] or __import__('os').system('id') or globals()['y` closes the quoted string and injects an arbitrary expression that debugpy then evaluates in the debuggee frame; the forwarded `setExpression` payload showed the escaped quotes plainly once I logged it.

Left as is this lets a malformed or hostile debug request steer evaluation well beyond a variable copy. The change rejects any `dstVariableName`/`srcVariableName` that is not a valid identifier before the expression is built, mirroring the guard the sibling handler already applies. Added a regression under the existing debugger tests that confirms such a request is refused locally rather than forwarded.

### DIFF
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -676,6 +676,15 @@ class Debugger:
         src_var_name = message["arguments"]["srcVariableName"]
         src_frame_id = message["arguments"]["srcFrameId"]
 
+        if not str.isidentifier(dst_var_name) or not str.isidentifier(src_var_name):
+            return {
+                "type": "response",
+                "request_seq": message["seq"],
+                "success": False,
+                "command": message["command"],
+                "message": "dstVariableName and srcVariableName must be valid identifiers",
+            }
+
         expression = f"globals()['{dst_var_name}']"
         seq = message["seq"]
         return await self._forward_message(
--- a/ipykernel/debugger.py
+++ b/ipykernel/debugger.py
@@ -681,6 +681,15 @@ class Debugger:
         src_var_name = message["arguments"]["srcVariableName"]
         src_frame_id = message["arguments"]["srcFrameId"]
 
+        if not str.isidentifier(dst_var_name) or not str.isidentifier(src_var_name):
+            return {
+                "type": "response",
+                "request_seq": message["seq"],
+                "success": False,
+                "command": message["command"],
+                "message": "dstVariableName and srcVariableName must be valid identifiers",
+            }
+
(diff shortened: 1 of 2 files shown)


sahvx655-wq:
Pushed a fix for the test_without_debugpy job. The new test asserted on reply["success"] unconditionally, but with debugpy uninstalled the debugger never starts, so process_request returns an empty dict and the whole thing failed with KeyError: 'success' rather than on anything to do with the guard itself.

It now branches on debugpy the same way the rest of that module does, asserting the rejection when debugpy is present and reply == {} when it isn't. Ran tests/test_debugger.py both ways locally, 12 pass in each.

sahvx655-wq:
Following on from the above, the two checks still showing red aren't from this branch. Test Lint is four "InteractiveShell | None" union-attr errors in debugger.py at lines 91, 446, 459 and 644, none of which this PR touches. I reproduced them against a clean 342cf58 worktree with the same mypy and IPython 9.16, so it looks like the unpinned ipython>=7.23.1 in the mypy hook picking up the newer optional get_ipython() return rather than anything here.

The windows qt5 3.14 job times out inside matplotlib's font_manager while it builds the font cache, which is environmental. Neither is something this branch can fix, but say the word if you'd rather I rebase once main is green.

sahvx655-wq:
Rebased onto main. The four union-attr errors behind the Test Lint failure were fixed there by f53affa, which casts the get_ipython() sites to InteractiveShell, and the last five nightlies on 7c9f9b4 are green. The diff is unchanged, only the base moved; the mypy hook and ruff both pass against the rebased branch locally.

The new run is green across all 36 checks, including the windows qt5 3.14 job that previously timed out in matplotlib's font cache and test_without_debugpy on its new 3.11 matrix entry. Ran tests/test_debugger.py with and without debugpy against the rebased branch, 12 pass in each.

JohanMabille (approved):
Thanks for the fix! I wonder if we should also return a response carrying the `copyToGlobals` command in case of success, but that's orthogonal to this PR.

PR #1542, opened 2026-08-03
Add env var to shut down kernel immediately after start for benchmarking

Carreau:
Setting IPYKERNEL_BENCHMARK_STARTUP_SHUTDOWN stops the event loop right after it starts, so total process wall-clock time measures kernel startup without needing an actual client to send a shutdown request.

This can be used in conjunction with -X importtime and profile viewer like tuna to work on speeding up importing and starting a kernel


Carreau:
I'm fine not merging this, just a suggestion if anyone want to have it later, I can keep it as a local patch.

ianthomas23 (approved):

PR #1544, opened 2026-08-04
Fix typing issue due to new releases

Carreau:
IPython types more things, and update some signatures to match upstream.


Carreau:
This is infuriating, GitHub action create a python end in install hatch that create python env to install precommit that install a python env to run mypy...

krassowski (approved):
Thank you!

PR #1545, opened 2026-08-09
Add an advertise_debugger option so kernel_info need not import debugpy

Carreau:
Deferring the debugger construction does not on its own keep debugpy out of a kernel: `kernel_info` imports `.debugger` to fill in `supported_features`, and every frontend sends a kernel_info_request at startup -- jupyter_client's `wait_for_ready` re-sends it until it gets a reply, on start and on every restart. Left alone, the import is merely moved from initialisation into the first shell request.

Add an `advertise_debugger` trait, auto|true|false, defaulting to auto so nothing changes by default. "auto" asks the debugger module as before; "true" and "false" answer kernel_info without importing anything, which lets a deployment that already knows whether debugpy is available keep it out of the process entirely.

The option only controls what is advertised: setting it to "true" where debugpy is unavailable does not make debug requests work.

--- 

We can also do Bool with allow_none=True, I don't care much, this look more explicit. Just for discussion for now.

### DIFF
--- a/ipykernel/kernelbase.py
+++ b/ipykernel/kernelbase.py
@@ -44,6 +44,7 @@ from traitlets.config.configurable import SingletonConfigurable
 from traitlets.traitlets import (
     Any,
     Bool,
+    CaselessStrEnum,
     Dict,
     Float,
     Instance,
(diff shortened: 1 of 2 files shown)


Carreau:
Alternative is caching on disk the environemnt, package version... etc, or just using find_spec to know wether debugpy is _installed_ without actually loading it. I'm happy leaving on auto for _most_ deployments and leaving this as a tunable value for large deployments even if i would _like_ to have it fast for everybody.

Carreau:
@krassowski I think I will need your help. 
It looks like jupyterlab does send a debug request on new kernel startup which forces debugpy to be loaded this is on top of the kernel_info reply that advertise whether debugging is available or not. I think it might be debugInfo

Some Opus exploration tells me it make sens on page refresh, but not on kernel restart.

PR #1546, opened 2026-08-11
Enable setting and resetting a thread-local parent header

jasongrout:
As mentioned in https://github.com/ipython/ipykernel/issues/1289#issuecomment-2557080838 and https://github.com/ipython/ipykernel/pull/1451, there should be a public api for setting and clearing a thread-local parent header. This PR creates a recommended public api for setting/resetting a single thread's parent without setting the global fallback header. An ipywidget OutputWidget can use this to temporarily redirect output from a single thread, for example.

Changelog entry:

> Output from threads can be explicitly routed using the new `get_ipython().set_thread_parent()`, which also returns a token that can be used to undo the set with `get_ipython().reset_thread_parent(token)`. Unlike `set_parent()`, using `set_thread_parent()` does not affect the default shell parent, so does not affect output routing in other threads. Since the thread parent is stored in a thread ContextVar, techniques for propagating a thread's ContextVars will also propagate the thread parent without having to explicitly call `set_thread_parent()`.

I also added a number of `x.parent = ...` parent setters to be more consistent - I'm not directly going to use those, but happy to remove those if there is an objection.

Claude Fable 5 assisted in this PR

### DIFF
--- a/ipykernel/displayhook.py
+++ b/ipykernel/displayhook.py
@@ -77,7 +77,6 @@ class ZMQDisplayHook:
         """Reset the parent header to undo the set_thread_parent call that returned the token."""
         self._parent_header.reset(token)
 
-
     def set_parent(self, parent):
         """Set the global and thread parent header."""
         self.parent_header = extract_header(parent)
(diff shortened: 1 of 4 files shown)


jasongrout:
@minrk, @krassowski - this finishes out the ideas from #1289 and #1451 to make it easy to set and reset just the thread parent. This will make the OutputWidget's temporary overriding the parent much easier (i.e., helping fix https://github.com/jupyter-widgets/ipywidgets/pull/4020 more cleanly than https://github.com/jupyter-widgets/ipywidgets/pull/4021). I think both of you have the context from #1289 and #1451 to review this?

jasongrout:
Thanks @minrk!

PR #1551, opened 2026-09-03
Defer the psutil import until it is actually needed

Carreau:
psutil is optional and its import is not cheap, yet it was paid for on every kernel startup even though most sessions never ask for usage information or need to enumerate child processes.

It is now imported through `_get_psutil()`, which caches the (possibly None) result on first use.


Carreau:
thanks!

PR #1553, opened 2026-09-03
Bump minimum Python to 3.11, update linting, drop dead branches

Carreau:
- requires-python = ">=3.11"; drop Python 3.10 from CI matrices
- remove code paths that only existed to support Python < 3.11:

Dropping 3.10 was done September of last year, so we are in schedule for this to be merged just one year after and a bit before the release of 3.15, and just about when Python 3.10 will drop out out security-only fixes.

note, 3.10 is also the first version that was released 12 month after it's predecessor, before that it was 18 month.

---  

IPython is already 3.11+, and this will drop quite a number of CI items to make the CI time bearable, until we start enabling 3.15 that will be out in about a month.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,6 @@ jobs:
           - qt5
           - qt6
         python-version:
-          - "3.10"
           - "3.11"
           - "3.12"
           - "3.13"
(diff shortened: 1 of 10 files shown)


Carreau:
thanks !

PR #1556, opened 2026-09-08
Add Zizmor

Carreau:
I did my best, but maybe I missed some permissions, we might found out later. 
It's the default setting of zizmor, we could get auditor level later if we want to be stricter.

Should fix failure in #1549 which can be rebased once this is merged.


Carreau:
@/ent:security

Yann-P (approved):
Thank you!

PR #1559, opened 2026-09-10
Fix output routing when threads inherit context

emecii:
Closes #1491

A background thread can inherit the executing cell's `ContextVar` state (the default on free-threaded CPython, and also possible with `copy_context`). Because ipykernel treated any context-local parent as belonging to the current thread, a persistent worker stayed pinned to the cell that created it.

Store the setting thread ID with each context-local parent, and fall back to the current global parent when another thread observes an inherited context. Explicit `set_parent` and `set_thread_parent` calls in the worker still pin output as before.

The regression runs a worker in a copied context and covers both stream and display output across three cells, including explicit global and temporary thread-local overrides.

🤖🤖 AI-assisted contribution.

### DIFF
--- a/ipykernel/displayhook.py
+++ b/ipykernel/displayhook.py
@@ -28,8 +28,8 @@ class ZMQDisplayHook:
         self.session = session
         self.pub_socket = pub_socket
 
-        self._parent_header: ContextVar[dict[str, Any]] = ContextVar("parent_header")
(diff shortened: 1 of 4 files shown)


Carreau:
+1 for me, but I'd like a review with someone with more threading experience.

Carreau (approved):

Issue #311, opened 2018-02-24
Blocking cell execution waiting for Widget interaction


kafonek:
Hi all,

I am interested in blocking further cell execution in a Jupyter notebook until some sort of widget interaction has occurred.  For instance, letting a user perform a "cell -> run all" command, then displaying a dropdown widget in one cell, and blocking in the next cell until the user has selected an option from that dropdown list.

In a regular notebook setup, widget changes (`comm_msg` over zeromq) won't be picked up by the kernel because it's blocking until the cell is completely executed.  If an [Asynchronous Loop](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Asynchronous.html) is involved, then all cells will be executed (from the "cell -> run all") command immediately if they aren't written in an async fashion.

One option is to override the `kernel.shell_handlers` functions to 'capture' the `execute_request` messages coming in on the stream while letting the `comm_msg` events go through, then to 'replay' the `execute_request` messages after the fact.  If I go with that pattern ([example](https://github.com/kafonek/widget_demos/blob/master/captured%20kernel%20events.ipynb), how do I make the output show up in the output cells for the original input cells instead of all in a single cell where the 'replay' happens?

A second option is to have the `ipykernel` listen for `execute_request` and `comm_*` events on different streams.  There's been discussion  of that idea (@dsblank  @AlexTugarev https://github.com/ipython/ipykernel/issues/65, @jasongrout https://github.com/jupyter/jupyter_client/issues/285), but it doesn't look like the issue has moved forward.  I don't understand the entire front-end/back-end system to know how significant that change would be.  Is this issue a non-starter (https://github.com/ipython/ipykernel/pull/302)?

Thanks.

SylvainCorlay:
In general, people have been trying to do the opposite: prevent the notebook from blocking, for which we offer multiple scenarios.

I don't think that there is a simple method besides changing the execution model of the kernel, or capturing execution requests as explained above.

davidbrochart:
Closed in #1565.

PR #1565, opened 2026-09-21
Allow async cell to process Comm messages

### LINKED ISSUE #646: Async widget methods deadlock due to on_msg not being called
We are developing a Jupyter widget which makes it possible to [embed interactive Unfolded maps into notebooks](https://docs.unfolded.ai/jupyter):

image

Our widget is embedded via an iframe and so the communications with the JavaScript side are asynchronous. Some of the functions we want to add to the widget need to return data asynchronously. To communicate between Jupyter and JS we are using widget's `send` and `on_msg` methods. Our methods return `future`s. It works fine - the messages are being sent and received, but only if we do it in two steps:

image

When we are trying to use `await` so that we can have the results and do something with them within one cell, the execution is blocked forever:

image

We tried to investigate the cause. It turns out that the callback we register with `widget.on_msg` only gets called once the execution of the cell is finished. Hence, if we wait until we obtain the result of the calculation within the same cell where the method was called, we end up in a deadlock.

So far we've had no success in finding a workaround, despite trying many things. For instance, here we run the communication in a separate thread and wait until the query response is added to a blocking queue:
image

Calling the above function fails (after the 2 seconds timeout) with an error saying that the queue is empty:
image
And we don't see the `_receive` function being called.

However, if we check the queue immediately after getting this error, we can see that the queue is not empty - the response was added to it, apparently, *_just after_* the error caused the cell execution to be cancelled: 
image

It might be the same issue as this one: https://github.com/ipython/ipython/issues/12786
but I am not completely sure. I tried installing and running the kernel with the fix from [this PR](https://github.com/ipython/ipykernel/pull/589), but I was getting errors related to traitlet messaging when trying to instantiate our widget.

davidbrochart:
Fixes #646 #311

- [x] Add tests
- [x] Implement the feature


davidbrochart:
I'll merge and make a release if there is no feedback by tomorrow.

JohanMabille (approved):

kylebarron:
This is exciting! the linked issues have been open for a very long time!

(fyi #311 was not closed automatically)

